PR #10, opened 2016-08-08
[WIP] Build upload step interactions and ember-osf code [PREP-92]

el1t:
# Purpose
Have the upload component setup what's needed for the rest of the form: create/select project, upload/select file, and save project guid.

# Changes
- Features
 - Add ember-osf project selection widget
 - Upload file to selected project
 - Create node before uploading
 - Show info/warnings during project selection
- Flattened preprint form structure
- Update `ember-cli-shims` and `ember-power-select`
- Add `ember-collection` as dependency (used in ember-osf components)

# Notes
Requires local osf and login to display project selection.

# Ticket
https://openscience.atlassian.net/browse/PREP-92


app/components/preprint-form-upload.js
```diff
@@ -10,8 +11,31 @@ export const State = Ember.Object.extend(Ember.Freezable, {
...
+    },
+    _setCurrentUser() {
+        this.get('currentUser').load().then(user => this.set('user', user));
+    },
+    onGetCurrentUser: Ember.observer('user', function() {
+        const user = this.get('user');
+        if (user) {
+            loadAll(user, 'nodes', this.get('userNodes'));
+        } else {
+            this.set('userNodes', Ember.A());
```
abought: I would recommend minimizing the amount of work that an individual component does to interact with the store. 

In particular for this case, multiple sections of the form might want access to the information, and shouldn't have to do so separately... one alternative is to have the route own the data (eg via setupController), and pass these variables into the component.

app/components/preprint-form-upload.js
```diff
@@ -10,8 +11,31 @@ export const State = Ember.Object.extend(Ember.Freezable, {
...
+    currentUser: Ember.inject.service(),
+    fileManager: Ember.inject.service(),
     state: State.START,
     uploadState: State.START,
+    selectedNode: null,
+    user: null,
+    userNodes: Ember.A([]),
+    init() {
+        this._super(...arguments);
+        if (this.get('session.isAuthenticated')) {
```
abought: Design question: is there a situation where a user can access the "add preprints" page when they are not logged in?

app/templates/components/preprint-form-project-select.hbs
```diff
@@ -0,0 +1,38 @@
+<h3>Choose existing OSF project:</h3>
```
abought: Minor quibble, but heading numbers are usually intended to convey organization/ outline.  If we want a heading to be a specific size, we should prefer a css class over skipping heading levels. (eg, discourage starting at h3 just to get smaller text)

http://webaim.org/techniques/semanticstructure/#contentstructure
> Technically, lower degree headings should be contained within headings of the next highest degree (i.e., one should not skip heading levels, such as from an <h2> to an <h4>, going down the document).

app/components/preprint-form-project-select.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+    userNodes: Ember.A(),
+    selectedNode: null,
+    isAdmin: Ember.computed('selectedNode', function() {
+        return ~this.get('selectedNode.currentUserPermissions').indexOf('admin');
```
abought: Though the tilde trick does have a following, it's much less readable or idiomatic than `indexOf !== -1`, so we tend to prefer the latter for clarity.

Also, rather than comparing to hard-coded strings like 'admin', we generally encourage using the constants defined in `ember-osf/const/permissions`. 

This is mainly to facilitate maintainability if we rework permissions names in the future.

app/components/preprint-form-upload.js
```diff
@@ -22,6 +46,21 @@ const component = CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            this.set('selectedNode', null);
+        },
+        selectNodeFile(file) {
+            this.send('selectFile', file.get('links.download'));
+        },
+        selectNode(node) {
+            console.log(node);
+        },
+        selectFile(url) {
+            console.log(url);
```
abought: You may want to give some thought to data flow: how will the selection choice be used outside this component, to fill in the overall form responses? Actions (or closure actions) might be helpful.

app/templates/components/preprint-form.hbs
```diff
@@ -1,33 +1,33 @@
 {{#with _names.[0] as |name|}}
-    {{#preprint-form-section id='preprint-form-upload' open=true}}
+    {{#preprint-form-section id='preprint-form-upload' open=true name=name}}
         {{preprint-form-header name=name valid=(get valid name)}}
-        {{preprint-form-upload verify=(action 'verify' name)}}
+        {{preprint-form-upload verify=(action 'verify' name) next=(action 'next' name)}}
```
abought: What is the advantage of having a single section of the form be a separate component? (do we intend to reuse these form elements in any way?)

Components are nice for isolating reusable elements, but if neither of those concerns apply, they also add complexity (behavior spread across extra files, and an extra layer of reasoning about data flow).

app/components/preprint-form-upload.js
```diff
@@ -10,8 +11,31 @@ export const State = Ember.Object.extend(Ember.Freezable, {
 }).create().freeze();
 
 const component = CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
```
abought: Low priority, but nice to have when you need it:

If you expect a component to be reusable, it's highly beneficial to provide a docstring describing how to use the component (eg what input variables does it expect, and what data types should they be). This docstring should be the authority on how to use the component and be updated when usage changes.

Here is one basic example, in the YUIDoc format we use.
https://github.com/abought/ember-osf/blob/4f2947e7a0c3622b8f1c21d89e321caadf919b4e/addon/components/comment-detail/component.js#L24-L24

app/components/preprint-form-upload.js
```diff
@@ -10,8 +11,31 @@ export const State = Ember.Object.extend(Ember.Freezable, {
...
+    session: Ember.inject.service(),
+    currentUser: Ember.inject.service(),
+    fileManager: Ember.inject.service(),
     state: State.START,
     uploadState: State.START,
+    selectedNode: null,
+    user: null,
+    userNodes: Ember.A([]),
+    init() {
+        this._super(...arguments);
```
abought: This is a case where we might be better off passing the user in to the component, rather than fetching it in an `init()` method.

app/templates/components/preprint-form-project-select.hbs
```diff
@@ -0,0 +1,38 @@
...
+                <p class="alert alert-warning" role="alert">
+                    This project is private and will be made public
+                </p>
+            {{/unless}}
+            {{#if fileSelect}}
+                {{file-browser}}
+            {{/if}}
+        {{else}}
+            <p class="alert alert-danger" role="alert">
+                You must be the admin of this component to share it.
```
abought: We may want to revisit the language here, including avoiding abbreviations ("admin") not used in the osf.

el1t:
Thank you Andy! Addressing these now.

app/controllers/add-preprint.js
```diff
@@ -1,87 +1,207 @@
...
+            loadAll(user, 'nodes', this.get('userNodes'));
+        } else {
+            this.set('userNodes', Ember.A());
+        }
+    },
+    onGetCurrentUser: Ember.observer('user', function() {
+        this._refreshNodes();
+    }),
+
+    filter: [{}, {}, {}],
```
abought: What is being filtered here? Is this hard-coded test data?

el1t: The objects get a `value` key set on them with the filter keyword. This is to allow ember to observe changes in the array indices themselves with `filter.@each.value`.

abought: Hmm. This appears to copy code from the previous structure. 

Per our discussion Monday, I still think a lot of refactoring remains on these sections- but so long as this is rearranging old code (not adding new usages of same patterns), then it's not a blocker for this PR.

app/controllers/add-preprint.js
```diff
@@ -1,87 +1,207 @@
...
-            }).catch(error => {
-                console.error(error);
-                this.get('toast').error('Could not create preprint project!');
+                this.set('preprintNode', node);
+                this._refreshNodes();
+                this.send('startUpload', resolve);
             });
-            //TODO: eventually make a call to set an OSF file as a preprint (will probably need a flag for such)
+        },
+        deleteProject(resolve) {
```
abought: Why is this function being passed a `resolve` argument (which is a function)?

This seems like: 
a) A strange use of promises or other semantics, and
b) If this is necessary, the parameters of these custom functions should be documented in a followup PR, to help future developers understand the data flow and the intended behavior

app/controllers/add-preprint.js
```diff
@@ -1,87 +1,207 @@
...
+        },
+        preUpload(ignore, dropzone, file) {
+            this.set('uploadFile', file);
+            return new Ember.RSVP.Promise(resolve => this.set('resolve', resolve));
+        },
+        startUpload(resolve) {
+            this.set('_url', `${this.get('preprintNode.files').findBy('name', 'osfstorage').get('links.upload')}?kind=file&name=${this.get('uploadFile.name')}`);
+            this.get('resolve')();
+            this.get('toast').info('File will upload in the background.');
+            resolve();
```
abought: Line 51 and 54 are calling two different functions called resolve? That seems... confusing.

app/controllers/add-preprint.js
```diff
@@ -1,87 +1,207 @@
...
+                // TODO: reset dropzone, since uploaded file has no project
+            }
+            resolve();
+        },
+        preUpload(ignore, dropzone, file) {
+            this.set('uploadFile', file);
+            return new Ember.RSVP.Promise(resolve => this.set('resolve', resolve));
+        },
+        startUpload(resolve) {
+            this.set('_url', `${this.get('preprintNode.files').findBy('name', 'osfstorage').get('links.upload')}?kind=file&name=${this.get('uploadFile.name')}`);
```
abought: We should avoid hardcoded links wherever possible. What does this do for us that the ember-osf `fileManager` service does not?

app/controllers/add-preprint.js
```diff
@@ -1,87 +1,207 @@
...
+        },
+        success() {
+            this.get('toast').info('File uploaded!');
+        },
+        deleteSubject(key, array = key.split('.')) {
+            this.set(key, null);
+            // Delete key manually
+            switch (array.length) {
+                case 2:
+                    delete this[array[0]][array[1]];
```
abought: Is this still relevant to how new taxonomies work, using live data from the server?

It seems the new structure is flatter and based on individual subject IDs, rather than deeply nested key names.

app/controllers/add-preprint.js
el1t: Ah it's a callback for specifying what to do next, alternative would be to check which state the form is in inside the controller.

app/controllers/add-preprint.js
```diff
@@ -1,87 +1,207 @@
...
+                    Object.keys(selected).length === 0) {
+                    // Deselecting a subject: if subject is last item in args,
+                    // its children are showing, and no children are selected
+                    this.send('deleteSubject', `selected.${prev}`, ['selected', ...arr.splice(0, i)]);
+                    args.popObject();
+                }
+                return `${prev}.${cur}`;
+            };
+            // Process past length of array
+            [...args.map(arg => arg.name || arg), ''].reduce(process);
```
abought: This is code from a previous PR. We may not have time to fix it in this PR, but it's... pretty weird.

app/controllers/add-preprint.js
el1t: It's not, but @hmoco was going to work on loading live data (it will also need to load incrementally, and search on the server)

app/controllers/add-preprint.js
el1t: Using `dropzone-widget` with `fileManager` would be best suited to a more thoroughly wrapped dropzone component that can suppress dropzone's default behavior and use `fileManager` behind the scenes. It's probably something for `ember-osf` to implement.

app/controllers/add-preprint.js
abought: Ok. As long as it's actively in progress to change, and we're not continuing development on this old path, that's fine by me. :)

app/controllers/add-preprint.js
el1t: Good point, perhaps I should rename the argument to `callback`?

app/controllers/add-preprint.js
abought: My point is that `fileManager` is supposed to be handling uploading actions. If you are manually managing file actions, and *especially* manually generating URLs, that's a sign that we need to address a gap. 

Feel free to create a ticket with specific task description and deliverable notes to make sure it gets followed up on.

app/controllers/add-preprint.js
abought: Closure actions are probably the way to go here for the future- they let you use an action in a way that returns a value for future followup. It's pretty handy, though admittedly very hard to find good documentation.

Try to avoid new uses of this callback pattern if possible, and document the meaning of the parameter in the code for now, please.

app/controllers/add-preprint.js
```diff
@@ -1,87 +1,207 @@
...
+    }),
+
     actions: {
-        preUpload(comp, drop, file) {
-            this.set('latestFileName', file.name);
-            return new Ember.RSVP.Promise(resolve => {
-                this.set('resolve', resolve);
-            });
+        verify(name, state) {
+            this.get('valid').set(name, state);
```
abought: For the future, we'll want to avoid writing our own validation mechanism in favor of a shared standard package. 

(eg no more development along these lines)

app/controllers/add-preprint.js
```diff
@@ -1,87 +1,207 @@
...
             node.save().then(() => {
-                this.set('_url', `https://test-files.osf.io/file?path=/${this.get('latestFileName')}&nid=${node.id}&provider=osfstorage`);
-                this.set('openModal', false);
-                this.get('resolve')();
-            }).catch(error => {
-                console.error(error);
-                this.get('toast').error('Could not create preprint project!');
+                this.set('preprintNode', node);
+                this._refreshNodes();
+                this.send('startUpload', resolve);
```
abought: In terms of timing, it seems possible that the `startUpload` action could begin before `refreshNodes` has ended. This could point to a potential async bug.

app/styles/app.scss
```diff
@@ -595,6 +596,9 @@ hr {
             }
         }
     }
+    button {
```
abought: This seems like a very general style, that would affect more than just your component. Something to watch out for.

app/templates/add-preprint.hbs
```diff
@@ -1,4 +1,166 @@
...
+                                    {{/if}}
+                                {{/liquid-bind}}
+
+                            {{else if (eq currentState _State.EXISTING)}}
+                                {{preprint-form-project-select userNodes=userNodes fileSelect=true selectedNode=preprintNode}}
+                                <button class="btn btn-default" {{action 'changeState' _State.START}}>Back</button>
+                                <button class="btn btn-primary" disabled={{unless preprintNode true}} {{action 'next' name}}>Next</button>
+
+                            {{else}}
+                                {{log 'preprint-form-upload: bad state'}}
```
abought: In general, try to clear log statements out before submitting a PR.

app/controllers/add-preprint.js
el1t: That's actually dead code my bad :smile:

app/controllers/add-preprint.js
el1t: `refreshNodes` and `startUpload` are independent, though—`refreshNodes` updates the list of nodes (`userNodes`), but the newly created node will already be set to `preprintNode`.

app/styles/app.scss
el1t: It's nested under an id selector.

app/templates/add-preprint.hbs
el1t: This one is more of a warning, since it should never occur under proper use.

app/styles/app.scss
abought: Ah. For the followup PR, styles for a specific component should go in a separate file, not the main shared app.scss. This will give us a bit more isolation and support refactoring.

Move them to `styles/componentName.scss`, and `@import` into app.scss at the top.

app/controllers/add-preprint.js
abought: That still leaves the question of what `this.resolve` is, of course....

At minimum, whatever you can't finish by tomorrow should be documented so future developers can understand and follow from there.

app/controllers/add-preprint.js
abought: Hmm. It still seems to me that calling `refreshNodes` is not the most efficient way to go about getting accurate information after saving a node. That sounds like a followup, not a blocker, though.

app/controllers/add-preprint.js
```diff
@@ -1,87 +1,207 @@
...
+            }
+            resolve();
+        },
+        preUpload(ignore, dropzone, file) {
+            this.set('uploadFile', file);
+            return new Ember.RSVP.Promise(resolve => this.set('resolve', resolve));
+        },
+        startUpload(resolve) {
+            this.set('_url', `${this.get('preprintNode.files').findBy('name', 'osfstorage').get('links.upload')}?kind=file&name=${this.get('uploadFile.name')}`);
+            this.get('resolve')();
```
abought: Out of curiosity: once the `resolve` method of the promise here is called, it would seem like it has done its work. Why are we still stashing it at all?

(This is one argument against stashing callbacks in some intermediate location, instead of being more explicit about the flow of data: it's easy to forget to clean up after ourselves. The solution is to change the data flow, not to blank out this field)

el1t: Yes it's unnecessary after use here. If the file upload functionality were implemented inside `ember-osf`, this would not be necessary. However, since the url constructed at time of upload, the file upload has to be delayed.

abought:
More work is needed for followup, but it doesn't appear to break anything immediately, and others need the reorganization. Merging.

app/controllers/add-preprint.js
abought: https://github.com/abought/ember-osf/blob/69fb8cf8e510b70a4e72cc5d4dbb7ea8d26cf1d2/addon/services/file-manager.js#L178-L178

?

PR #12, opened 2016-08-08
Feature/edit view page

geeksnglitter:
## Purpose
Create the html structure and JS interactions for the view page. Add the supplementary file browser.

## Changes
Clean up view page HTML.
Add share-popup.
Add supplementary files widget.
Make MFR switch between files.

## Ticket

## Next Steps
- [x] Add in file data
- [ ] Error handling

## Future additions
- Display file images based on file type
- Make sure folder actually links to project
- Make supplementary file component display correctly for mobile
- Disable arrows when you are at the end of a file
- Make supplementary file browser appear directly below rendered component (make length of iframe variable, add animation)
- Fix share popover (it squishes)


app/components/supplementary-file-browser.js
```diff
@@ -0,0 +1,62 @@
+import Ember from 'ember';
+import $ from 'jquery';
```
abought: You can also access `Ember.$` (global jquery) or `this.$.()` (selector for elements inside this container)

app/components/supplementary-file-browser.js
```diff
@@ -0,0 +1,62 @@
...
+        if(this.get('fullList')){
+            return this.get('fullList').slice(this.get('startValue'), this.get('startValue') + this.get('numShowing'));
+        }
+    }),
+    init: function(){
+        this._super(...arguments);
+        this.set('selectedFile', this.get('primaryFile'));
+    },
+    didInsertElement: function () {
+        $('[data-toggle="tooltip"]').tooltip({});
```
abought: In general, we're trying to avoid embedding jquery selectors / presentation logic in the javascript controller, when possible. 

Can we do this with markup instead? There is a library called `hint-css` that Sam and I discussed evaluating for ember-osf, but haven't had a chance to look at. I'd be interested to hear your thoughts on it, as we're not 100% wed to porting over all the current OSF tooltip styling code to ember.

app/templates/preprints/view.hbs
```diff
@@ -3,74 +3,51 @@
...
+                    <p>
+                        <strong>APA</strong>
+                        <br>
+                        Cummings, J. N., Butler, B., & Kraut, R. (2002). The quality of online social
+                        relationships. Communications of the ACM, 45(7), 103-108.
+                    </p>
+                    <br>
+                        {{!PROJECT BUTTON}}
+                    <div class="p-lg osf-box-lt" id="view_project">
+                        <span> <a type="button" class="btn btn-primary pull-right" href={{concat model.baseUrl model.project}}>View project</a> The project for this paper is available on the Open Science Framework</span>
```
abought: Don't projects have a self link? That would let us avoid hard-coding URLs in templates, which would help maintainability.

app/routes/preprints.js
```diff
@@ -7,12 +7,14 @@ export default Ember.Route.extend({
...
...2d5"}}, {name: 'greetings.txt', kind: "file", links:{download: "https://test-files.osf.io/v1/resources/qudgr/providers/osfstorage/57a379cbda3e24004afba2d5"}}, {name: 'sup.txt', kind: "file", links:{download: "https://test-files.osf.io/v1/resources/qudgr/providers/osfstorage/57a379cbda3e24004afba2d5"}}, {name: 'RENAMED', links:{download: "https://test-files.osf.io/v1/resources/qudgr/providers/osfstorage/57a379cbda3e24004afba2d5"}}, {name: 'howdy.docx', kind: "file", links:{download: "https://test-files.osf.io/v1/resources/6kf7d/providers/osfstorage/57a393b9da3e2401ee2ab1e2"}}, {name:'hiya.pdf', kind: "folder"}, {name: 'wuzzup.jpg', kind: "file", links:{download: "https://test-files.osf.io/v1/resources/qudgr/providers/osfstorage/57a379cbda3e24004afba2d5"}}, {name: 'wuddup.txt', kind: "file", links:{download: "https://test-files.osf.io/v1/resources/qudgr/providers/osfstorage/57a379cbda3e24004afba2d5"}}, {name: 'greetings.txt', kind: "file", links:{download: "https://test-files.osf.io/v1/resources/qudgr/providers/osfstorage/57a379cbda3e24004afba2d5"}}, {name: 'PRIMARYFILE', links:{download: "https://test-files.osf.io/v1/resources/qudgr/providers/osfstorage/57a379cbda3e24004afba2d5"}}],
```
abought: Is there a good way to centralize where test data lives until a production server is ready? (I assumed that was the purpose of mirage, but we seem to have gotten away from that)

app/templates/components/supplementary-file-browser.hbs
```diff
@@ -0,0 +1,26 @@
...
+        <button class="show-more-button text-muted left" {{action 'moveLeft'}}>
+            <i class="fa fa-caret-left"></i>
+        </button>
+        {{#liquid-bind supplementList elementId="file-list" tagName="ul" class="p-l-md p-r-md" use=scrollAnim as |list|}}
+            {{#each list as |supplement|}}
+                <li class="small-files col-xs-2 p-md box {{if (eq selectedFile supplement) 'selected'}}">
+                    {{#if (eq supplement.kind "folder")}}
+                        <i class="fa fa-folder fa-3x" aria-hidden="true" href={{concat model.baseUrl model.project}}><h6>{{supplement.name}}</h6></i> {{!Actually make a link here}}
+                    {{else}}
+                        <i class="fa fa-file-text fa-3x" aria-hidden="true" {{action 'changeFile' supplement}}><h6>{{supplement.name}}</h6></i>
```
abought: For accessibility reasons, we try to avoid markup that treats text elements (`span`, `i`, etc) as clickable. Screen readers don't automatically know to apply "click here" behaviors, because they use markup as a guide to what the document does.

For example, we might rewrite this as `<button {{action 'lightsCamera'}}><i class"fa fa-lorna">{{supplement.name}}</i>`

Is the heading tag a requirement for some other library? Screen readers also use headings to indicate a document headline, so using them for label text may be  confusing.

app/templates/preprints/view.hbs
```diff
@@ -3,74 +3,51 @@
...
-                    {{!TODO: EMBER OSF MFR }}
-
-                </div> {{!END MFR COL DIV}}
+            <div class="row p-l-md"> {{!PREPRINT INFO ROW}}
+                <div class="col-md-10">
+                    <h5>{{model.preprint.authors}}</h5>
+                    <h6>Added on {{moment-format model.preprint.date}}</h6>
+                </div>
+                <div class="col-md-2">
+                    <h6 class="m-t-lg"><span>Provider: {{#link-to "index"}}SocArXiv{{/link-to}} </span><img src="preprints/assets/img/logo_socarxiv.jpg" width="30"></h6>
```
abought: If you happen to see any weirdness with image display after running inside the OSF, this may be helpful- rather than hardcoding images, we're using styles throughout.

(ember uses this to handle rootURL behavior)

https://github.com/CenterForOpenScience/ember-preprints/commit/5685f833828c262669a71f5b4b4b8363457f8f28#diff-2de80de59c74fd0a521d4a0dc3531cddR734

app/templates/preprints/view.hbs
```diff
@@ -3,74 +3,51 @@
...
-                                View project
-                            </a>
-                        </div> {{!END BUTTON COL 4}}
-                    </div>
-                    <br>
+                    <div class="p-md osf-box-lt" id="download_project">{{!SHARE ROW}}
+                        <i class="fa fa-download" aria-hidden="true"></i>
+                        <a href={{model.downloadLink}}>Download</a>
+                        <i class="fa fa-edit" aria-hidden="true"></i>
+                        {{#link-to "preprints.edit" model.preprint.id}}Edit{{/link-to}}
```
abought: If you pass `preprint` instead of `preprint.id`, it will use the model given. Basically, it saves the page from making another request.

(the catch, of course, is that the model you pass has to be what the model hook would have returned normally! Eg it doesn't work so well if the model hook would have returned an RSVP.hash)

abought:
Added a few incidental comments on potentially useful tricks. This is, of course, recognizing the page is a work in progress, and coming along. Look forward to seeing the new stuff!

app/components/supplementary-file-browser.js
caneruguz: @abought  hint-css looks very promising, and initializing bootstrap tooltip has  had some drawbacks in some other projects for me too. @geeksnglitter can you take a look at switching this with hint-css?

abought: I don't know if it supports popovers, but if we can avoid the extra workarounds we have to do, that seems like it would be nice. :)

app/routes/preprints.js
geeksnglitter: Yes, we usually use mirage. I was using this for quick testing because we didn't already have mocked files. Should get some data closer to real in here soon.

app/components/supplementary-file-browser.js
geeksnglitter: Actually I've removed the tooltip from my current implementation and am only using a popover. Will revaluate if UI changes to use tooltips again. 👍

app/styles/app.scss
```diff
@@ -462,12 +462,6 @@ main {
     padding-top:1.1em;
 }
 
-
-#mfr {
-    padding-left: 25px;
```
caneruguz: You can use osf-style helper for padding like this:
```
   ... class="p-l-lg p-r-lg"
```
https://centerforopenscience.github.io/osf-style/style.html#layout-helper

app/templates/components/supplementary-file-browser.hbs
```diff
@@ -0,0 +1,26 @@
+{{#file-renderer download=selectedFile.links.download
+    width="100%" height="700"}}
+{{/file-renderer}}
+
+{{#if supplementList}}
```
caneruguz: Following from another comment @abought made, I'm wondering if it makes sense to separate the file browser here as a component.  It will not be reused within our app.

app/templates/preprints/view.hbs
```diff
@@ -3,74 +3,51 @@
...
+                    <h6 class="m-t-lg"><span>Provider: {{#link-to "index"}}SocArXiv{{/link-to}} </span><img src="preprints/assets/img/logo_socarxiv.jpg" width="30"></h6>
+                </div>
+            </div> {{!END PREPRINT INFO ROW}}
+            <div class="row">
+                <hr color="#C7FFC7">
+            </div>
+            <div class="row p-md"> {{!CONTENT ROW}}
+                <div class="col-md-7"> {{!LEFT COL DIV}}
+                    {{#if model.supplement}}
+                        {{supplementary-file-browser primaryFile=model.primary files=model.supplement}}
```
caneruguz: Following from another comment @abought made, I'm wondering if it makes sense to separate the file browser here as a component.  It will not be reused within our app.

app/templates/preprints/view.hbs
```diff
@@ -3,74 +3,51 @@
...
+                    <p> {{model.preprint.abstract}} </p>
 
+                    <h3> Citation: </h3>
+                    <p>
+                        <strong>APA</strong>
+                        <br>
+                        Cummings, J. N., Butler, B., & Kraut, R. (2002). The quality of online social
+                        relationships. Communications of the ACM, 45(7), 103-108.
+                    </p>
+                    <br>
```
caneruguz: Don't use <br> for layout (to create space between things). use ```<p class="m-b-md">``` or something similar

caneruguz:
@geeksnglitter there is a socarxiv image here where the provider logo is going. This meta data about the provider will not be coming by the demo. I know that it is in the wireframe but let's remove the picture from ui and the PR.

app/templates/components/share-popup.hbs
```diff
@@ -0,0 +1,2 @@
+<i class="fa fa-share-square-o" aria-hidden="true"></i>
+<a id="share-popover" data-html="true" rel="popover" data-placement="top" data-content='<div><a><i aria-hidden="true" class="fa fa-twitter-square"></i>Tweet</a><br><a><i aria-hidden="true" class="fa fa-facebook-square"></i>Post to Facebook</a><br><a><i aria-hidden="true" class="fa fa-linkedin-square"></i>Share on LinkedIn</a><br><a><i aria-hidden="true" class="fa fa-envelope-square"></i>Send in email</a><br></div>' data-original-title="Share">Share</a>
```
caneruguz: Making these links functional will be a next step but not necessary for this PR.

app/styles/app.scss
geeksnglitter: This was a remnant of old code. I've been using the helper class for padding :)

PR #13, opened 2016-08-08
[hold][WIP] add tags widget to preprint form

ml3ha:
Add ember-osf's tags widget into the preprint form. This is a front-end only feature.

### DIFF
--- a/app/templates/components/preprint-form-basics.hbs
+++ b/app/templates/components/preprint-form-basics.hbs
@@ -9,7 +9,7 @@
     </label>
     <label>
         Tags:
-        {{input class="form-control" value=tags}}
+        {{tags-widget}}
     </label>
 </div>
 <div class="col-md-6">
--- a/bower.json
+++ b/bower.json
@@ -1,17 +1,18 @@
 {
-  "name": "preprint-service",
-  "dependencies": {
-    "ember": "~2.7.0",
-    "ember-cli-shims": "0.1.1",
-    "ember-qunit-notifications": "0.1.0",
-    "osf-style": "https://github.com/CenterForOpenScience/osf-style.git#5dcf5997ad2808ffc7c8564f94cde102bf7a4a96",
-    "font-awesome": "~4.6.3",
-    "jquery": "^2.2.4",
-    "pretender": "~1.1.0",
-    "Faker": "~3.1.0",
-    "bootstrap-treeview": "^1.2.0",
-    "dropzone": "^4.3.0",
-    "bootstrap-sass": "^3.3.6",
-    "toastr": "^2.1.2"
-  }
+    "name": "preprint-service",
+    "dependencies": {
+        "ember": "~2.7.0",
+        "ember-cli-shims": "0.1.1",
+        "ember-qunit-notifications": "0.1.0",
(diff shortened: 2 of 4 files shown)


bower.json
```diff
@@ -1,17 +1,18 @@
...
-    "font-awesome": "~4.6.3",
-    "jquery": "^2.2.4",
-    "pretender": "~1.1.0",
-    "Faker": "~3.1.0",
-    "bootstrap-treeview": "^1.2.0",
-    "dropzone": "^4.3.0",
-    "bootstrap-sass": "^3.3.6",
-    "toastr": "^2.1.2"
-  }
+    "name": "preprint-service",
```
caneruguz: @ml3ha was this formatting change necessary? It's cluttering the PR

package.json
```diff
@@ -46,6 +46,7 @@
     "ember-cli-test-loader": "^1.1.0",
     "ember-cli-uglify": "^1.2.0",
     "ember-collapsible-panel": "1.0.0",
+    "ember-cp-validations": "2.9.3",
```
caneruguz: Probably not part of this PR.

bower.json
ml3ha: @caneruguz The large differences here are because of formatting issues. When I initially added the jquery.tagsinput dependency, it was indented differently from everything else, which puzzled me. So I auto-indented the rest of the lines which is why it looks different.

abought: Most JS tools indent things by 2 characters by default, but our style guide calls for 4.

Usually, we leave bower and package.json alone, because the tools that auto-regenerate those files do unpredictable things with indentation and don't always obey our rules. In general we just try to match the indentation that was already there to avoid noisy PRs.

app/templates/components/preprint-form-basics.hbs
```diff
@@ -9,7 +9,7 @@
     </label>
     <label>
         Tags:
-        {{input class="form-control" value=tags}}
+        {{tags-widget}}
```
abought: h/t to @caneruguz for noticing: as written, this tags widget won't be able to do anything.

See documentation here. Essentially, the component must be passed three things:
http://abought.github.io/ember-osf/classes/tags-widget.html

1. A list of tags to display
2. An action that implements adding a tag
3. An action that implements removing a tag

See sample usage in the ember-osf dummy app:
https://github.com/abought/ember-osf/blob/19aee1f5ec2d6067b8fb9d80966994888d883335/tests/dummy/app/templates/nodes/detail/files/provider/file.hbs#L53-L53

ember-cli-build.js
```diff
@@ -77,6 +77,8 @@ module.exports = function(defaults) {
         development: 'bower_components/bootstrap-treeview/src/js/bootstrap-treeview.js',
         production: 'bower_components/bootstrap-treeview/dist/js/bootstrap-treeview.min.js'
     });
+    app.import('bower_components/jquery.tagsinput/src/jquery.tagsinput.js');
```
abought: Sadly, this is necessary for now. 

It's not automatic because we didn't want every data model consumer to bloat their apps with our bower dependencies. Thanks for adding this. :)

abought:
Done with review. As discussed on flowdock, the big blocker will be making sure that the tags widget is initialized with the correct arguments passed in.

caneruguz:
Closing to redo with conflict resolution and fully working version.

PR #14, opened 2016-08-08
Implement search page

terroni:
## Purpose

Add the search page for browsing and filtering preprints.

## Changes

To avoid merge conflicts and to adhere to how `ember-share` is structured, the search page is under the `discover` route. Metadata on preprints are pulled from SHARE's elasticsearch and are displayed in the newly designed search result element. Search results can be sorted by relevance and upload date. Additional styling has been added to `taxonomy-tree` to match wireframes (namely including checkboxes). 

**Work left to do:** 

- Load taxonomy tree from new taxonomy endpoint. The current `bootstrap-treeview` UI looks good (styled as it is in the wireframe), but is limited as far as dynamic loading of data goes (e.g. you can't add child nodes yet, at least not through standard methods). You'd have to recreate the tree every time you add subjects, which could work but would be a bit of a hack. I've looked into `ember-cli-jstree` which seems to have more functionality, but I wasn't able to get things working properly. The styling definitely needs to be adjusted if used (there's a Bootstrap theme you can plug in), but it conveniently includes a loading symbol and can fill data from ajax calls. 
- "Active filters" visibility is computed properly but needs to be dynamically added to as filters are added. A hard-coded string is currently being used just to show that it works.

## Side effects

None

## Ticket

https://openscience.atlassian.net/browse/PREP-82


app/components/search-result.js
```diff
@@ -0,0 +1,24 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+
+    numMaxChars: 400,
+    truncateDescription: true,
+    displayDescription: Ember.computed('result', 'truncateDescription', function() {
```
abought: There may not be a value named "result" for this computed to watch...?

terroni: I added a default value to make up for that. Caner will be adding a component of his own for the equivalent of what I'm calling `search-result`, so I'm not focusing too much on this component.

abought:
Per PR note, you can define a passthrough in mirage/config.js, which lets requests to specific URLs get handled normally. We do that so we can talk to localhost OSF for login:
https://github.com/abought/ember-preprints/blob/3977f463d3d588e9f66f678f12113e895744346f/mirage/config.js#L6-L6

app/adapters/application.js
```diff
@@ -0,0 +1,4 @@
+import JSONAPIAdapter from 'ember-data/adapters/json-api';
+
+export default JSONAPIAdapter.extend({
```
abought: Since this is the default adapter, the application adapter file and its associated unit tests can be removed. (this is basically what would happen if the file didn't exist at all)

app/controllers/discover.js
```diff
@@ -0,0 +1,170 @@
...
+            sort
+        };
+
+        return this.set('queryBody', queryBody);
+    },
+
+    togglePagination() {
+        // prev
+        if (this.get('page') > 1) {
+            this.set('showPrev', true);
```
abought: You might consider having the showPrev and showNext variables be computeds. That way, you don't need to manually change them.

There are even some convenience methods to reduce boilerplate. 
http://emberjs.com/api/classes/Ember.computed.html#method_gt

app/controllers/discover.js
```diff
@@ -0,0 +1,170 @@
...
+        },
+
+        sortBy(option) {
+            this.set('sortBy', option);
+            this.set('page', 1);
+            this.loadPage();
+        },
+
+        linkToAddPreprint() {
+            this.transitionToRoute('/preprints/add');
```
abought: This helper should probably use view names (dot separated), not URLs (with slashes).

app/router.js
```diff
@@ -7,16 +7,18 @@ const Router = Ember.Router.extend({
...
-    this.route('add-preprint', { path: '/preprints/add' });
+  this.route('add-preprint', { path: '/preprints/add' });
 
-    this.route('browse');
+  this.route('browse');
 
-    this.route('page-not-found', { path: '/*wildcard' });
+
+  this.route('page-not-found', { path: '/*wildcard' });
+  this.route('discover');
```
abought: Conventionally, the wildcard route appears last.

(some routing systems will return the view for the first pattern that matches, in which case it would match the 404 page before discover)

app/routes/discover.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+
+    model() {
+        return Ember.RSVP.hash({
+//            preprints: this.store.findAll('preprint'),
```
abought: In general we try to avoid returning a hash from the model hook.  Helpers like `link-to` expect to be given a single model, and the page ends up not rendering correctly.

app/templates/components/search-result.hbs
```diff
@@ -0,0 +1,21 @@
+<div>
+    <h4><a>{{result.title}}</a></h4>
+    {{#each result.contributors as |contributor|}}
+        <a>{{contributor.familyName}}, {{contributor.givenName}}</a>
+    {{/each}}
+    <p {{action "toggleShowDescription"}} style="cursor: pointer; cursor: hand;">{{displayDescription}}</p>
```
abought: Avoid clickable text elements where possible.

Use markup with semantic meaning: `button` for something that performs an action, and `a` for something that takes you to a new place.

That will also save you from the need to embed styles in markup (custom styles should be in the CSS file). If you use a button, the bootstrap css class `btn-link` may be useful.

app/templates/components/search-result.hbs
```diff
@@ -0,0 +1,21 @@
...
+
+    <div class="row">
+        <div class="col-sm-6 text-muted">
+            <p>Released on: {{moment-format result.date_updated "YYYY-MM-DD hh:mm a"}}</p>
```
abought: Rather than hard-coding date formats in every use of moment.js, we may want to consider defining a custom date format once, to help maintainability/ consistency.
https://github.com/stefanpenner/ember-moment#global-default-output-format

app/templates/components/search-result.hbs
```diff
@@ -0,0 +1,21 @@
...
+        </div>
+        <div class="col-sm-6">
+            {{#each result.tags as |tag|}}
+                <span class="label label-default">
```
abought: Purely for future reference, there is an ember-osf tags widget. (max is currently experimenting with it) Probably not an immediate priority for this PR though.

http://abought.github.io/ember-osf/classes/tags-widget.html

abought:
Added a few notes that may be helpful to you based on apparent sticky bits, but otherwise looks like things are coming along nicely.

app/routes/index.js
```diff
@@ -5,22 +5,22 @@ export default Ember.Route.extend({
...
-
             preprints: this.store.findAll('preprint'),
             subjects: this.store.findAll('taxonomy')
         });
     },
     actions: {
+        // TODO: properly transfer subject to discover route
         goToSubject(sub) {
-            this.transitionTo('browse', { queryParams: { subject: sub } });
+            this.transitionTo('discover', { queryParams: { subject: sub } });
```
caneruguz: Are you making this name change based on wireframes or mockups?

app/components/search-header.js
```diff
@@ -0,0 +1,4 @@
+import Ember from 'ember';
```
caneruguz: Is there a reason to separate these into components? I don't think search page itself is going to be reusable.

app/routes/index.js
terroni: This was just to emulate ember-share, which uses 'discover' instead of 'browse'. I also figured this would avoid potential merge conflicts along the way since I'd be starting from scratch in a separate route.

app/components/search-header.js
terroni: Again this was to emulate ember-share, but I agree that there isn't really much need for this

PR #17, opened 2016-08-09
Preprints data models for new backend

abought:
## Purpose
Add models that allow app to talk to the new preprints backend. Depends on field name and endpoint changes in this PR, which may not be merged yet as of writing: https://github.com/CenterForOpenScience/osf.io/pull/6118/files

[Can generate test data](https://github.com/CenterForOpenScience/osf.io/pull/6113) fast using script in OSF repo:
`python -m scripts.create_fakes -u erin@email.io --nprojects 3 --preprint True`

Will break some UI. Go team!

screen shot 2016-08-09 at 10 59 40 am
screen shot 2016-08-09 at 10 59 47 am

## TODO
- Define models and shared behaviors
  - [x] Authors
  - [x] Preprints
  - [x] Taxonomy (subject can be removed)- depends on OSF PR to rearrange endpoints
- [x] Verify the preprint data renders in templates (though changes will be needed)
- [x] Verify that any additional fields can/should be removed
- [x] Implement contributors/author behavior (may need to DRY synonyms and behaviors with ember-osf)
  - Dirty relationships tracking on node save: https://github.com/abought/ember-osf/blob/30965bb675d0c601e8a7521b9c481c461455e090/addon/models/node.js#L120-L120
  - NodeActionsMixin contrib mgmt actions need to be able to operate on a field called "authors" (it's possible that the alias will be a sufficient workaround here, but need to verify. @pattisdr may have some code that can help test this fast) https://github.com/abought/ember-osf/blob/6e72c6f6bec916eaa73be2912c70197a85b456a8/addon/mixins/node-actions.js#L121-L121
- ~~Swap out old preprints model for new one and update pages that were depending on mirage~~ (will let page authors make updates as relevant)
- [x] ~~Remove relevant mirage code for preprints dummy data~~ (will do in future PR; keep as reference for people converting pages over)
- Test various relationship fetching and behavior
  - UI decision: relationships fields respect pagination (eg only fetch first 10 results). How do we handle summaries with lots of authors? @caneruguz 

## Summary of changes
- Replace mirage models with real data for taxonomy and preprints
- Remove subjects model (now unused)
- Change pages so they load, and rely on others to get them converted correctly to new model structure

## Notes for reviewers
- Best to generate test data as described in 6113. 
- Templates have not been converted to work with new models because others are actively rewriting the templates. Will let them handle missing/renamed fields during the updates.

- According to @hmoco, "authors" are just another name for contributors. At present, we expect them to have all the same behavior as contributors.

## Ticket
https://openscience.atlassian.net/browse/PREP-97


app/models/preprint.js
```diff
@@ -1,17 +1,37 @@
...
-    license: DS.attr(),
-    path: DS.attr(),
+    // TODO: May be a relationship in the future pending APIv2 changes
+    subjects: DS.attr(),
+    provider: DS.attr('string'),
+    date_created: DS.attr('date'),
+    date_modified: DS.attr('date'),
+    abstract: DS.attr('string'),
     tags: DS.attr(),
-    doi: DS.attr(),
```
caneruguz: We still have a field for doi as a string entry that we will validate.

abought: Interesting! Where is the field defined for the API?

https://github.com/abought/osf.io/blob/99ac101be6a94fd5950ff9c0c317922ec1c12c6d/api/preprints/serializers.py#L28-L28

caneruguz:
@abought RE: "summaries with lots of authors? "
We should use the ```+ remaining author number``` after a number of  authors are listed. The number will certainly be under 10, I think about 5 is enough, OSF has 3. In any case we don't need to load beyond first page of contributors. 

```
   Gonzales, Bower, Michel, Hu, Kerim + 12 more
```

abought:
> In any case we don't need to load beyond first page of contributors.

Excellent! The meta field should provide info on total number of results. Works for us!

app/models/preprint.js
```diff
@@ -1,17 +1,37 @@
...
-    subject: DS.attr(),
-    abstract: DS.attr(),
-    publisher: DS.attr(),
-    project: DS.attr(),
-    supplementalMaterials: DS.attr(),
-    figures: DS.attr(),
-    license: DS.attr(),
-    path: DS.attr(),
+    // TODO: May be a relationship in the future pending APIv2 changes
+    subjects: DS.attr(),
```
pattisdr: Can you explain to me the difference between subjects and taxonomies?

abought: My understanding is that the taxonomy endpoint is a giant bucket of every possible subject name at any level of a nested hierarchy.  The "subjects" field reflects the child categories of that given taxonomy category.

However, this isn't currently reflected in the test data, so pinging @erinspace to verify.

app/routes/browse.js
```diff
@@ -12,7 +12,7 @@ export default Ember.Route.extend({
     model() {
         return Ember.RSVP.hash({
             preprints: this.store.findAll('preprint'),
-            taxonomy: this.store.find('taxonomy', 1)
+            taxonomy: this.store.findAll('taxonomy')
```
pattisdr: So the on the browse route we are returning `taxonomy: this.store.findAll('taxonomy')` and for the index route, `subjects: this.store.findAll('taxonomy')`. This is confusing.

abought: Good point, and Terron will be tackling this soon. 

Realistically we need to refactor this route once real data comes in; we should avoid RSVP.hash in the model hook as much as possible, and we'll need to filter by query params. The current changes were the minimum needed to make the page load at all.

app/models/preprint.js
hmoco: That is correct!

pattisdr:
I know this isn't part of your work here, but fetching preprints is just a single call to /preprints/ so only getting first ten, and preprints count is just then ten, instead of using the total meta count.
screen shot 2016-08-09 at 2 05 33 pm

abought:
Indeed. We should ticket to uodate pages for the new models; a variety of UI is kind of off kilter now. 

Since other authors are actively editing the affected pages I'm hesitant to do a rewrite. First goal is to just get working models in front of people.

app/models/taxonomy.js
```diff
@@ -1,5 +1,20 @@
...
+ */
+
+/**
+ * Model for OSF APIv2 preprints. This model may be used with one of several API endpoints. It may be queried directly. In the future, there will be multiple taxonomy endpoints under the same namespace.
+ * For field and usage information, see:
+ * https://api.osf.io/v2/docs/#!/v2/Plos_Taxonomy_GET
+ * @class Taxonomy
+ */
+export default OsfModel.extend({
+    type: DS.attr('string'),
```
pattisdr: Do we need type in the model?

pattisdr: Nevermind, yes we do.

pattisdr:
Other UI things, authors is fetching preprint.authors instead of preprint contributors, preprint.date instead of preprint.date_modified. Taxonomy-tree not filled out so am not seeing taxonomies.

pattisdr:
@abought - models look great, necessary serializers and adapters are correctly extending osf serializers and adapters.  Looks good to merge.

abought:
Yup, those gaps track with the screenshots I posted of breaking UI.... definitely page authors will be responsible for getting their pages working.

Does adding a contributor work with these models?

pattisdr:
This is for adding a contributor to a preprint correct?

abought:
That would be nice. I think that part of code may eneed tweaks. If you think the rest looks good, we can make the relationships tweaking happen in a followup pr.

pattisdr:
If people are waiting on these models it should happen in a follow-up PR - it will take me awhile to hook everything up to test this.

PR #19, opened 2016-08-10
Feature/Author Management [PREP-81]

pattisdr:
<h3> Ticket </h3>
https://openscience.atlassian.net/browse/PREP-81

<h3> Purpose </h3>

Adds author management for preprints.  
❗ Needs companions 
- https://github.com/CenterForOpenScience/ember-osf/pull/111, adds pagination, reorder contributors, addUnregisteredContributors, addContributor actions
- https://github.com/CenterForOpenScience/osf.io/pull/6130 creation of users via cookie auth
- https://github.com/CenterForOpenScience/osf.io/pull/6090 Fix premature check if contrib is admin
- https://github.com/CenterForOpenScience/osf.io/pull/6138 Reordering contributors via API
- Contributor save instead of node save or alternative, for error handling https://github.com/CenterForOpenScience/ember-osf/pull/112

<h3>  TODO  </h3>
- [x] List contributors
- [x] Add contributors
- [x] Search contributors (currently filtering on v2 users)
- [x] Paginate contributor search results (needs https://github.com/CenterForOpenScience/ember-osf/pull/111)
- [x] Add unregistered contributors
- [x] Update contributors (each modification sends immediate request rather than bulk updates)
- [x] Remove contributors
- [x] Popovers with explanations
- [x] Allow cookie auth to work for creating unregistered contributors https://github.com/CenterForOpenScience/osf.io/pull/6130
- [x] Template logic to minimize requests that will fail (instead of modals informing user that request is not allowed, or allowing request to go through and using growl box).  Disable buttons when clicking will result in failed request.
- [x] Sending emails for unregistered contributors (HOLD - can follow demo)
- [x] ~~Get rid of hardcoding node in route.~~
- [x] ~~Unregistered contributors added here are missing field used to determine permissions checking.  Because this field is missing, if an unregistered contributor is added in the demo, it is possible that some requests may be allowed to go through that would be prevented in the UI otherwise.~~
- [x] ~~Drag and drop (API v2 additions pending)~~
- [ ] Catching API v2 errors - some done

<h3> Current State</h3>
screen shot 2016-08-09 at 10 55 19 pm


app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            this.set('fullName', '');
+            this.set('email', '');
+        },
+        updateQuery(value) {
+            this.set('query', value);
         },
+        // Requests a particular page of user results
+        findContributors(page) {
+            var query = this.get('query');
+            var _this = this;
```
abought: What is the purpose of `_this`?

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
     permissionChanges: {},
     bibliographicChanges: {},
     valid: Ember.computed.alias('newContributorId'),
+    permissionToggle: false,
+    bibliographicToggle: false,
```
abought: Design question for @caneruguz : do we intend for preprints to expose the concept of "bibliographic" contributors? Does there need to be UI for that functionality?

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+    removalToggle: false,
+    stillAdmin: Ember.computed('isAdmin', function() {
+        return this.get('isAdmin');
+    }),
+    addState: 'emptyView', // There are 3 view states on left side of Authors panel. Default state just shows search bar.
```
abought: I'm a bit leery of magic string constants controlling this, as it makes refactoring tedious. Not a high priority if time is a pressure, but long term we should revisit our approach to showing several pages of behavior inside the same page.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+    query: null,
+    // Total contributor search results
+    totalSearchResults: Ember.computed('searchResults.[]', function() {
+        let searchResults = this.get('searchResults');
+        if (searchResults && searchResults.links) {
+            return Math.ceil(searchResults.links.meta.total);
```
abought: Why do we call `Math.ceil`? Isn't the total number of results already an integer?

app/components/preprint-form-authors.js
pattisdr: The reason I exposed this was because even if bibliographic info is hidden, the bibliographic rules still apply - enforced on the model.  So for example, you might not be able to remove a contributor because doing so would leave no bibliographic contributors.   But we wouldn't be even telling them that bibliographic contributors part is a constraint.  

So then you would be prevented from making a request here, but no way to change it from this view, or explanation of why this is happening. However users could manipulate this information via the OSF.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+        } else {
+            return;
+        }
+    }),
+    // Total pages of contributor search results
+    pages: Ember.computed('totalSearchResults', function() {
+        let totalSearchResults = this.get('totalSearchResults');
+        return Math.ceil(totalSearchResults / 10);
```
abought: Avoid hardcoding an assumption that pages always have 10 results.

Actually, we needed a field for "number of pages" elsewhere, so it should already be calculated and present under meta.total.

(The names are confusing. Sorry.)

https://github.com/abought/ember-osf/blob/7595c72f52b193d66fc2c00ffe5102a1c44ded22/addon/serializers/osf-serializer.js#L120-L120

app/components/preprint-form-authors.js
pattisdr: Good catch, that was a mistake copied over from pages, where I did want to use Math.ceil, for the number of pages.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+    }),
+    // Search results excluding users that are already contributors
+    newSearchResults: Ember.computed('searchResults.[]', 'contributors.[]', 'addState', function() {
+        let searchResults = this.get('searchResults');
+        let contributors = this.get('contributors');
+        let userIds = contributors.map((contrib) => contrib.id.split('-')[1]);
```
abought: This is such a common method on contributors that we have a helper method.

https://github.com/abought/ember-osf/blob/a845b69029431eed1ab380c19f2ac6b6320a28b3/addon/models/contributor.js#L20-L20

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+        return searchResults.filter((result) => !Ember.A(userIds).contains(result.id));
```
abought: As written, this re-creates an array in order to evaluate each individual result, which is a bit inefficient.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+    }),
     actions: {
-        addContributor(userId, permission, isBibliographic) {
-            this.sendAction('addContributor', userId, permission, isBibliographic);
+        // Adds contributor then redraws view - addition of contributor may change which update/remove contributor requests are permitted
+        addContributor(user) {
+            this.sendAction('addContributor', user.id, 'write', true);
```
abought: As noted elsewhere, we strongly prefer closure actions for this use case, in order to take advantage of error handling. If the request fails, we don't want the UI to get stuck in the wrong place. (in which toggling will turn things on or off in unexpected ways)

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            this.toggleProperty('permissionToggle');
+            this.toggleProperty('bibliographicToggle');
+            this.toggleProperty('removalToggle');
+        },
+        // Adds unregistered contributor, then clears form and switches back to search view.
+        // Should wait to transition until request has completed.
+        addUnregisteredContributor(fullName, email) {
+            let res = this.attrs.addUnregisteredContributor(fullName, email, 'write', true);
+            var _this = this;
```
abought: As noted elsewhere, `_this` is unnecessary. Arrow functions are more concise, and way more convenient! :)

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            res.then(function() {
+                _this.toggleProperty('bibliographicToggle');
+                _this.toggleProperty('permissionToggle');
+                _this.toggleProperty('removalToggle');
+            });
+            this.set('addState', 'searchView');
```
abought: You may want to put this stuff inside a single reset function to DRY future cleanup.

Additionally, just as an aside, there is a function called `setProperties` that can set multiple things at once. Not a big issue, but convenient.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            this.set('fullName', '');
+            this.set('email', '');
+        },
+        updateQuery(value) {
+            this.set('query', value);
```
abought: What advantage does this offer over, say, the `mut` template helper?

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            if (query) {
+                _this.sendAction('findContributors', query, page);
+                this.set('addState', 'searchView');
+            }
+        },
+        // Removes contributor then redraws contributor list view - removal of contributor may change
+        // which additional update/remove requests are permitted.
         removeContributor(contrib) {
             this.sendAction('removeContributor', contrib);
+            this.toggleProperty('removalToggle');
```
abought: See notes on closure actions and error handling elsewhere.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            this.toggleProperty('permissionToggle');
+            this.toggleProperty('bibliographicToggle');
+            this.removedSelfAsAdmin(contrib, contrib.get('permission'));
+            this.get('contributors').removeObject(contrib);
         },
-        permissionChange(contributor, permission) {
+        // Updates contributor then redraws contributor list view - updating contributor
+        // permissions may change which additional update/remove requests are permitted.
+        updatePermissions(contributor, permission) {
             this.set(`permissionChanges.${contributor.id}`, permission.toLowerCase());
```
abought: What is the advantage of this intermediate data structure?

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            this.sendAction(
```
abought: Closure actions, error handling... you know the drill, but tagging this where noted.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            this.set('permissionChanges', {});
+            this.toggleProperty('permissionToggle');
+            this.toggleProperty('removalToggle');
+            this.removedSelfAsAdmin(contributor, permission);
+
         },
-        bibliographicChange(contributor, isBibliographic) {
+        // Updates contributor then redraws contributor list view - updating contributor
+        // bibliographic info may change which additional update/remove requests are permitted.
+        updateBibliographic(contributor, isBibliographic) {
```
abought: See previous question about bibliographic contributor functionality.

abought: I will note that bibliographic contributors aren't mentioned in the wireframe for this page

https://openscience.mybalsamiq.com/projects/preprintjuly2016/Add_authors

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+        // There are 3 view states on left side of Authors panel.  This switches to searching contributor results view.
+        searchView() {
+            this.set('addState', 'searchView');
+            this.set('fullName', '');
+            this.set('email', '');
+        },
+        emptyView() {
+            this.set('addState', 'emptyView');
+            this.set('searchQuery', '');
+            Ember.$('.searchQuery')[0].value = '';
```
abought: We try to avoid manual jquery manipulation of the page for a variety of reasons, among them that it locks together markup/class names with controller logic.

If you must do so, consider using `this.$()` instead of `Ember.$`. As far as I can tell, the latter selects from the entire page (and any other things outside this component instance that use the same class). The former should select only things in or under the container for this component, which is more specific.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+                    '<dd><ul><li>Read privileges</li> ' +
+                        '<li>Add and configure preprint</li> ' +
+                        '<li>Add and edit content</li></ul></dd>' +
+                '<dt>Administrator</dt><dd><ul>' +
+                    '<li>Read and write privileges</li>' +
+                    '<li>Manage authors</li>' +
+                    '<li>Public-private settings</li></ul></dd>' +
+                '</dl>'
+        });
+        Ember.$('#bibliographic-popover').popover({
```
abought: See note above about `Emnber.$` vs `this.$`

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+        emptyView() {
+            this.set('addState', 'emptyView');
+            this.set('searchQuery', '');
+            Ember.$('.searchQuery')[0].value = '';
+        }
+    },
+    // Temporary popover content for authors page. Will need to find alternative to jquery selectors.
+    didInsertElement: function() {
+        Ember.$('#permissions-popover').popover({
+            content: '<dl>' +
```
abought: Can you add a FIXME or TODO? This is not something we want to lose track of.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            'A registered administrator is a user who has both confirmed their account and has administrator privileges.'
+        });
+    },
+    /**
+    * If user removes their own admin permissions, many things on the page must become
+    * disabled.  Changing the stillAdmin flag to false will remove many of the options
+    * on the page.
+    */
+    removedSelfAsAdmin(contributor, permission) {
+        if (this.get('currentUser').id === contributor.id.split('-')[1] && permission !== 'ADMIN') {
```
abought: See note above about helper methods and permissions constants.

app/components/preprint-form-authors.js
```diff
@@ -10,26 +10,158 @@ export default CpPanelBodyComponent.extend(PreprintFormFieldMixin, {
...
+            this.set('stillAdmin', false);
```
abought: We may want to consider how variables are named here... "still admin" is a bit of an odd concept on the page?

abought: Should there be a model helper method for whether a user is an admin on a project? It seems to be a question we ask a lot.

pattisdr: So what I mean by stillAdmin...when the page is loaded the user might be an admin. But if I change the permissions, it might no longer be an admin. The original isAdmin isn't updated, that was something I passed in as an argument.

app/components/unregistered-contributor-form.js
```diff
@@ -0,0 +1,45 @@
...
+       ]
+    },
+});
+
+export default Ember.Component.extend(Validations, {
+    fullName: null,
+    username: null,
+    isFormValid: Ember.computed.alias('validations.isValid'),
+    actions: {
+        emptyView() {
```
abought: Why does this action need to be passed in? And is the boilerplate here really necessary?

(can you just do something like `{{action emptyView}}`, using the passed in action name instead of a string reference? The more error handling you need in this component the less viable that is, of course)

app/components/validated-input.js
```diff
@@ -0,0 +1,44 @@
+// Copied from ember-cpi-validations/tests/dummy/app - should eventually be part of ember-osf but getting errors with dependencies
+
+// BEGIN-SNIPPET validated-input
```
abought: This seems odd. I did see some `v-get` errors in ember-osf, but:
1. Other validation fields still showed up and worked even if v-get didn't, and
2. Max is reporting that validation mostly works in his project. (not developing inside an addon)

I'd encourage us to make sure this is necessary in the actual use context, before including this in the final PR.

app/controllers/add-preprint.js
```diff
@@ -11,6 +12,13 @@ export default Ember.Controller.extend({
         uploadMultiple: false,
         method: 'PUT'
     },
+    canEdit: Ember.computed('isAdmin', 'isRegistration', function() {
+        return this.get('isAdmin') && !(this.get('model').get('registration'));
+    }),
+    isAdmin: Ember.computed(function() {
+        return this.get('model').get('currentUserPermissions').indexOf('admin') >= 0;
```
abought: Small syntax note- you can use `.get('model.currentUserPermissions')` to fetch nested values in a single get.

abought: Try to avoid hard-coded permissions strings in favor of the named constants.

app/controllers/add-preprint.js
```diff
@@ -11,6 +12,13 @@ export default Ember.Controller.extend({
         uploadMultiple: false,
         method: 'PUT'
     },
+    canEdit: Ember.computed('isAdmin', 'isRegistration', function() {
+        return this.get('isAdmin') && !(this.get('model').get('registration'));
```
abought: This seems like a question we ask a lot, and might make sense to add helper methods on the model class.

app/controllers/add-preprint.js
```diff
@@ -82,6 +90,45 @@ export default Ember.Controller.extend({
...
+         * findContributors method.  Queries APIv2 users endpoint on full_name.  Fetches specified page of results.
+         * TODO will eventually need to be changed to multifield query.
+         *
+         * @method findContributors
+         * @param {String} query ID of user that will be a contributor on the node
+         * @param {Integer} page Page number of results requested
+         * @return {Record} Returns specified page of user records matching full_name query
+         */
+        findContributors(query, page) {
+            var _this = this;
```
abought: `_this` should be unnecessary (use arrow functions)

app/controllers/add-preprint.js
```diff
@@ -82,6 +90,45 @@ export default Ember.Controller.extend({
...
+         * @return {Promise} Returns a promise that resolves to the updated node
+         * with the new contributor relationship.
+         */
+        addContributor(userId, permission, isBibliographic) {
+            var node = this.get('_node');
+            var contributor = this.store.createRecord('contributor', {
+                id: `${node.get('id')}-${userId}`,
+                permission: permission,
+                bibliographic: isBibliographic
+            });
```
abought: Can we reuse some of the logic from `NodeActionsMixin` via a super call? We should avoid copy-paste duplicate code if at all possible.

app/controllers/add-preprint.js
```diff
@@ -82,6 +90,45 @@ export default Ember.Controller.extend({
...
+            // For now, fetching contributors with loadAll method.  This requires some manual management of the contributors list.
+            node.get('contributors').pushObject(contributor);
+            this.get('contributors').pushObject(contributor);
```
abought: It appears that you add this user to the list of known contributors before the save operation occurs, much less before it succeeds. If the request fails, this could cause the page to have an inaccurate picture of reality: it thinks someone was added, but the server said they were not.

app/helpers/conditions-for-contrib-removal.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function conditionsForContribRemoval(params/*, hash*/) {
+    var contributorToRemove = params[0];
+    var contributors = params[1];
+    if (contributors) {
+        var minRegisteredAdmins = false;
+        var minBibliographic = false;
+        contributors.forEach(function(contributor) {
+            if (contributor.id !== contributorToRemove.id) {
+                if (contributor.get('permission') === 'admin' && contributor.get('unregisteredContributor') === null) {
```
abought: Use named permission constants wherever possible. (ember-osf/const/permissions).

app/helpers/conditions-for-contrib-removal.js
```diff
@@ -0,0 +1,31 @@
...
+                    minRegisteredAdmins = true;
+                }
+                if (contributor.get('bibliographic')) {
+                    minBibliographic = true;
+                }
+            }
+        });
+        return minRegisteredAdmins && minBibliographic;
+    } else {
+        return params;
```
abought: This function returns either a boolean, or an array? Is that the intended behavior?

app/controllers/add-preprint.js
pattisdr: I wanted the contributor from the super call.  The super call returns a promise that resolves to a node?

app/helpers/min-admins.js
```diff
@@ -0,0 +1,29 @@
+import Ember from 'ember';
+
+/**
+ * minAdmins helper - used to determine if the user should be able to update permissions
+ * of a particular contributor.  False if user is trying to update the permissions
+ * of the sole admin.  All projects need at least one administrator.
+ *
```
abought: Thanks for adding documentation! But note that a few more fields are needed for this to work with YUIDoc, Documenting the parameters and return types are useful, but excluding the method name causes it not to appear in auto-generated docs at all.

(YUIDoc doesn't really try to understand code; it just strings together docs based on docstrings.)

app/helpers/min-admins.js
```diff
@@ -0,0 +1,29 @@
...
+    var contrib = params[0];
+    var contributors = params[1];
+    if (contributors) {
+        var registeredAdmins = 0;
+        contributors.forEach(function(contributor) {
+            if (contributor.get('permission') === 'admin' && contributor.get('unregisteredContributor') === null) {
+                registeredAdmins++;
+            }
+        });
+        if (registeredAdmins === 1 && contrib.get('permission') === 'admin' && contrib.get('unregisteredContributor') === null) {
```
abought: See notes above re: named constants, and ambiguous return type (boolean vs array) for this function.

app/helpers/min-bibliographic.js
```diff
@@ -0,0 +1,29 @@
+import Ember from 'ember';
```
abought: See requirements question  re: bibliographic contribs elsewhere in this PR.

app/helpers/conditions-for-contrib-removal.js
pattisdr: It seems like this helper is called when the page is loaded the first time but I don't have any contributors loaded yet.  I just returned the default params if that is the case.

app/helpers/permission-map.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+
+/**
+* permissionMap helper.  Maps short form of the contributor's permissions
+* passed in params to the long form.
+*
+*/
+export function permissionMap(params/*, hash*/) {
+    var map = {
```
abought: There are already named constants in ember-osf that might be useful. We'd like to centralize this sort of thing where possible.
https://github.com/abought/ember-osf/blob/14afd9a7974573d04cab3efc379a3504218a5942/addon/const/permissions.js#L12-L12

The advantage of putting the constants elsewhere is that they can be used for other purposes as well, like rendering a select box/ dropdown for permissions.

app/helpers/permission-to-remove-contributor.js
```diff
@@ -0,0 +1,23 @@
...
+/**
+* permissionToRemoveContributor helper.  Checks to see if user has proper permissions
+* to remove contributor.  The user either must be an admin or trying to remove herself.
+* The project cannot be a registration.
+*/
+export function permissionToRemoveContributor(params/*, hash*/) {
+    var contributor = params[0];
+    var currentUser = params[1];
+    var stillAdmin = params[2];
+    var node = params[3];
```
abought: Not important and tangential syntax note:

[Destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) is a nice concise way to do this in ES6.

app/helpers/permission-to-remove-contributor.js
```diff
@@ -0,0 +1,23 @@
...
+    if (currentUser) {
+        var currentUserId = currentUser.get('currentUserId') || currentUser.get('id');
+        var removeSelf = contributor.id.split('-')[1] === currentUserId;
```
abought: See note elsewhere about helpers. Just the fact that you're doing this so many times in your PR is a good sign that it makes sense to centralize the common logic. :)

(One rough rule of thumb: if you see it twice, consider whether a helper would be useful; three or more times, think about refactoring)

app/styles/app.scss
```diff
@@ -556,12 +558,13 @@ hr {
             opacity: 1;
         }
     }
-    .row {
-        margin: 0;
-    }
-    label, input, textarea {
-        display: block;
-    }
+    // THESE STYLES ARE INTERFERING WITH MY STYLES.
```
abought: I'm just going to go out on a limb and assume this will be resolved before merging. For example, by using more distinctive style names, namespaced to your specific component.

app/styles/authors.scss
```diff
@@ -0,0 +1,61 @@
...
+.col-division-left {
+    padding-left: 20px;
+    border-left: 1px solid #ccc;
+}
+
+.col-division-right {
+    border-right: 1px solid #ccc;
+}
+
+.search-box {
```
abought: In general, we encourage components to use more distinctive style names, to avoid interfering with other components. Eg https://github.com/abought/ember-osf/blob/1c3fa17675dcb2d5ad5b4fa6237da312d44242d2/addon/components/file-browser/style.scss#L1-L1

app/templates/components/preprint-form-authors.hbs
```diff
@@ -1,56 +1,142 @@
...
+                                <i class="glyphicon glyphicon-search"></i>
+                            </button>
+                        </span>
+                    </div>
+                </form>
+            </div>
+        </div>
+        <div class="row">
+            <div class="col-xs-11">
+                {{#if (not (eq addState 'emptyView'))}}
```
abought: There is a `not-eq` helper...

app/templates/components/preprint-form-authors.hbs
```diff
@@ -1,56 +1,142 @@
...
+                                        </td>
+                                        <td> <button class="btn btn-success btn-small" {{action "addContributor" result}}> Add </button> </td>
+                                    </tr>
+                                {{/each}}
+                            </table>
+                        {{else}}
+                            No results found.
+                        {{/if}}
+                        {{#if (gt pages 1)}}
+                            {{osf-paginator totalSearchResults=totalSearchResults maxPages=5 fetchResults=findContributors query=query}}
```
abought: Why are we restricting the number of pages?

app/templates/components/preprint-form-authors.hbs
```diff
@@ -1,56 +1,142 @@
...
+                        {{/if}}
+                    {{/if}}
+                    {{#if (eq addState 'unregisteredView')}}
+                        <h3> Add Unregistered User </h3>
+                        <div class="user-validation">
+                            {{unregistered-contributor-form emptyView=(action 'emptyView') addUnregisteredContributor=(action 'addUnregisteredContributor')}}
```
abought: General comment: try to use standard nomenclature for method names where possible, eg `reset` if this is a reset button. I'm not entirely sure what `emptyView` is emptying, but it implies that the component isn't very isolated. (who does the component need to handle emptying the parent? An alternative data flow would be to have the component get a save action, and the save action- passed down to the component- would define any cleanup the parent wanted to do in the buggle world outside the component)

app/templates/components/preprint-form-authors.hbs
```diff
@@ -1,56 +1,142 @@
...
+                </tr>
+                {{#each contributors as |contrib|}}
+                    <tr class="contributor-row">
+                        <td> <span class="fa fa-bars sortable-bars"></span> </td>
+                        <td>
+                            <img class="m-l-xs" src={{contrib.users.links.profile_image}} height=40 width=40>
+                            {{#if contrib.unregisteredContributor}}
+                                {{contrib.unregisteredContributor}}
+                            {{else}}
+                                <a href={{contrib.users.links.html}}> {{contrib.users.fullName}} </a>
```
abought: Line 107 and here: use helper methods as possible. Accessing to model fields should not depend on the internal structure of the API JSON response; these helper methods provide a consistent and abstracted way to get that value.

 https://github.com/abought/ember-osf/blob/ea1462dfd277155da6a3d803075fd7e375bb8638/addon/models/user.js#L37-L37

app/templates/components/preprint-form-authors.hbs
```diff
@@ -1,56 +1,142 @@
...
+                            {{/if}}
+                        </td>
+                        <td class="vert-align">
+                            <div class="form-group">
+                                {{#if (and stillAdmin (and canEdit (minAdmins contrib contributors permissionToggle)))}}
+                                    <select class="form-control" onchange={{action (action 'updatePermissions' contrib) value='target.value'}}>
+                                        <option value=READ selected={{eq contrib.permission 'read'}}> Read </option>
```
abought: In the osf contributors modal, I have an example of using the ember-osf "permissionsSelector" text to render this select box, using logic that only hardcodes permissions names in one place.

app/templates/components/preprint-form-authors.hbs
pattisdr: Mistake: the maxPages doesn't do anything anymore.

app/templates/components/preprint-form-authors.hbs
```diff
@@ -1,56 +1,142 @@
...
+                            <img class="m-l-xs" src={{contrib.users.links.profile_image}} height=40 width=40>
+                            {{#if contrib.unregisteredContributor}}
+                                {{contrib.unregisteredContributor}}
+                            {{else}}
+                                <a href={{contrib.users.links.html}}> {{contrib.users.fullName}} </a>
+                            {{/if}}
+                        </td>
+                        <td class="vert-align">
+                            <div class="form-group">
+                                {{#if (and stillAdmin (and canEdit (minAdmins contrib contributors permissionToggle)))}}
```
abought: This won't be a blocker for the demo, but this logic seems very convoluted and hard to follow.

While I'm not a fan of the extreme way ember locks down logic in templates, this might be a smidge much....

app/templates/components/validated-input.hbs
```diff
@@ -0,0 +1,24 @@
+{{!-- Copied from ember-cpi-validations/tests/dummy/app - should eventually be part of ember-osf but getting errors with dependencies --}}
```
abought: See note above; it's a bit surprising to me that we need to copy this code over. If we can avoid it, that would be strongly preferred.

abought:
Pass complete. If we have to prioritize for the demo, error handling will be the most visible thing in the demo. 

Various other notes about duplication and logic flow added.

app/templates/components/validated-input.hbs
pattisdr: I copied this from the dummy app?  I didn't think this was part of the package.

app/components/preprint-form-authors.js
```diff
@@ -1,7 +1,7 @@
 import Ember from 'ember';
```
abought: Please use pod-style components when possible; this will make it easier to add/remove components later as needed.

(similar to the file/folder structure in the components directory of ember-osf)

`ember g component something --pod`

Renaming the existing files to match should also work.

pattisdr: I thought @caneruguz didn't want us to use pod-styles.

app/components/preprint-form-authors.js
pattisdr: Do you have any alternative ideas?

app/components/preprint-form-authors.js
caneruguz: That's correct, I told @pattisdr to not use pod style for a project with a low number of components finding and deleting component files is not so much of a hassle and for me having several component.js file is more of a headache. But I wouldn't fight this if using pods is a better standard and beyond a matter of opinion.

app/components/preprint-form-authors.js
caneruguz: We intended this to be a feature eventually to allow researchers to add people who can administer by not be shown as authors, I think since it fits better with our existing models we should implement it here. A UI challenge would be to clarify if clarification is necessary about why authors added would not be bibliographic.

app/styles/app.scss
caneruguz: @pattisdr you can keep these commented out, it would be very hard for you to fix it. I will go through and fix styles, we shouldn't be using such general styles across the app. Let's merge this in first.

app/components/preprint-form-authors.js
abought: Hmm. There are a few considerations here. So far, we've been using pods in ember-osf and osf.io apps, according to the thinking below:

- We should probably be consistent across projects to avoid fragmentation. Ember-osf, osf.io, and ember-share all use pod style at the moment, for reference.

- The "classic" structure puts related files in too many different places, making it hard to clean up (or to see everything required for the component to work). At a minimum, styles should be added to their own css file, not app.css. (I think it's reasonable to expect that we will refactor as requirements evolve; we should set a standard that minimizes unused files)
  - Dawn does this well. Some of our other PRs present, less so. https://github.com/CenterForOpenScience/ember-preprints/pull/20#discussion-diff-74318401R751

- Pods do have some major drawbacks; we had to do some hoop jumping to make ember-osf recognize style files outside the `styles/` folder. It's genuinely annoying, but a solvable problem.

- Pod-like behavior is the direction ember is headed, though not pods themselves. The official long-term picture is probably the unfinished [module unification](https://github.com/emberjs/rfcs/pull/143) RFC. They specifically call out the "titlebar problem" of confusing filenames. :)

-

app/components/preprint-form-authors.js
pattisdr: I do not know how to use this - I will look into it.

app/components/preprint-form-authors.js
abought: Honestly, the real issue is that we're presenting several pages inside a single route, which is a bit different from ember's core assumptions about how route is tied to the state of the application.

I'd need to experiment a bit more with the routing mechanisms to see what we could do, and just haven't had the bandwidth to explore these sorts of questions. I still think the notion of nested routing (a `/search/` child route) has potential, but there are questions about what happens when you change things in the child route that we don't know enough about yet. 

Ember workarounds that show three "pages" in one route are very common- the nested router isn't a solution to everything. But I think there's potential we haven't tapped yet.

Somewhere in the back of my mind is the tabbed diagram here:
http://fromrailstoember.com/9-nested-routes-equals-nested-ui/

abought: tl;dr: This is not a concern for this PR, but it takes time for alternate ideas to percolate so we can use them to approach future problems. If you had to triage, I'd still put user-facing stuff first.

app/components/preprint-form-authors.js
abought: Makes sense. Just confirming there's a use case. :)

app/components/preprint-form-authors.js
abought: Right, it's kind of weird to me that there are two sources of authority on whether the user is an admin. That may mean that an action isn't updating the state in whatever is passed in- this could point to an underlying issue with how data / notification of changes flows throughout the different levels of the page.

app/controllers/add-preprint.js
abought: Do any of the current usages  of the mixin method depend on getting the node back? (in our osf.io branch, eosf, or elsewhere)

It's possible that the solution is for `addContributor` to return a contributor (after node save succeeds). We probably just returned the first promise we had handy. As we use the method, keep an open mind- we may have opportunities to improve how things work. Duplicating code- especially for custom behaviors that we may refactor- is more of a last resort.

app/helpers/conditions-for-contrib-removal.js
abought: Hmm. The downside is that even an empty array evaluates as truthy- in the absence of data, the page is still making decisions based on this result. That may not be what was intended.

Is there no data at all, or just no useful default?

`if([]){console.log('whee')}`

app/styles/app.scss
abought: Agreed. As long as there's a path forward, sounds good by me.

In other addons, we try to namespace our styles to make them less likely to collide with other components. This will be an area to ticket for future improvements.

app/templates/components/validated-input.hbs
abought: Ah. Maxwell also copied this from the dummy app. Whee!

abought: That said, some of the stuff I: read implied that v-get broke because I was in an addon. If you haven't tried it in this app, that's worth a look before just commenting it out.

app/components/preprint-form-authors.js
abought: It's useful for marking when you want to change a value; you can even specify the value something will change to. We use it in the addon modal select box for permissions, etc.

It's nice to know about for future development, but again, I know we'll need to triage the big stuff and merge soon. This is rather a ways down the list.

app/components/preprint-form-authors.js
pattisdr: Yes nested routes are worth looking into.  I will put this on the backburner for now though.

app/components/preprint-form-authors.js
pattisdr: Good catch, I don't think this is needed here.  It was being used before when we were doing bulk updates of contributors.

app/components/preprint-form-authors.js
pattisdr: Yes, day one when I started this I looked at the wireframe and saw problems with omitting bibliographic information and pointed it out to Caner.  I failed to update the wireframe, apologies.

app/components/preprint-form-authors.js
pattisdr: I just wasn't sure how to access the value of the search box here.  When I use an input helper I can get the value of it easily but if I don't use a helper I can't figure out how to access it without jquery.

app/components/unregistered-contributor-form.js
pattisdr: You're right I don't think I need it here - I'm just learning that you don't have to pass in the actions at every level.  I used to do sendAction at every level of every nested component.  It's pretty cool that you can just pass the action name.

app/components/validated-input.js
pattisdr: @abought I am not sure I understand.   I don't think this is part of the ember-cp-validations input, I borrowed this from the ember-cp-validations dummy app.  So did max, he also copied the whole thing out.  Max is doing almost the exact same thing I'm doing - ours our both working.  We are going to have very similar validated-input components though so need to come up with a way to consolidate.

app/controllers/add-preprint.js
pattisdr: If I return a contributor, this is not a promise, correct?  Because we don't want to call contributors.save(), this sends a POST request to nodes/node_id/contributors.

app/helpers/permission-to-remove-contributor.js
pattisdr: Yes, thanks, this was a WIP PR and I hadn't gotten to cleaning this up yet.   This was on my list.  All the same, you've pointed out several things in this PR that I hadn't noticed so I appreciate the thorough review.

pattisdr:
I've addressed about 2/3's of comments so far - pausing here.

app/controllers/add-preprint.js
abought: I'm thinking something like line 129 below, but done in the parent method: returning the newly contributed contributor in the .then after the `node.save` promise resolves.

app/templates/components/validated-input.hbs
pattisdr: You're right Andy - it seems to work in this context.  So now my validated-input component is working with no modifications from the original copied from the ember-cp-validations dummy app.  Hopefully Max and I can share the validated-input component without having to duplicate.  I see he made some other modifications to his.

abought: Yup, just looking out for the merge conflicts. :) 

Since this is a dummy app helper, there's probably no other way to pull this in, and duplicating code from the addon is fine in that case. If you find the component to be that useful, someday you might consider sending a PR to the authors making the component available by default.

app/templates/components/preprint-form-authors.hbs
pattisdr: I had just named it the empty view.  There are three possible views here.  This just switched it to the empty view. You're right - it's poorly named.

app/helpers/conditions-for-contrib-removal.js
pattisdr: ok - so I was definitely getting errors before in my console about how certain params couldn't be found, hence the check in all my helpers if the params were there first.  But I am not seeing that anymore so I removed the return of the array data type.

app/controllers/add-preprint.js
pattisdr: Yep this seems to work, changed in https://github.com/CenterForOpenScience/ember-osf/pull/111

abought: Thanks! Standard confirmation question: do any other uses of this function (including in stalled OSF prs) depend on the old return value?

app/components/preprint-form-authors.js
caneruguz: So what I'm hearing is that there are major drawbacks to this that effect code, not to mention the continuous confusion of looking at a lot of component.js tabs during development while the benefits are 1 - consistency, we are already doing it elsewhere, 2 - we might forget to remove files when components are no longer needed, 3 - ember is moving towards something like pods. 

I'm genuinely curious about what benefits for code quality makes using pods worthwhile to have roundabout ways to solve issues like recognizing styles outside styles/ folder for instance. Sam might have some more insights on this. I'll ask him to join in.

I think the project scope is small enough, and components few enough that we can restructure this later.

abought: At minimum, Dawn's pattern of putting styles in a separate file (not app.scss) is highly valuable and recommended. The rest... well, I did include downsides for a reason. Neither option is amazing. :)

samchrisinger: With regard to **finding/deleting components**: We should be using the ember-cli to generate stubs for all components. `ember g` and the inverse `ember d` should make this a complete non-issue. My experience is that between my git tooling and the ember-cli cleaning up after myself is really easy

I don't use a tabbed editor (or at the very least don't pay attention to tab names), so I can't really relate to difficulty finding component files. Actually pod structure does make it a bit faster for me to work on a component's controller/template logic, but the difference here is pretty negligible. 

> Pods do have some major drawbacks; we had to do some hoop jumping to make ember-osf recognize style files outside the styles/ folder. It's genuinely annoying, but a solvable problem.

IMO this is an unanswered question for _Ember in general_, i.e. totally indep. of code structure. There's no difficulty using structure for components and still packaging component styles in `/app/styles/components/`. But like Andy suggests, we absolutely should keep component styles in separate files. 

Sorry there is nothing very definitive here. I do prefer pod styles-- the semantics of bundling controller/template/style logic in a single directory really appeals to me. There is a lot to be said for consistency in the codebase, and I do think the Ember community is moving this direction (favoring pod structure). But I can't say one way is right or wrong.

app/templates/components/validated-input.hbs
ml3ha: @abought I did change the validated-input.hbs file a bit since there were is a case where a `textarea` had to be validated and not just an `input`. So I just added an extra condition that can be specified when creating the component.

app/styles/app.scss
el1t: I wouldn't say these styles are general _across the app_, as they are nested under `.preprint-form-section`. Though I agree they could be more specific.

app/styles/authors.scss
el1t: Couldn't these be namespaced under `#preprint-form-authors`?

app/styles/authors.scss
```diff
@@ -0,0 +1,78 @@
...
+
+.author-header {
+    padding-top: 15px;
+}
+
+.sortable-item {
+  transition: all .125s;
+}
+
+.sortable-item.is-dragging {
```
el1t: This can be nested as follows: (also avoiding `transition: all` as it's expensive)
```scss
.sortable-item {
    transition: background-color 125ms;
    &.is-dragging {
        transition-duration: 0;
        background: white;
        z-index: 10;
    }
}
```

app/styles/authors.scss
abought: That's one way to do it. Though that's not so much namespacing, as making sure that your CSS selector only applies in certain circumstances, based on the markup of other elements around where this is used. 

There's a few approaches to organizing CSS and naming things, by some pretty clever people- you wouldn't think it's a subject for study, but it really is! So several valid approaches, but calling it `search-box` as a one-off class is just a smidge too far on the generic end.

pattisdr:
Here are remaining issues I am stuck on:

- Main issue: When requests are failing for adding/updating contributors, cannot catch error (removing, searching working fine).
- mut template helper for update query https://github.com/CenterForOpenScience/ember-preprints/pull/19#discussion_r74293362
- stillAdmin variable (used when user removes their own admin permissions).  Having a hard time updating isAdmin variable https://github.com/CenterForOpenScience/ember-preprints/pull/19#discussion_r74296462
- canEdit helper method on node? When moved helper to node, was getting errors not having constructor on node https://github.com/CenterForOpenScience/ember-preprints/pull/19#discussion_r74306027
- Some tests are failing -  undefined is not a constructor (evaluating 'contributor.get('permission')')

abought:
I can take a look at these issues, if that would help.

- As we discussed, we know that not everything will go in this PR. Some comments are thoughts for the future, and totally ok to omit from this PR:
  - canEdit helper method will be useful eventually, but not required for now. I'd have to look at the constructor error to see exactly where the problem was coming from.
  - Mut helper is mentioned as a nice trick, but if boilerplate works, we can and should move ahead with as is.

For the others, we can take a look together. It's possible, for example, that your test factories aren't defining contributors for nodes by default, in which case you might get errors for contributors being undefined.

Per error handling, it depends how the actions are called and what they return. Again, would look on a case-by-case basis.

app/styles/authors.scss
pattisdr: Thanks for the tip @el1t.

abought:
Verifying related required PRs have been merged
- [x] CenterForOpenScience/ember-osf#111
- [x] CenterForOpenScience/osf.io#6130 creation of users via cookie auth
- [x] CenterForOpenScience/osf.io#6090 Fix premature check if contrib is admin
- [x] CenterForOpenScience/osf.io#6138 Reordering contributors via API
- [ ] Contributor save instead of node save or alternative, for error handling CenterForOpenScience/ember-osf#112

The last one will be left open pending further adapter explorations. (mostly blocked on time)

abought:
Basic functionality is there; merging to support integration of add/edit page.

One item discussed for the future- it would be great to offer context-sensitive feedback when a user is added. Currently, adding/editing a user doesn't display any indication that data was sent to the server and the operation succeeded. @caneruguz may have some thoughts on the UI, and whether toast vs a more context-specific label would be best here.

abought:
Note to self/others:

There is a hard-coded node specific to Dawn's local OSF. You may get errors on a different machine until you change this line:
https://github.com/abought/ember-preprints/blob/d1924777cd64c8921be3f6cecb0dcade21317eaa/app/routes/add-preprint.js#L8-L8

This will be taken care of in data flow improvements for the page

PR #20, opened 2016-08-10
Feature/form validations

ml3ha:
Add basic form validations for preprint form basics using ember-cp-validations.

These validations are:

- Title: required, maximum length validator of 300 characters
- Abstract: required, maximum length validator of 5000 characters and min of 20 characters (follows arXiv's style)
- DOI: optional, but if entered, uses a regex to check if the DOI is valid

Fixed a lot of things:

- Now validation error message only shows up after user focuses on the input field.
- There is a green success highlighting of the input fields after the user's input is valid.


app/styles/app.scss
```diff
@@ -748,3 +748,18 @@ hr {
     background-size: cover;
     background-image: url("img/logo_ua.png");
 }
+
+.validated-input .input-error {
```
abought: If possible, don't add component styles to app.scss.

We prefer "pod style components": they should live in a folder `componentName`, with two files:
`component.js`
`template.hbs`

See example: https://github.com/abought/ember-osf/tree/4f2947e7a0c3622b8f1c21d89e321caadf919b4e/addon/components/file-browser

If you have custom styles, then to keep the component together in a single concise package, we prefer those go in the same folder, in a file called `style.scss`. You can generate a new pod component at any time using a command line flag like so:
`ember generate component myComponentIsNamedBob --pod`

This makes it easier to move and remove components, so that we minimize leftover cruft lying around in other folders. For now, just renaming the folder and files would be sufficient to make it magically work.

abought:
This PR has merge conflicts that will need to be addressed before it could be considered for inclusion.

ml3ha:
Resolved merge conflicts

app/components/validated-input/component.js
```diff
@@ -0,0 +1,43 @@
+import Ember from 'ember';
+const {
+    isEmpty,
+    computed,
+    defineProperty,
+} = Ember;
+
+import Validations from '../../validators/preprint-form-validator';
```
abought: Why does your general component (display logic for data) import a validator for one very specific use case?

Shouldn't the model you pass in be responsible for telling it what fields are valid?

app/components/validated-input/component.js
```diff
@@ -0,0 +1,43 @@
+import Ember from 'ember';
```
abought: Is this all new code, or heavily based on something derived from elsewhere?

In the latter case, it's helpful to at *least* link to the original at the top of the file. The even more polite course of action, long term, would be to add a message to this repo saying you used someone else's code, similar to the convention of the `NOTICE` file in one of our other projects:

https://github.com/CenterForOpenScience/osf.io/blob/develop/NOTICE

Most of the stuff we rely on is open source, but every now and then the code we want to use has restrictions. Not a big blocker here, but checking licenses before including things can keep your future employers out of trouble.

app/validators/preprint-form-validator.js
```diff
@@ -0,0 +1,38 @@
+import { validator, buildValidations } from 'ember-cp-validations';
+
+const Validations = buildValidations({
```
abought: Out of curiosity, why does this need to be defined in a separate file? It doesn't seem like something we'd especially need to be reusable.

app/validators/preprint-form-validator.js
```diff
@@ -0,0 +1,38 @@
...
+    title: {
+        description: 'Title',
+        validators: [
+            validator('presence', true),
+            validator('length', {
+                // minimum length for title?
+                max: 300,
```
abought: Out of curiosity how did you arrive at this number (and why)?

For comparison, I looked in the OSF source code to see what the underlying field allows. There, the requirements are `1<=title<=200` characters:

https://github.com/abought/osf.io/blob/dfd3d4e332fcc5534fa385515f288b6848b9a5e6/website/project/model.py#L731-L731

app/validators/preprint-form-validator.js
```diff
@@ -0,0 +1,38 @@
...
+        ]
+    },
+    abstract: {
+        description: 'Abstract',
+        validators: [
+            validator('presence', true),
+            validator('length', {
+                // currently min of 20 characters -- this is what arXiv has as the minimum length of an abstract
+                min: 20,
+                max: 5000
```
abought: I'm actually not sure that OSF descriptions have a maximum length at all. It's possible the preprint service will have different design restrictions, though.

abought:
Just a few small comments to address, then this will be cleared for merge. Should be pretty quick to do in the AM.

1. Don't bake validations into the general validation component
2. Make a note of licensing terms where appropriate
3. The validation rules here are different from the OSF. If they are driven by our requirements that's ok, but if they're numbers picked out of a hit, please match them to rules used elsewhere.

abought:
Also, alas, travis is still failing. You can run tests locally using:

`npm run check-style && ember test`

app/validators/preprint-form-validator.js
ml3ha: No specific reason -- I thought I had seen 300 characters for maximum length for a title in other websites. However I can change this to 200 if it matches with the OSF's style.

app/components/validated-input/component.js
ml3ha: Yes, this is a mistake. I originally thought I had to extend the validations in the component in order for the component to work. This will be removed

app/validators/preprint-form-validator.js
ml3ha: I defined it in a separate file so it could be used in the controller.

app/validators/preprint-form-validator.js
abought: In this case, please do. Front-end validation is a way of giving the user immediate feedback for better user experience- we don't want to give the user false confidence if the backend wouldn't accept that value.

app/components/validated-input/component.js
ml3ha: For now I've just included a link to the ember-cp-validations github repo to the top of the file.

app/validators/preprint-form-validator.js
abought: Right, but couldn't you define it at the top of the controller instead? (instead of importing it)

ml3ha: That is true. For some reason when I was writing the validations I thought it would be useful to have it in a separate file in case I needed to import it in multiple places (which is what I had originally, but then realized it only needed to be used in the controller). I will delete the `preprint-form-validator.js` file and consolidate the validations into the controller.

app/components/validated-input/component.js
abought: Thanks. I created a followup ticket:
https://openscience.atlassian.net/browse/PREP-110

app/routes/add-preprint.js
```diff
@@ -1,4 +1,8 @@
 import Ember from 'ember';
 
 export default Ember.Route.extend({
+    model() {
```
abought: Please remove this- it looks to be dummy code.

abought:
Just one thing to tweak and then good to go.

app/controllers/add-preprint.js
```diff
@@ -9,7 +10,43 @@ export const State = Object.freeze(new Ember.Object({
...
+                max: 200,
+            })
+        ]
+    },
+    abstract: {
+        description: 'Abstract',
+        validators: [
+            validator('presence', true),
+            validator('length', {
+                // currently min of 20 characters -- this is what arXiv has as the minimum length of an abstract
```
abought: Nice research of similar products in the space- benchmarking similar services will help us build something tailored for preprints. Thanks for checking!

PR #22, opened 2016-08-11
View page improvements/fixes

geeksnglitter:
## Purpose
Fix some view page problems!
Page was loading very slowly because it was waiting for multiple promises to resolve.
There were a a lot of links that did not quite work.

## Changes
- [x] Remove RSVP hash - use setupController
- [x] Make download links work
- [x] Make contributors actually display

## Ticket


app/components/supplementary-file-browser.js
```diff
@@ -6,11 +6,21 @@ export default Ember.Component.extend({
...
+    showArrows: Ember.computed('numShowing', 'supplementList.[]', function() {
+        return (this.get('numShowing') < this.get('files').length);
+    }),
     files: Ember.computed('fileList', 'fileList.[]', 'primaryFile', function() {
         //Returns the list with primaryFile moved to the front
         let files = this.get('fileList');
 
+        //TODO: Grab get primary file from preprint object
+        if (files) {
+            const fakePrimary = files.get('firstObject');
```
caneruguz: Why fake? This is showing actual file in the project. Should we change the variable name?

app/components/supplementary-file-browser.js
```diff
@@ -6,11 +6,21 @@ export default Ember.Component.extend({
...
+            this.set('primaryFile', fakePrimary);
+            this.set('selectedFile', fakePrimary);
```
caneruguz: You can set multiple properties at once 
http://emberjs.com/api/classes/Ember.Component.html#method_setProperties

app/components/supplementary-file-browser.js
```diff
@@ -6,11 +6,21 @@ export default Ember.Component.extend({
...
+        }
+        //TODO: Remove above lines. They are temporary. Pass real primary to component
```
caneruguz: Is this because primary file is not available from the api endpoint?  Shouldn't that be ready?
 @erinspace

app/templates/components/supplementary-file-browser.hbs
```diff
@@ -5,9 +5,11 @@
 {{#if supplementList}}
     <h5>{{selectedFile.name}}</h5>
     <section>
-        <button class="show-more-button text-muted left" {{action 'moveLeft'}}>
-            <i class="fa fa-caret-left"></i>
-        </button>
+        {{#if showArrows}}
+            <button class="show-more-button text-muted left" {{action 'moveLeft'}}>
```
caneruguz: Let's not use text-muted to change the style. It should be applied to items that need to be muted because of being disabled or something similar.

app/templates/components/supplementary-file-browser.hbs
```diff
@@ -5,9 +5,11 @@
 {{#if supplementList}}
     <h5>{{selectedFile.name}}</h5>
     <section>
-        <button class="show-more-button text-muted left" {{action 'moveLeft'}}>
-            <i class="fa fa-caret-left"></i>
-        </button>
+        {{#if showArrows}}
```
caneruguz: Is there logic to remove only one direction of the buttons when there is no more to scroll?

app/components/supplementary-file-browser.js
erinspace: `primary_file` is indeed available via `v2/preprints/` on the OSF side!

app/templates/components/supplementary-file-browser.hbs
geeksnglitter: Not currently. I'll add it.

app/templates/preprints/view.hbs
```diff
@@ -34,7 +38,7 @@
                 </p>
                 {{!PROJECT BUTTON}}
                 <div class="p-lg osf-box-lt" id="view_project">
-                    <span> <a type="button" class="btn btn-primary pull-right" href={{concat model.baseUrl model.project}}>View project</a> The project for this paper is available on the Open Science Framework</span>
+                    <span> <a type="button" class="btn btn-primary pull-right" href={{concat config.OSF.url model.id}}>View project</a> The project for this paper is available on the Open Science Framework</span>
```
abought: Instead of manually creating the URL in the page, we should be able to rely on the URLs that are already part of the API response. Eg,

`node.links.html`

app/templates/components/supplementary-file-browser.hbs
abought: Total aside, but I never realized that class had a deeper convention / underlying meaning. Neat!

PR #23, opened 2016-08-11
Update payload to match new parser for file relationship

erinspace:
Fixes bug pointed out by [comment in #21](https://github.com/CenterForOpenScience/ember-preprints/pull/21/files#diff-ea08ccd8a33b794610a037162feb43d5R8) that relationship field name was odd. Relationship field name should now be passed to match name of primary_file in preprint serializer (was before relying on the type field).

**Note!** This PR relies on merging https://github.com/CenterForOpenScience/osf.io/pull/6141 which fixes the preprint relationship parser to look at the appropriate keys instead of relying on the type passed in with data


abought:
Thanks for checking into this!

PR #26, opened 2016-08-12
Feature/page animation

ml3ha:
Add basic transitions to routes using Liquid Fire. These transitions are all in `transitions.js`. 

The transitions are done simply by specifying them in `transitions.js` and then changing the `{{outlet}}` in `application.hbs` to `{{liquid-outlet}}`

See http://ember-animation.github.io/liquid-fire/

Also added some margins to the search bar since it was way too high up...

### DIFF
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -6,7 +6,7 @@
     </li>
 {{/osf-navbar}}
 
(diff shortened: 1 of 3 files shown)


caneruguz:
Will redo with passing tests.

PR #28, opened 2016-08-12
Feature/subjects

hmoco:


abought:
I don't know how to tell you this, but this PR already has merge conflicts.

app/helpers/if-filter.js
```diff
@@ -0,0 +1,10 @@
+import Ember from 'ember';
+
+export function ifFilter(params) {
+    if ((typeof params[1] === 'undefined') || ((params[0].toLowerCase()).indexOf(params[1].toLowerCase()) !== -1)){
```
abought: Can you please add a comment describing the purpose of this helper?

We use YUIDoc format- here is a sample:
https://github.com/abought/ember-osf/blob/9264fb5e8cdc52f30436bd9668f8a2905a0389a4/addon/helpers/elem-id.js#L9-L9

app/templates/components/preprint-form-subjects.hbs
```diff
@@ -17,42 +17,48 @@
...
     <div class="col-md-4">
-        <ul>
+        <ul style="overflow: scroll">
             {{input value=filter.[1].value class="form-control" placeholder="Search"}}
             {{#with path.[0] as |taxonomy|}}
                 {{#each filteredPath.[0].children as |element|}}
-                    <li {{action 'select' taxonomy element}} class={{if (eq path.[1].name element.name) 'selected'}}>
-                        {{element.name}}
-                    </li>
+                    {{#if (if-filter element.name filter.[1].value)}}
```
abought: Implementation question: 

What are filter[0], 1, 2, etc? Why are these stored as hard-coded array indices?

app/templates/components/preprint-form-subjects.hbs
```diff
@@ -17,42 +17,48 @@
...
+                        {{element.name}}
+                    </li>
+                {{/if}}
             {{/each}}
         </ul>
     </div>
 
     <div class="col-md-4">
-        <ul>
+        <ul style="overflow: scroll">
```
abought: When possible, and as time permits, we encourage styles to go in a `componentName.scss` file, rather than in templates.

app/controllers/add-preprint.js
```diff
@@ -95,55 +95,61 @@ export default Ember.Controller.extend(Validations, NodeActionsMixin, {
...
-                        this.get(`filter.${i + 1}.value`).indexOf(child.name || child) !== -1)
-                };
-            }
-            return path;
-        });
+    * Subjects section
+    */
+    topFilter: '',
+    midFilter: '',
+    botFilter: '',
```
abought: Will every taxonomy always have exactly three levels of filtering (and no more)?

app/controllers/add-preprint.js
```diff
@@ -95,55 +95,61 @@ export default Ember.Controller.extend(Validations, NodeActionsMixin, {
...
-                children: ['j', 'k']
-            }]
-        }, {
-            name: 'l'
-        }].filter(taxonomy =>
-            !this.get('filter.0.value') || taxonomy.name.indexOf(this.get('filter.0.value')) !== -1
-        );
+    sortedTaxonomies: Ember.computed('taxonomies', function() {
+        var _this = this;
+        this.get('store').query('taxonomy', { filter: { parent_ids: 'null' }, page: { size: 100 } }).then(results => {
```
abought: We're probably going to have to merge this as is- but for future reference, I'd encourage you to abstract away common operations like this query. In particular, fetching 100 results can be very slow, and it would be good to define the page size for all these operations in one place in case we decide to change/ edit it later.

app/templates/add-preprint.hbs
```diff
@@ -104,39 +104,45 @@
...
                                 {{#with path.[0] as |taxonomy|}}
                                     {{#with path.[1] as |category|}}
                                         {{#each filteredPath.[1].children as |element|}}
-                                            <li {{action 'selectSubject' taxonomy category element}}
-                                                class={{if (get (get (get selected taxonomy.name) category.name) element) 'selected'}}>
-                                                {{element}}
-                                            </li>
+                                            {{#if (if-filter element.name botFilter)}}
+                                                <li {{action 'selectSubject' taxonomy category element}}
+                                                    class={{if (get (get (get selected taxonomy.name) category.name) element.name) 'selected'}}>
```
abought: Why are there three get helpers? This seems a bit complex.

abought:
I have some concerns that this implementation locks us into a narrow definition of a taxonomy, but we need something to go forward with integration, so merging as is for now.

Also, in the future, please add a PR description to help with review. (tips for generating nested test data etc)

PR #30, opened 2016-08-15
Feature/Author Management Follow-Up

pattisdr:
# Purpose

Continued improvement of author management portion of preprints.

# Changes
- [x] Remove stillAdmin variable.  One variable, isAdmin, is used.
- [x]  Clicking cancel on 'Add Unregistered Contributor' view switches back to search results. 
- [x] Pressing 'enter' automatically adds user.
- [x] mut template helper for updating query 
- [x] If search results include user that is already an author, show checkmark instead of removing user from list.
- [x] div flash green on success, red on error
- [x] ~~change addUnregisteredContributor action to reflect new API changes~~  (HOLD pending https://github.com/CenterForOpenScience/osf.io/pull/6157)
- [ ] Accomodate new ember-osf changes - contributor.save() instead of node.save()


abought:
Thanks! Could you let me know when this is done, so we can make sure it gets integrated into the rest of the page?

As for the `stillAdmin` thing, I'm pretty sure it happens because `currentUserPermissions` is essentially a calculated field provided by the server. 

Since we don't change the node locally, the node record itself isn't dirty- so the new node fields aren't fetched when we send a request to the contributors endpoint. Basically, there's a case where the node record can fail to reflect the picture of reality as known by the server.

app/templates/components/preprint-form-authors.hbs
```diff
@@ -114,12 +121,12 @@
                                 {{#if contrib.unregisteredContributor}}
                                     {{contrib.unregisteredContributor}}
                                 {{else}}
-                                    <a href={{contrib.users.profileUrl}}> {{contrib.users.fullName}} </a>
+                                    <a href={{contrib.users.links.html}}> {{contrib.users.fullName}} </a>
```
abought: Can you clarify why the computed shortcut was replaced with a hardcoded path to the field here?

tests/unit/helpers/user-is-contributor-test.js
```diff
@@ -0,0 +1,29 @@
...
+
+
+test('Returns false if user is not a contributor', function(assert) {
+    var user = Ember.Object.create({
+        id: '12345'
+    });
+    var contributors = [Ember.Object.create({
+        userId: 'abcde'
+    })];
+  let result = userIsContributor([user, contributors]);
```
abought: Tests!!! 🌊 🐙 🌴

abought:
Thanks for the changes, and demo earlier! Merging as is so we can integrate and keep moving forward.

app/templates/components/preprint-form-authors.hbs
pattisdr: It just wasn't working. I do not know why. When the page loaded, contrib.users.profileUrl was undefined, so none of the profile links worked.

abought: Hmm, that's very strange- it's a helper on the user model, and should be available as long as the relationship can be loaded. (in which case both ways of getting the data would be there?)

Not urgent, but certainly something we'll keep an eye out for.

PR #31, opened 2016-08-15
Implement tree view of subject and allow for adding filters

hmoco:


caneruguz:
@hmoco can you check in on this PR?

app/components/taxonomy-tree.js
```diff
@@ -2,48 +2,52 @@ import Ember from 'ember';
...
-                */
-
-                // getChecked() is a valid method, but it's not in the README for bootstrap-treeview
-
-                this.$('#taxonomyTree').treeview('toggleNodeChecked', [data.nodeId, { silent: true }]);
+    cache: {},
+    init() {
+        this._super(...arguments);
+        var _this = this;
+        this.get('store').query('taxonomy', { filter: { parent_ids: 'null' }, page: { size: 100 } }).then(function(results) {
```
caneruguz: @hmoco Can you put a comment on the temporariness of the 100? Also make it a constant and place more globally. that should also be where the comment should go. 
We do have to do a sanity check that this indeed will bring all subitems.

app/components/taxonomy-tree.js
```diff
@@ -2,48 +2,52 @@ import Ember from 'ember';
...
+                Ember.set(item, 'showChildren', false);
+                return;
+            }
+            let children = item.children;
+            if (children && children.length > 0) {
+                Ember.set(item, 'showChildren', true);
+                return;
+            }
+            this.get('store').query('taxonomy', { filter: { parent_ids: item.id }, page: { size: 100 } }).then(function(results) {
+                    var topLevel = [];
```
caneruguz: Is toplevel the right naming here? in this instance isn't it just a level?

app/components/taxonomy-tree.js
```diff
@@ -2,48 +2,52 @@ import Ember from 'ember';
...
+            if (item.showChildren) {
+                Ember.set(item, 'showChildren', false);
+                return;
+            }
+            let children = item.children;
+            if (children && children.length > 0) {
+                Ember.set(item, 'showChildren', true);
+                return;
+            }
+            this.get('store').query('taxonomy', { filter: { parent_ids: item.id }, page: { size: 100 } }).then(function(results) {
```
caneruguz: This code is written twice with minor differences. Can you make it into a function for DRY? function(parent_ids, size, what to set to i.e. _this). If it's worth the time. Not a requirement.

app/controllers/discover.js
```diff
@@ -171,6 +171,15 @@ export default Ember.Controller.extend({
...
             this.set('page', 1);
             this.loadPage();
+        },
+
+        selectSubjectFilter(subject) {
+            let match = this.get('activeFilters').filter(function(item) {
+                return item.indexOf(subject.text) !== -1;
+            });
+            if (!match.length) {
+                this.get('activeFilters').pushObject('subject:' + subject.text);
```
caneruguz: Why is there a need to prefix it with "subject:"? Is this because of previous understanding of how share would send subjects?

hmoco: activeFilters holds all filters. It should be an object with ` { type of filter: values }`, but right now it is a list of strings with prefixes. One of my todos after demo is refactor this.

caneruguz:
Merging this now, changes will be comign in the next PR from the same branch.

PR #32, opened 2016-08-16
Provider support for discover page

hmoco:
Add the provider filter to the discover page. Also uses subject and provider filters to update search when those are added or removed.

### DIFF
--- a/app/components/taxonomy-tree.js
+++ b/app/components/taxonomy-tree.js
@@ -1,22 +1,29 @@
 import Ember from 'ember';
 
+//currently using the max page size our api allows, to not have to paginate. This
+// (should, test incoming) get all children of any given subject
+var pageSize = 100;
+
 export default Ember.Component.extend({
     store: Ember.inject.service(),
     cache: {},
+    _parseResults(results) {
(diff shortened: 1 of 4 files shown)


caneruguz:
@hmoco Can you provide descriptions for your PR's?

app/components/taxonomy-tree.js
```diff
@@ -1,22 +1,29 @@
...
 
+//currently using the max page size our api allows, to not have to paginate. This
+// (should, test incoming) get all children of any given subject
+var pageSize = 100;
+
 export default Ember.Component.extend({
     store: Ember.inject.service(),
     cache: {},
+    _parseResults(results) {
+        var parsed = [];
```
chrisseto:
```javascript
return results.map(result => ({
  id: result.id,
  text: result.get('text'),
  children: [],
  showChildren: false
}));
```

caneruguz:
@hmoco I'm merging this but the optimizations such as the one chris suggested should be happening soon. I'll make sure we get back to this and fix. It's not functionally important but better code.

PR #37, opened 2016-08-18
Refactor search filter code

hmoco:
Move all search filter from being string based with a prefix, to an object holding current filters.


abought:
Manual testing of the discover page reviews a few oddities. We're going to roll this into another PR for now, but can you please follow up?

Most of these comments refer to initial load of page. Sometimes buttons start working after I take other actions (like sorting).

1. The "refine your search" items look like checkboxes, but (mostly) act like radio buttons that only allow one selection at a time. This is slightly confusing.
2. If I click "clear filters", the labels are on the same line.
3. If I try to select a subject filter, the checkbox does nothing. (clickable but not selected) In the console, I see `discover.js:267 Uncaught TypeError: Cannot read property 'filter' of undefined`
4. Why does the subject tree show an expand icon next to things with no children? If I try to expand, I get `discover.js:27 Uncaught TypeError: Cannot read property 'slice' of undefined`

screen shot 2016-08-19 at 10 32 15 am

screen shot 2016-08-19 at 10 33 27 am

abought:
Also: clicking "clear filters" does not clear subject filters.

screen shot 2016-08-19 at 10 40 46 am

abought:
Once all filters are cleared, I can no longer see search results. (even though I had them when I first loaded the page). My console froze up, but at some point I got this error message which might be helpful:

`discover.js:137 Uncaught TypeError: Cannot read property 'indexOf' of undefined`

screen shot 2016-08-19 at 10 41 58 am


Sorry for the notes, but hope these examples and error message links help you trace things a bit.

abought:
Lastly, any idea why we get all these "rerender" warnings? I;m noticing some sluggish page behavior, and ember might be trying to warn us about... something?

screen shot 2016-08-19 at 10 46 00 am

abought:
Rolled into #45, but requesting author create tickets to follow up on outstanding issues with the page.

PR #40, opened 2016-08-18
Support cookie auth, token auth, and enforcement of user being logged in

abought:
# Ticket
https://openscience.atlassian.net/browse/PREP-118

❗ This PR should be merged with:
https://github.com/CenterForOpenScience/ember-osf/pull/116
⚠️  It will not build/run until that PR has been merged

# Purpose
- Support both token and cookie auth from the same codebase.
- Define what happens if a user tries to visit a route that requires them to be logged in
- Make login and navbar functionality more consistent with the OSF

# Summary of changes
- When a user is not logged in, they are now sent to an external site, then returned to the page that needed auth (for cookie auth-- token auth returns them to whatever was registered as the redirect URI)
- The same codebase now works with token and cookie auth. (see notes in EOSF companion PR to test this)

# Notes for reviewers
## Tips for testing
See section on [companion PR](https://github.com/CenterForOpenScience/ember-osf/pull/116) for details.

## Anomalies
When viewing on `localhost:5000`, both auth methods seem to work correctly, whether the user is logged in or not.

When viewing on other ports (eg `localhost:4200`), unauthenticated GET requests fail with CORS errors. This appears to be [by design](https://lists.w3.org/Archives/Public/public-webappsec/2012Jul/0032.html). Basically, this means we’d need the consuming site to be in the [CORS Origin whitelist](https://github.com/abought/osf.io/blob/f0bb77351322667d240c05b7230e944eab12094f/api/base/settings/defaults.py#L112) in order to demo with it on a non-OSF server.

If the user provides a valid token, it seems we explicitly allow cross-origin requests. But not if they are visiting the site without being logged in. The reason we do not see errors for unauthenticated views on port 5000 is because the flask app is explicitly whitelisted.

# Checklist of expected behaviors
## Basic behaviors
- [x] Can toggle between cookie and token auth by changing [a setting](https://github.com/abought/ember-preprints/blob/f57fbd845a8960a824dea9b4594468356c60343c/config/environment.js#L4-L4) (may need to `localStorage.clear()` for the switch to work!)
- The right kind of auth gets sent, respecting the config variables
  - [x] Cookies send cookies
  - [x] Tokens add an `Authorization: Bearer x` heading

## Recognize authentication when appropriate
- Don't try to send credentials when logged out
  - [x] For cookies
  - [x] For tokens
- [x] If user has a cookie, the navbar should show that they are logged in on first page view- seamless transition from flask app to ember app.
- When not logged in, trying to access pages that require authentication redirects to CAS...
    - [x] The page for cookies
    - [x] The page for tokens
- When logged in, can access pages that require authentication (no redirect to CAS)
  - [x] Cookies
  - [x] Tokens

## Logging in and out
- Clicking "Log in" on the navbar takes the user to the appropriate login page
  - [x] For cookies
  - [x] For tokens
- Clicking "Log out" acts as expected
  - [x] For cookies, redirect to `/goodbye`
  - [x] For tokens, stay on page but clear session. Navbar clears current user, and user is still logged out when they refresh page

## Reading and writing data
- When not logged in
 - Can read public data sufficient to render homepage
    - [x] Cookies
    - [x] Tokens
- When logged in: 
 - Can read public and private data
    - [x] Cookies
    - [x] Tokens
  - Can write data to basic API endpoints
    - [x] Cookies
    - [x] Tokens
  - Can upload files via dropzone widget
    - [x] Cookies
    - [x] Tokens
  - Can upload files using fileManager service
    - [x] Cookies
    - [x] Tokens


abought:
Reminder to self: the CI failures are expected; this is asking for things from ember-osf that haven't been merged yet.

PR #41, opened 2016-08-18
View page: Add all authors/ Add tooltips

geeksnglitter:
A few fixes for the view page. 

Fixing tooltips to use hint.css.
Adding all authors - actually grab incrementally later.

Comment on ticket mentions other things that still need to be fixed.


abought:
Folded into #45.

PR #43, opened 2016-08-19
[rfr] Fix errors with contributor updates

abought:
# Ticket
https://openscience.atlassian.net/browse/PREP-121

# Purpose
Fix errors with add/edit page as reported by @pattisdr 

# Summary of changes
- Rename `selectedNode` to `node` due to requirement of nodeActionsMixin.
- Use a slightly nicer default DOI value


pattisdr:
Confirmed that this fixes the errors I was seeing.

abought:
Folded into #45 per @chrisseto request.

PR #44, opened 2016-08-19
Move upload logic into a component

chrisseto:


chrisseto:
requires https://github.com/CenterForOpenScience/ember-osf/pull/115

PR #49, opened 2016-08-22
Do some of the things

chrisseto:


abought:
Reviewer note for future users:

Use of the subject widget will require regenerating new fake "subjects" in your DB test data. (or remapping existing ones)

Otherwise, due to recent field names in the DB model (parent_ids --> parents), the query will return 0 results and the subject selector will appear blank.

abought:
It's logey in places, but mostly works. See note above for testing.

PR #52, opened 2016-08-22
[WIP] Feature/Creating New Component For Preprint

pattisdr:
# Purpose

- Turns "Select a file from OSF" portion into a 3-step form.  
- Creating a component copies file and adds contributors from existing project


pattisdr:
Sending follow-up PR

PR #64, opened 2016-08-23
Feature/Reset button on upload portion

pattisdr:
# Purpose

Adds functionality to reset button.   After file and node have been locked, and user later returns to upload portion of the form, they have the option to reset the preprint.

The reset button deletes projects created and files uploaded as part of preprint process and then refreshes page.

### DIFF
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -348,7 +348,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
                 }));
             });
 
-            Ember.RSVP.all(filePromises).then(() => {
+            Ember.RSVP.allSettled(filePromises).then(() => {
                 this.get('projectsCreatedForPreprint').forEach((project) => {
                     promisesArray.push(project.destroyRecord().then(() => {
                         this.get('toast').info('Project removed!');
@@ -358,7 +358,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
                 });
             });
 
-            Ember.RSVP.all(promisesArray).then(() => {
+            Ember.RSVP.allSettled(promisesArray).then(() => {
                 if (promisesArray.length > 0 || filePromises.length > 0) {
                     window.setTimeout(
                     function() {
@@ -367,9 +367,8 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
                 } else {
                     window.location.reload();
                 }
-
             });
-
         }
     }
 });
+
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -350,7 +350,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
 
             Ember.RSVP.all(filePromises).then(() => {
                 this.get('projectsCreatedForPreprint').forEach((project) => {
-                    promisesArray.push(this.store.adapterFor('node').ajax(project.get('links.self'), 'DELETE').then(() => {
+                    promisesArray.push(project.destroyRecord().then(() => {
                         this.get('toast').info('Project removed!');
                     }).catch(() => {
                         this.get('toast').error('Could not remove created project/component');
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -131,6 +131,8 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
     chooseExistingProjectHeader: '1. Choose existing OSF Project',
     createComponentHeader: '3. Convert this project or copy file to new component',
     fileAndNodeLocked: false,
+    projectsCreatedForPreprint:  Ember.A(),
+    filesUploadedForPreprint: Ember.A(),
 
     isAdmin: Ember.computed('node', function() {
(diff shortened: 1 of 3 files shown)


app/controllers/submit.js
```diff
@@ -333,8 +338,38 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
...
+            var filePromises = [];
+            this.get('filesUploadedForPreprint').forEach((file) => {
+                filePromises.push(this.get('fileManager').deleteFile(file).then(() => {
+                    this.get('toast').info('File removed!');
+                }).catch(() => {
+                    this.get('toast').error('Could not remove uploaded file');
+                }));
+            });
+
+            Ember.RSVP.all(filePromises).then(() => {
```
abought: This seems to wrap the success or failure states of many different operations into one. Since all fails fast, it's possible that we could end up in a state where some files had been deleted, but not others, and we had not correctly given this page an accurate picture of reality at the end of these operations.

Would it be more appropriate to use `RSVP.allSettled` here, and check the failure state? (note the return values look a bit different for this method, see [the docs](http://emberjs.com/api/classes/RSVP.html#method_allSettled))

I could reasonably consider this to be something for a followup PR, but it is an error case to be aware of.

app/controllers/submit.js
```diff
@@ -333,8 +338,38 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
...
+                this.get('projectsCreatedForPreprint').forEach((project) => {
+                    promisesArray.push(this.store.adapterFor('node').ajax(project.get('links.self'), 'DELETE').then(() => {
```
abought: Can you please clarify the intent of this line vs using ember's native methods like `node.deleteRecord().save()` or `node.destroyRecord()`?

Are we intentionally circumventing the store here to avoid cleaning up related record items? Should we be?

app/controllers/submit.js
```diff
@@ -333,8 +338,38 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
...
+                    }));
+                });
+            });
+
+            Ember.RSVP.all(promisesArray).then(() => {
+                if (promisesArray.length > 0 || filePromises.length > 0) {
+                    window.setTimeout(
+                    function() {
+                        location.reload();
+                    }, 3000);
```
abought: What is the purpose of this 3-second wait?

abought:
Light functional testing works, but before merging there are a few questions to clarify- most particularly surrounding the 3-second reload delay. (just waiting x seconds doesn't guarantee something has finished happening; want to make sure this isn't a potential bug source outright)

app/controllers/submit.js
pattisdr: The purpose of the 3 second wait was just so the user could see the success messages.  This branch isn't hit until the promises have resolved.

app/controllers/submit.js
pattisdr: I don't think node.deleteRecord() works?  The delete record on the nodeActions mixin finds the node that has already been loaded on the route.  So I temporarily did this until we can add functionality to ember0sf.

app/controllers/submit.js
pattisdr: I hadn't looked into RSVP.allSettled - thanks - I'll look more into this in the AM.

app/controllers/submit.js
abought: We seem to be discussing two separate but related  things here...
- The method on the model object itself, [built into ember](http://emberjs.com/api/data/classes/DS.Model.html#method_deleteRecord)
- An [action](https://github.com/abought/ember-osf/blob/0f93704050569ad2cda033a81e4f3a9075a391c4/addon/mixins/node-actions.js#L68-L68) in the NodeActionsMixin, which uses the insta-save version of the method under the hood

I wasn't aware of anything in ember-osf that was straight-out breaking deletions. That's interesting to know- if it's not tracked yet, might be a good idea to make it so when you have a chance.

abought: Oops, didn't refresh page to see this comment. Sounds good!

app/controllers/submit.js
abought: Sounds good; thanks.

abought:
Comments cleared, pending developer discretion on the deletion change. `allsettled` could be done separately in a later followup in the interests of demo timelines.

app/controllers/submit.js
```diff
@@ -358,7 +358,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
                 });
             });
 
-            Ember.RSVP.all(promisesArray).then(() => {
+            Ember.RSVP.allSettled(promisesArray).then(() => {
```
abought: Small heads up- this may affect any error handling you do. Since failure results are summarized in objects, IIRC- if a request fails it still kicks you into the `.then` rather than `catch`.

Relevant if you want to add a future mechanism to warn the user that deletion/ cleanup failed; all vs allsettled don't appear to be drop-in replacements. http://emberjs.com/api/classes/RSVP.html#method_allSettled

Though I'm pretty fried, so rather than pretending to be 20, it's off to bed I go. Thanks for the hard work!

PR #65, opened 2016-08-24
add preprints and subjects to search and display

hmoco:


app/controllers/discover.js
```diff
@@ -97,13 +99,15 @@ export default Ember.Controller.extend({
...
+                source.subjects = source.subjects.map(function(each) {return {text: each};});
                 source.providers = source.sources.map(item => ({name: item}));
                 source.contributors = source.contributors.map(function(contributor) {
                     return {
                         users: {
-                            familyName: contributor.family_name,
-                            givenName: contributor.given_name,
-                            id: contributor['@id']
+                            familyName: contributor
+                            // familyName: contributor.family_name,
```
abought: Why are these lines commented out? Did the layout of fields returning from SHARE change?

hmoco: yep, that seems to be the case, contributors is now just an array of strings.

hmoco: Ideally we can get share to give us the whole thing again soon.

PR #66, opened 2016-08-24
Fix issue with subject selector

abought:
# Purpose
Fix a bug with the subject selector on submit page

# Steps to reproduce bug that should be fixed
1. Select a subject at the first and second levels of the taxonomy.
2. Select a third-level subject
3. Select a different second level subject
4. Select the original second-level subject. Note that the third-level section does not update with correct children.


abought:
Seems to work in daylight; if there are other issues we'll revisit.

PR #71, opened 2016-08-24
Feature/index page counts

erinspace:
### Add two counts to index home page

- Number of hosted preprints
- Number of searchable preprints from SHARE

screen shot 2016-08-24 at 10 52 34 am

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -16,8 +16,10 @@
                 <div class="preprint-search col-xs-10 col-xs-offset-1 col-md-8 col-md-offset-2 m-v-lg" >
                     {{search-preprints search="search"}}
                     <p class="p-l-md">
-                        <span class="f-w-lg">{{model.preprints.length}}</span> preprints as of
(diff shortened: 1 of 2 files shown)


app/templates/index.hbs
```diff
@@ -16,8 +16,10 @@
                 <div class="preprint-search col-xs-10 col-xs-offset-1 col-md-8 col-md-offset-2 m-v-lg" >
                     {{search-preprints search="search"}}
                     <p class="p-l-md">
-                        <span class="f-w-lg">{{model.preprints.length}}</span> preprints as of
-                        <span>{{moment-format model.theDate "MMMM DD, YYYY"}}</span>
+                        <span class="f-w-lg">{{model.preprints.length}} hosted preprints, </span>
```
abought: If you have more than 10 results, how many does `model.preprints.length` display here? Ember data only fetches the first page of results for relationships used this way, so you may get an incorrect number. `links.meta.total` can be a good, API-provided alternative way of getting the more accurate count.

erinspace: Awesome! Thank you. Will make a follow up PR!

PR #73, opened 2016-08-24
Remove our count, format SHARE count with commas

erinspace:
## Purpose

As per discussion with product and co, remove "hosted" count and format SHARE preprint count with commas to make it readable!

screen shot 2016-08-24 at 3 17 33 pm

### DIFF
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -7,6 +7,9 @@ export default Ember.Route.extend({
     model() {
         var getTotalPayload = '{"size": 0, "from": 0,"query": {"bool": {"must": {"query_string": {"query": "*"}}, "filter": [{"term": {"type.raw": "preprint"}}]}}}';
 
+        function formatNumberWithCommas(x) {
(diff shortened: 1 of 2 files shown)


app/routes/index.js
```diff
@@ -7,14 +7,17 @@ export default Ember.Route.extend({
     model() {
         var getTotalPayload = '{"size": 0, "from": 0,"query": {"bool": {"must": {"query_string": {"query": "*"}}, "filter": [{"term": {"type.raw": "preprint"}}]}}}';
 
+        function formatNumberWithCommas(x) {
```
abought: Maybe not relevant here, but this looks super useful.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

The browser compatibility chart seems to suggest it works in safari, if you let it choose the locale for you. (eg it doesn't understand all the arguments, because Safari is weird)

erinspace: Yes yes yes wonderful!!! I was looking for something like this! Thanks andy!!!

PR #100, opened 2016-08-27
[quick] Search result layout and author name order

abought:
# Purpose
1. Improve search result display on `/discover` page
2. Change authors to display with `first Last` consistently across pages- don't use different fields or different orders
3. Right align date on search results for now. I'll defer visual tweaks to @caneruguz 's talents.

screen shot 2016-08-27 at 7 37 19 pm

# Changes
- Change how author names are displayed. Where possible, use `firstName lastName` format in place of `fullName` or `last, first`.
  - The `fullName` field in the OSF is intended for profile display; individual name fields are intended to provide more fine-grained control over how name is shown.

# Tracking
https://trello.com/c/6fD8lRQ6/60-improve-layout-of-search-results

# Spec
 Formatting of the preprint summaries could be refined. 

 - ~~Top-left alignment for title~~ (seems to already be like this?)
 - [x] Move date down to bottom-right alignment
 - [ ] Source (OSF or SSRN or ARXIV) aligned right Right above or below date (will check with @caneruguz on preferred styling)
 - ? No commas after names if the “|” is also there (not sure what this is asking for?)
 - ? No space before comma in list of tags
 - [ ] Tighter in vertical space on top and bottom with the expand arrow being closer to the bottom row of text

### DIFF
--- a/app/templates/components/search-result.hbs
+++ b/app/templates/components/search-result.hbs
@@ -45,9 +45,11 @@
                             {{if index '| '}}{{provider.name}}
                         {{/each}}
                     </span>
-                    {{#if result.date_modified}} {{! moment-format will use current time if null}}
-                        <span>{{moment-format result.date_modified "MMMM YYYY"}}</span>
-                    {{/if}}
+                    <span class="pull-right">
+                        {{#if result.date_modified}} {{! moment-format will use current time if null}}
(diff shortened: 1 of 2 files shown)


abought:
Merging these quick tweaks. There are a few visual items I've triaged, but I'd like @caneruguz 's opinion on the proposed layout changes here. (spacing and alignment)

PR #101, opened 2016-08-28
View page file download button should match the file picker

abought:
# Purpose
Let download button on preprint view page provide the correct download link for the file currently selected in the file picker.

screen shot 2016-08-27 at 8 50 18 pm

https://trello.com/c/SkGOyIWO/93-download-button-should-download-the-currently-selected-file


erinspace:
Tested locally and all looks great! :+1:

PR #108, opened 2016-08-29
Only display carrots when there are children

erinspace:
## depends on https://github.com/CenterForOpenScience/osf.io/pull/6238

carats now only appear when there is a 3rd row of taxonomy

### before
screen shot 2016-08-28 at 8 57 52 pm

### after
screen shot 2016-08-28 at 8 58 59 pm

### ticket 
https://trello.com/c/GETV8yQ1/111-preprints-discover-the-caret-does-no-action

### DIFF
--- a/app/components/taxonomy-tree.js
+++ b/app/components/taxonomy-tree.js
@@ -14,7 +14,8 @@ export default Ember.Component.extend({
                 id: result.id,
                 text: result.get('text'),
                 children: [],
-                showChildren: false
+                showChildren: false,
+                childCount: result.get('child_count')
             });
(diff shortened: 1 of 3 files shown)


mfraezz:
It might look nicer if the checkboxes were aligned. Is it possible to add some kind of offset to taxonomy items that don't have children?

erinspace:
okay

erinspace:
@mfraezz the expanded carat width is one pixel wider than the unexpanded carat width :(

caraterss

erinspace:
victory.

jumpin

PR #114, opened 2016-08-29
Add tooltips to subjects

hmoco:
https://trello.com/c/0WrR1Nc2/19-show-tooltip-examples-of-5-subdisciplines-on-hover-of-category

### DIFF
--- a/app/components/taxonomy-top-list.js
+++ b/app/components/taxonomy-top-list.js
@@ -2,62 +2,22 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
     subjectTooltips: {
-        'Physical sciences': '<p>Astronomical Sciences</p>' +
-            '<p>Chemistry</p>' +
-            '<p>Materials Science</p>' +
-            '<p>Mathematics</p>' +
-            '<p>Physics</p>',
-        'Engineering and technology': '<p>Civil Engineering</p>' +
-            '<p>Digital Imaging</p>' +
-            '<p>Fluidics</p>' +
-            '<p>Nanotechnology</p>' +
-            '<p>Synthetic Biology</p>',
-        'Biology and life sciences': '<p>Ecology</p>' +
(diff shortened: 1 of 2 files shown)


app/components/taxonomy-top-list.js
```diff
@@ -1,6 +1,64 @@
...
+            '<p>Marine and aquatic sciences</p>' +
+            '<p>Mineralogy</p>',
+        'Ecology and environmental sciences': '<p>Biogeochemistry</p>' +
+            '<p>Environmental Geology</p>' +
+            '<p>Natural Resources</p>' +
+            '<p>Soil Science</p>' +
+            '<p>Sustainability Science</p>'
+    },
+    didRender() {
+        Ember.$('.subject-tooltips').tooltip();
```
abought: Are we not using hint-css for tooltips now?

caneruguz: Ah yes, my bad I should have told @hmoco about this. Can we change it to use hint css? An example is on the view page https://github.com/CenterForOpenScience/ember-preprints/blob/develop/app/templates/components/supplementary-file-browser.hbs#L15

abought: One other small question: is this intended to be a single tooltip? That's quite lengthy. (and may not get updated when the taxonomy changes)

caneruguz: @abought thanks for the question. As I see it, each subject has five examples and for each only those five will be in the tooltip. We did discuss how the ones should be selected and decided to manually add curated ones and not do a random or top of the list approach. Yes we will need to keep an eye on taxonomy changes, and probably already now since we added new top level items but we will just have to keep up with that maintenance.

caneruguz:
@hmoco I added a note to the Trello card about new taxonomy items we added to the top. We may have to make a decision about that soon. I don't know if that's a hindrance yet.

PR #122, opened 2016-08-29
Fix Relationship Validation [OSF-6969]

caseyrollins:
#### Purpose
- Fixes failing relationship validation for preprint payload.
- Changes the `primary_file` relationship type to match the `primary_file` relationship key. 

#### Ticket
- Relates to [OSF-6969](https://openscience.atlassian.net/browse/OSF-6969)

### DIFF
--- a/app/serializers/preprint.js
+++ b/app/serializers/preprint.js
@@ -8,7 +8,7 @@ export default OsfSerializer.extend({
             primary_file: {
                 data: {
                     id: snapshot.belongsTo('primaryFile', { id: true }),
-                    type: 'file'
+                    type: 'primary_file'
                 }
             }
         };


brianjgeiger:
I don't know. This seems a liiiittle too complicated of a change. Could you try simplifying a bit?

(This is good if the tests pass. 🎉 🐧 🎉 )

abought:
Hmm. I notice that this actually reverts a change made from `file` to `primary_file` under the hood- any specific reason? See 38cd2a7cbe0c9f58153a5feb16ced610216c3f76

(ultimately, I thought these were still `type:file` for the API under the hood)

brianjgeiger:
Okay, looked over this more. Going to revert out the old change and not merge this.

brianjgeiger:
Feel free to close this, someone who has push access to this repo.

caseyrollins:
@abought, basic relationship validation was added recently (per [#6189](https://github.com/CenterForOpenScience/osf.io/pull/6189), to resolve [OSF-])

mfraezz:
>Hmm. I notice that this actually reverts a change made from primary_file to file previously- any specific reason? See https://github.com/CenterForOpenScience/ember-preprints/commit/38cd2a7cbe0c9f58153a5feb16ced610216c3f76

Reverting that probably isn't for the best, as it would set the `relationship_key` to be `preprint_file`, not `primary_file`. This PR seems like it's the right change

abought:
In this case I'm referring to the change to the type field, not the relationship key. (Sorry, wi two things changed,  that could have been more clear)

My understanding was that the type field had to match a type in the API, eg file, rather than a field name.

mfraezz:
Ah, okay. That makes more sense.

And so far as I can tell, the `type` field wasn't being validated as a legitimate type with a view in the API, only to be the same as the relationship key -- local testing with current API `develop` and this change indicates that it this works (that is, I get a `201` when submitting a preprint with this, or a `400` without, and it does become a valid preprint with a file). ¯\\\_(ツ)_/¯

brianjgeiger:
The reason this issue is closed and the other one reverted is because that was not how it's supposed to work. We will go down a better path for this in the near-ish future.

caseyrollins:
Also, the reverted change is [this](https://github.com/CenterForOpenScience/osf.io/pull/6189) PR, not the `primary_file`/`file` change @abought mentioned above.

PR #126, opened 2016-08-30
Don't preserve state across page loads [Trello]

abought:
# Ticket

https://trello.com/c/cXY0EInK/141-search-is-not-cleared-when-navigated-to-via-an-internal-link-e-g-search-nav
# Purpose

Fix issue where search page was remembering filter and subject selections from a previous visit. (when navigating via links from within ember)
# Summary of changes

Since controllers are singletons, a state-reset method needs to be defined and called on each page load.
# Side effects

Since query params are based on filters, and filters are reset on page load, this causes the back button to not remember state. (eg it resets to a generic search, even though query params ostensibly track what to load)

Long term this page will need some internal refactoring for how data is fetched, but that's a post-MVP change. This PR simply ensures that search and submit pages are reset on every visit.
### Search page
1. Go to search page. Select two providers and two subjects. Change search result sort order. 
2. On the project navbar, click "Add a preprint". 
3. From the /submit page, on the navbar again, click "Search" to return to the search page. The search filters should be reset to their defaults.
### Submit page

General workflows: make sure that once a preprint is started, the page completely reloads on next visit to the submit page. (and does not remember a previous preprint, node, etc)
# TODO
- [x] Reset method for search page (and ensure it's called before view)
- [x] Reset method for content page (and ensure it's called before view)
- [x] Vet uses of not-so-clever bindings, like `computed.oneWay` for content page controller

### DIFF
--- a/app/controllers/discover.js
+++ b/app/controllers/discover.js
@@ -30,8 +30,8 @@ export default Ember.Controller.extend({
     osfProviders: ['OSF', 'PsyArXiv', 'SocArXiv', 'engrXiv'],
 
     // Parameters tracked on the controller; default values set in `manualResetController` method
+    facetFilters: null,
     activeFilters: null,
-
     sortByOptions: null,
     numberOfResults: 0,
     showActiveFilters: null, //should always have a provider, don't want to mix osfProviders and non-osf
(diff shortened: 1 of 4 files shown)


caneruguz:
Note to reviewers, there may be no longer need for this fix, need to check

caneruguz:
This fix is still needed but it conflicts with important new changes. Will take this up on next release.

abought:
Does this track any known issue? It's so thoroughly out of date that I'm tempted to just close it.

(And for future reference, `resetController` may be promising)

abought:
This PR is almost a year stale. I'm closing it to clean up history, but commenting in case the issue still exists for a more modern fix.

PR #127, opened 2016-08-30
[WIP] Show doi and share links for search results

caneruguz:
Instead of showing all links provided by share we only show DOI and share links


caneruguz:
This solution is not preferred, will create new PR with another solution

PR #129, opened 2016-08-30
MathJax rendering

hmoco:
Render mathjax present on search results, view preprint page and submit previews.

https://openscience.atlassian.net/browse/PREP-145

### Search page `/discover`:
Search for "smooth cubic". Make sure the math characters are rendered. A math character might look like `${\mathbb R}$` in text, but like ℝ when rendered.
### On submission `/submit`:
For both abstract and title, use `Smooth Cubic Curves over ${\mathbb C}$ and ${\mathbb R}$ `. Attempt to reopen those panels and edit abstract and title and make sure that they are rendered properly on preview.
### On content page `/:guid (of the project created with testing submit)`:
Make sure both abstract and title get rendered properly.

### DIFF
--- a/app/components/search-result.js
+++ b/app/components/search-result.js
@@ -9,7 +9,7 @@ export default Ember.Component.extend({
         arXiv: 'arxiv'
     },
     didRender() {
-        MathJax.Hub.Queue(['Typeset', MathJax.Hub]);
+        MathJax.Hub.Queue(['Typeset', MathJax.Hub]);  // jshint ignore:line
     },
     numMaxChars: 300,
     showBody: false,
--- a/app/components/search-result.js
+++ b/app/components/search-result.js
@@ -3,10 +3,10 @@ import Ember from 'ember';
 export default Ember.Component.extend({
     providerUrlRegex: {
         //'bioRxiv': '', doesnt currently have urls
-        'Cognitive Sciences ePrint Archive': 'cogprints',
-        OSF: 'osf',
-        PeerJ: 'peerj',
-        arXiv: 'arxiv'
+        Cogprints: /cogprints/,
+        OSF: /https?:\/\/((?!api).)*osf.io/, // Doesn't match api.osf urls
+        PeerJ: /peerj/,
+        arXiv: /arxivj/
     },
     didRender() {
         MathJax.Hub.Queue(['Typeset', MathJax.Hub, this.$()[0]]);  // jshint ignore:line
@@ -26,23 +26,28 @@ export default Ember.Component.extend({
         return result.description.slice();
     }),
 
-    osfID: function() {
+    osfID: Ember.computed('result', function() {
         let re = /osf.io\/(\w+)\/$/;
-        if (this.get('result.osfProvider'))
+        if (this.get('result.providers').find(provider => provider.name === 'OSF'))
             for (let i = 0; i < this.get('result.lists.links.length'); i++)
                 if (re.test(this.get('result.lists.links')[i].url))
                     return re.exec(this.get('result.lists.links')[i].url)[1];
         return false;
-    }.property('result'),
+    }),
 
-    hyperlink: function() {
-        var rule = this.get('providerUrlRegex')[this.get('result.providers.0.name')];
-        var matches = [];
-        if (rule) {
-            matches = this.get('result.lists.links').slice().filter(each => each.url.indexOf(rule) !== -1);
-        }
-        return matches.length ? matches[0].url : this.get('result.lists.links.0.url');
-    }.property('result'),
+    hyperlink: Ember.computed('result', function() {
+        var re = null;
+        for (let i = 0; i < this.get('result.providers.length'); i++)
+            re = this.providerUrlRegex[this.get('result.providers')[i].name] || null;
+
+        if (!re) return this.get('result.lists.links.0.url');
+
+        for (let j = 0; j < this.get('result.lists.links.length'); j++)
+            if (re.test(this.get('result.lists.links')[j].url))
+                return this.get('result.lists.links')[j].url;
+
+        return this.get('result.lists.links.0.url');
+    }),
 
     actions: {
         toggleShowBody() {
--- a/app/components/search-result.js
+++ b/app/components/search-result.js
@@ -8,7 +8,9 @@ export default Ember.Component.extend({
         PeerJ: 'peerj',
         arXiv: 'arxiv'
     },
-
+    didRender() {
+        MathJax.Hub.Queue(['Typeset', MathJax.Hub]);
+    },
     numMaxChars: 300,
     showBody: false,
     footerIcon: Ember.computed('showBody', function() {
--- a/app/components/search-result.js
+++ b/app/components/search-result.js
@@ -9,7 +9,7 @@ export default Ember.Component.extend({
         arXiv: 'arxiv'
     },
     didRender() {
-        MathJax.Hub.Queue(['Typeset', MathJax.Hub]);  // jshint ignore:line
+        MathJax.Hub.Queue(['Typeset', MathJax.Hub, this.$()[0]]);  // jshint ignore:line
     },
     numMaxChars: 300,
     showBody: false,
@@ -23,7 +23,7 @@ export default Ember.Component.extend({
         if (result.description && result.description.length > this.numMaxChars) {
             return result.description.substring(0, this.numMaxChars) + '...';
         }
-        return result.description;
+        return result.description.slice();
     }),
 
     osfID: function() {
--- a/app/controllers/content.js
+++ b/app/controllers/content.js
@@ -22,6 +22,12 @@ export default Ember.Controller.extend({
              this.set('authors', contributors));
     }),
 
+    applyMathJax: Ember.observer('model', function() {
+        Ember.run.next(this, function() {
+            MathJax.Hub.Queue(['Typeset', MathJax.Hub]);
+        });
+    }),
+
     actions: {
         expandMFR() {
             this.toggleProperty('fullScreenMFR');
--- a/app/controllers/content.js
+++ b/app/controllers/content.js
@@ -22,12 +22,6 @@ export default Ember.Controller.extend({
              this.set('authors', contributors));
     }),
 
-    applyMathJax: Ember.observer('model', function() {
-        Ember.run.scheduleOnce('afterRender', this, function() {
-            MathJax.Hub.Queue(['Typeset', MathJax.Hub]);  // jshint ignore:line
-        });
-    }),
-
     actions: {
         expandMFR() {
             this.toggleProperty('fullScreenMFR');
--- a/app/controllers/content.js
+++ b/app/controllers/content.js
@@ -23,7 +23,7 @@ export default Ember.Controller.extend({
     }),
 
     applyMathJax: Ember.observer('model', function() {
-        Ember.run.next(this, function() {
+        Ember.run.scheduleOnce('afterRender', this, function() {
             MathJax.Hub.Queue(['Typeset', MathJax.Hub]);  // jshint ignore:line
         });
     }),
--- a/app/controllers/content.js
+++ b/app/controllers/content.js
@@ -24,7 +24,7 @@ export default Ember.Controller.extend({
 
     applyMathJax: Ember.observer('model', function() {
         Ember.run.next(this, function() {
-            MathJax.Hub.Queue(['Typeset', MathJax.Hub]);
+            MathJax.Hub.Queue(['Typeset', MathJax.Hub]);  // jshint ignore:line
         });
     }),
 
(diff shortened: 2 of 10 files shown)


app/controllers/content.js
```diff
@@ -22,6 +22,12 @@ export default Ember.Controller.extend({
              this.set('authors', contributors));
     }),
 
+    applyMathJax: Ember.observer('model', function() {
+        Ember.run.next(this, function() {
```
abought: Can you please comment on the uses of `Ember.run` in this PR?

While sometimes necessary inside unit tests, this is usually not necessary within page code.

app/controllers/content.js
```diff
@@ -22,6 +22,12 @@ export default Ember.Controller.extend({
              this.set('authors', contributors));
     }),
 
+    applyMathJax: Ember.observer('model', function() {
```
abought: Can you please clarify why this observer tracks the model (aka blank preprint)? 

The `model` hook is a blank preprint (to hold subject and DOI, mainly). But since most fields on the page with mathematical data (like subject and title) live on the `node`, it seems strange for this observer to watch the preprint.

If you just want this to run once when the page starts, you might consider moving the logic to `setupController `instead. Otherwise, it's not clear to me this code is firing when you expect, for the reasons expected.

app/controllers/submit.js
```diff
@@ -247,8 +256,8 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
             this.set('node.title', this.get('nodeTitle'));
             let node = this.get('node');
             node.save();
+            this.notifyPropertyChange('node.title');
```
abought: Given that we call `this.set()` on title above, can you please comment on the reason for needing `notifyPropertyChange` here? I would expect observers to pick up on the property change without this line, since we appear to be using normal setting mechanisms.

app/index.html
```diff
@@ -15,6 +15,15 @@
     <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,600,300,700' rel='stylesheet' type='text/css'>
 
     {{content-for "head-footer"}}
+    <script type="text/javascript"
+        src="//cdnjs.cloudflare.com/ajax/libs/mathjax/2.6.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML">
+    </script>
+    <script type="text/x-mathjax-config">
```
abought: I notice that this configuration is done in `index.html`, and that actual pages just call `Typeset`.

In the AM, can we touch base on this? I've not used MathJax before and want to be sure I understand what this is doing. 

In particular, I want to be sure this doesn't impose a performance penalty if MathJAX might try to do something on every page view / render, even if most pages don't need it.

app/index.html
```diff
@@ -15,6 +15,15 @@
     <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,600,300,700' rel='stylesheet' type='text/css'>
 
     {{content-for "head-footer"}}
+    <script type="text/javascript"
+        src="//cdnjs.cloudflare.com/ajax/libs/mathjax/2.6.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML">
```
abought: It seems that this URL configures it to load TeX, LaTeX, and MathML- very versatile, but also larger file size and possibly slower to render.  

Requirements question: What kind of content, exactly, are we rendering? Do we need to support multiple different notation systems for mathematical content, or can we perhaps ask MathJAX CDN for a more limited subset of its functionality?

http://docs.mathjax.org/en/latest/configuration.html#configuring-mathjax-after-it-is-loaded

app/components/search-result.js
```diff
@@ -8,7 +8,9 @@ export default Ember.Component.extend({
         PeerJ: 'peerj',
         arXiv: 'arxiv'
     },
-
+    didRender() {
+        MathJax.Hub.Queue(['Typeset', MathJax.Hub]);  // jshint ignore:line
```
abought: My understanding is that usually, there is another argument here. By omitting it, we may be causing mathjax to do far more work than is necessary.

http://docs.mathjax.org/en/latest/api/hub.html#Typeset
> Calls the preprocessors on the given element (or elements if it is an array of elements), and then typesets any math elements within the element. **If no element is provided, the whole document is processed.**

Since this is a component, I would recommend searching for the relevant text area you want to process within that component. Ember provides a jquery selector relative to the root of the given component instance, for example, via `this.$.()`.

abought:
Added a few notes, including some questions on ways to possibly improve rendering performance.

Particularly since some of the questions require seeing this in action, can you please update the PR description with a few examples of usage, to make testing this easier? (specific examples of content and what to look for)

app/controllers/submit.js
```diff
@@ -184,6 +184,15 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
         return this.get('isAdmin') && !(this.get('node.registration'));
     }),
 
+    renderMathJax: Ember.observer('uploadValid', 'node.title', function() {
+        Ember.run.next(this, function() {
+            //node.title gets changed on every character input, so debounce this.
+            Ember.run.debounce(this, function() {
```
abought: It seems that you really only want to render mathjax once- when the form section switches from input mode ("here are some textboxes") to preview mode ("here is what it would look like").

In that case, you may want a one-time typeset operation to be called as part of an action- observers might not be the mechanism you want at all. It shouldn't be necessary to debounce an operation on a field that only displays when the user is not able to edit the underlying value.

abought:
Pass complete. Will touch base with developer in AM.

app/index.html
```diff
@@ -15,6 +15,15 @@
     <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,600,300,700' rel='stylesheet' type='text/css'>
 
     {{content-for "head-footer"}}
+    <script type="text/javascript"
```
abought: The below links may not apply to this instance of manually editing index.html, because we will always want MathJax from a CDN (to support local development) instead of only using the library when in production. But noted for posterity: 
https://github.com/abought/ember-preprints/blob/31b35b85b6c84e15e1a9227e6e743079330c04e7/ember-cli-build.js#L36-L36

https://www.npmjs.com/package/ember-cli-inline-content

app/controllers/content.js
abought: Based on our discussion- ew, routes are way less smart than components. That is annoying.

There doesn't appear to be a direct hook for after dom elements are inserted, but there *is* potentially a more appropriate way to use `Ember.run`. Rather than waiting until the next runloop happens to run (and hoping things are set up the way you want), you can very explicitly identify when something should be run:

http://emberjs.com/api/classes/Ember.run.html#method_next
> ...Note that there are often alternatives to using run.next. For instance, if you'd like to schedule an operation to happen after all DOM element operations have completed within the current run loop, you can make use of the afterRender run loop queue...

app/index.html
hmoco: Yep, that should not cause any unknown performance hit, due to ` skipStartupTypeset`, meaning MathJax will be configured but not actually scan the page on startup.

app/controllers/content.js
hmoco: This is for the content page, so there is no ` node `, only the model. Since there is no editing on the content page, but it is dependent on a promise, this observer will fire once, when the model has been loaded.

abought: a) I'm not sure that's true- the model is the blank preprint, and some fields like DOI and subject are connected by two-way binding. (editing may happen even if saving does not)

b) Don't rely on a side effect of what you think will happen- be explicit.  If you only want something to fire once when the page loads, fire it in an appropriate event, not an observer. This will reduce the likelihood of future bugs, and as a bonus, will probably be more reliable for unit tests. (I've had problems in the past where unit tests of page rendering did not pick up on the observer in time, and moving the event to a rendering lifecycle hook fixed the failing test)

app/controllers/content.js
hmoco: Good call on using` schedule` over `next`!

hmoco:
OK, I think this is good for another look @abought, only thing I really couldn't see a good way to do was to pass a specific element to mathjax on the submit page.

app/components/search-result.js
```diff
@@ -8,7 +8,9 @@ export default Ember.Component.extend({
         PeerJ: /peerj/,
         arXiv: /arxivj/
     },
-
+    didRender() {
+        MathJax.Hub.Queue(['Typeset', MathJax.Hub, this.$()[0]]);  // jshint ignore:line
```
abought: Minor and not a blocker- but this should accept an array of elements. 
http://docs.mathjax.org/en/latest/api/hub.html#Typeset

We could just as easily only apply mathjax to the fields of interest, rather than everything under that search result.

app/controllers/submit.js
```diff
@@ -261,6 +268,10 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
             node.save()
                 .then(() => this.send('next', this.get('_names.2')))
                 .catch(()=> this.send('error', 'Could not save information; please try again'));
+            //Hack here because run.next and run.scheduleOnce afterRender dont pick up until the next edit.
+            Ember.run.later(this, function() {
+                MathJax.Hub.Queue(['Typeset', MathJax.Hub, Ember.$('#abstract')[0]]);  // jshint ignore:line
+            }, 300);
```
abought: How did you arrive at 300 ms as the time for this delay?

app/routes/content.js
```diff
@@ -7,6 +7,9 @@ export default Ember.Route.extend(ResetScrollMixin, {
     },
     setupController(controller, model) {
         controller.set('activeFile', model.get('primaryFile'));
+        Ember.run.scheduleOnce('afterRender', this, function() {
+            MathJax.Hub.Queue(['Typeset', MathJax.Hub]);  // jshint ignore:line
```
abought: Insofar as all DOM elements should be ready, does `Ember.$` not work here? This is rendering on the entire document, not just the parts of the content page of interest.

config/environment.js
```diff
@@ -25,8 +25,8 @@ module.exports = function(environment) {
             // when it is created
         },
         SHARE: {
-            baseUrl: process.env.SHARE_BASE_URL || 'https://staging-share.osf.io/',
-            searchUrl: process.env.SHARE_SEARCH_URL || 'https://staging-share.osf.io/api/search/abstractcreativework/_search'
+            baseUrl: process.env.SHARE_BASE_URL || 'https://share.osf.io/',
```
abought: I assume we want to do this, but can you confirm that this is an intentional change (vs accidental commit)?

Not sure how staging gets its config, so we should verify that the rule change doesn't affect QA (eg if they think that search is pointing at one server, and it suddenly is looking at somewhere else instead).

app/controllers/submit.js
abought: Also, any idea why these things don't pick up until after next edit? Just handwaving here, but is there for example some aspect of rendering that is affected by the save operation in line 268, and we're just waiting for the promise above to resolve?

config/environment.js
hmoco: Shoot, that is definitely not supposed to be there.

app/index.html
abought: Discussed with developer: there's a fair chance that we don't need MathML, but the file size is small enough to be manageable either way. 

We'll pare down in the future if performance becomes an issue later.

abought:
These testing notes were very helpful- thanks!

Overall it works for me and looks better. I do have two questions for the develop before moving on, esp wrt the 300ms delay. Hopefully these are quick to resolve.

app/controllers/submit.js
```diff
@@ -247,8 +252,10 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
             this.set('node.title', this.get('nodeTitle'));
             let node = this.get('node');
             node.save();
+            Ember.run.scheduleOnce('afterRender', this, function() {
+                MathJax.Hub.Queue(['Typeset', MathJax.Hub, Ember.$('.preprint-header-preview')[0]]);  // jshint ignore:line
```
abought: Not for this PR- but eventually, we'll want to DRY these selector names in one place when we clean up this page. 

This code is now closely tied to the markup of a template (just changing a style tag could affect how/if mathjax renders!), so we'll want to task this as part of that future cleanup.

app/routes/content.js
```diff
@@ -7,6 +7,9 @@ export default Ember.Route.extend(ResetScrollMixin, {
     },
     setupController(controller, model) {
         controller.set('activeFile', model.get('primaryFile'));
+        Ember.run.scheduleOnce('afterRender', this, function() {
+            MathJax.Hub.Queue(['Typeset', MathJax.Hub, [Ember.$('.abstract')[0], Ember.$('#preprintTitle')[0]]]);  // jshint ignore:line
```
abought: Why the mix of CSS classes and selectors here?

hmoco: Because `.abstract ` was already in place and specific enough, while preprintTitle didnt have anything already in place that was specific enough.

abought:
Overall it looks good and appears to work.

The markup is a bit weird and maintainability could be improved, but for a quick feature my blocking items have been addressed. I will move this forward for @icereval to look at.

icereval:
@caneruguz LGTM

PR #131, opened 2016-09-01
Fix search result links by separating hyperlinks [PREP-152] 

caneruguz:
JIRA Ticket: https://openscience.atlassian.net/browse/PREP-152

Cleans up the view of links so hyperlinks and information links are displayed separately. 

Bonus: Slight style changes that makes the search result box look much better
- heading font weight changed to normal
- border removed, different box shadow added 
screen shot 2016-09-01 at 3 58 28 pm


caneruguz:
Travis Check is about an error that happens in develop as well

PR #134, opened 2016-09-01
[rfr] Custom titles for each page of site

abought:
# Ticket
https://openscience.atlassian.net/browse/PREP-119

# Purpose
Instead of every page on the site having a title of "OSF | Preprints", allow each page on the site to add its own suffix- eg `OSF Preprints | Search` or `OSF Preprints | This preprint name` .

This will improve usability in the tab bar and back button history. Addon documentation here:
http://tim-evans.github.io/ember-page-title/

(screenshot is outdated; extra separator removed. See above text example)
screen shot 2016-09-01 at 5 17 47 pm

# Summary of changes
- Custom page titles
- Unrelated, but: Make date field names consistent with the capitalization used on the node model (and ember conventions)

### DIFF
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -1,4 +1,4 @@
-{{title 'OSF | Preprints' separator=' | '}}
+{{title 'OSF Preprints' separator=' | '}}
 
 {{osf-navbar signupUrl=signupUrl loginAction=(action 'login') hideSearch=true}}
 {{preprint-navbar}}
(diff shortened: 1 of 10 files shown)


app/templates/application.hbs
```diff
@@ -1,3 +1,5 @@
+{{title 'OSF | Preprints' separator=' | '}}
```
caneruguz: IS the | in between OSF and Preprints necessary? I realize that this is the structure for regular osf pages like ```OSF | Home``` but I think in this case, since OSF Preprints is a service and its title already has OSF, we can remove this.

abought: > IS the | in between OSF and Preprints necessary? I realize that this is the structure for regular osf pages like OSF | Home but I think in this case, since OSF Preprints is a service and its title already has OSF, we can remove this.

This makes sense. Ive pushed the requested change.

PR #135, opened 2016-09-02
[rfr] Index page should load even if ES is down

abought:
# Ticket
https://openscience.atlassian.net/browse/PREP-151

# Purpose
Allow the preprints app to load even when ES is down. On this page, elasticsearch is used specifically for the "x searchable counts" label that appears under the search box.

Also simplify the model hook on the index page, and don't show count box at all when ES doesn't load.

When ES is running:
screen shot 2016-09-02 at 11 32 48 am

When ES is down (or inaccessible due to network etc)
screen shot 2016-09-02 at 11 32 33 am

### DIFF
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -1,8 +1,6 @@
 import Ember from 'ember';
-import config from 'ember-get-config';
-import ResetScrollMixin from '../mixins/reset-scroll';
 
-const getTotalPayload = '{"size": 0, "from": 0,"query": {"bool": {"must": {"query_string": {"query": "*"}}, "filter": [{"term": {"type.raw": "preprint"}}]}}}';
+import ResetScrollMixin from '../mixins/reset-scroll';
 
 export default Ember.Route.extend(ResetScrollMixin, {
     model() {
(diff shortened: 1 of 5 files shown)


app/routes/index.js
```diff
@@ -2,31 +2,28 @@ import Ember from 'ember';
...
+        this._super(controller, model);
+        controller.set('currentDate', new Date());
 
-        return Ember.RSVP.hash({
-            theDate: new Date(),
-            subjects: this.store.query('taxonomy', { filter: { parents: 'null' }, page: { size: 20 } }),
-            sharePreprintsTotal: sharePreprintsTotal
-        });
+        // Fetch total number of preprints. Allow elasticsearch failure to pass silently.
+        Ember.$.ajax({
```
icereval: Would this make more sense to be in the controller? Seems a bit overreaching for the router to perform page display logic.

abought: Strictly speaking, this is in the `setupController` method, which sets the property on the controller (line 23)...

abought: Also, there's a general convention in ember for data fetching to be owned by the route. (esp wrt the data required for first page load). Even though we don't want the model hook to block on a failed request, it makes sense to keep the basic data in the same area instead of creating a separate file.

Also, since controllers are singletons, we want to rerun the network request on every visit and not simply serve back old state from a previous in-app visit to the page. The general recommendation is to do such cleanup in `setupController`, while crossing your fingers and waiting patiently for routable components...!

It's a good question; this is one of those cases where Ember is... fun.

icereval: We also never expect to recompute this value during the life of the sessions, its probably better to have it in the Controller's `init` method (http://emberjs.com/api/classes/Ember.Controller.html#method_init)

A quick glance at the travis-ci emberjs application for usages of `setupController` and `init`. It appears they are using authorization for `setupController` and one-time model property population from `init`.

abought: It sounds like this comes down to a question of how often we expect the SHARE count to increment. Since you're more familiar with that infrastructure, I've gone ahead and pushed the requested changes.

PR #136, opened 2016-09-02
[rfr] Contributor search: support filtering on multiple fields

abought:
# Ticket
https://openscience.atlassian.net/browse/PREP-90

# Purpose
Allow the `/submit` page "Authors -> search by name" box to search across a variety of name fields instead of just `full_name`.

See description of the multifield contributor search syntax here: https://github.com/CenterForOpenScience/osf.io/pull/6145

# How to test this
In the OSF, go to the settings page (`/settings`). Edit the profile fields so that `full_name`, `given_name`, `middle_names`, and `family_name` all have different values.

Then on the "add a preprint" (`/preprints/submit`) page, search for each of those values in turn. (taking care to search for someone else in between so that it's clear your search is doing something)

In the old search, the user would only show up if you matched the value in the full_name field. Now, a match for any of those fields will return the user, and the search result will display their `full name`.

### DIFF
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -272,19 +272,23 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
             this.send('next', this.get('_names.1'));
         },
         /**
-         * findContributors method.  Queries APIv2 users endpoint on full_name.  Fetches specified page of results.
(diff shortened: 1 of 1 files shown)


caneruguz:
Tested functionality and reviewed code.

PR #137, opened 2016-09-02
Share social media buttons

hmoco:
Adds social media share buttons to the view preprint page. Matches the behavior and buttons of public projects social media sharing. Uses liquid-fire to animate the toggling of the share row.


caneruguz:
Made new PR here: https://github.com/CenterForOpenScience/ember-preprints/pull/142

PR #138, opened 2016-09-02
[rfr] [as-is] Fix search page hanging when special characters present in query [PREP-160]

abought:
# Ticket
https://openscience.atlassian.net/browse/PREP-160

# Purpose
Fix an issue where certain characters with special meaning in elasticsearch were causing search page to hang. Reported multiple times on sentry post-announcement.

The root cause is that these characters could not be parsed in a query body according to this list:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters

Essentially: ES is returning a 200 response code, and informing the user of the failed query in the payload response. This means that the search page was hanging on a failed query.

# How to test this
Go to `/discover`. Search for something with a quotation mark: `anteaters united"`. Before the fix, there should be errors in the JS console, and the spinner should be stuck on loading.

After the fix, results should be returned.

# Notes for reviewers
I haven't worked with elasticsearch extensively- so even though this stops the query from failing, I'm not 100% sure that the meaning of the special characters is being correctly translated over to the search server. 

⚠️  This is worth a closer look on those grounds; I'm putting this up as-is before heading off.

### DIFF
--- a/app/controllers/discover.js
+++ b/app/controllers/discover.js
@@ -1,6 +1,8 @@
 import Ember from 'ember';
 import config from 'ember-get-config';
 
+import { elasticEscape } from '../utils/elastic-query';
+
 var getProvidersPayload = '{"from": 0,"query": {"bool": {"must": {"query_string": {"query": "*"}}, "filter": [{"term": {"type.raw": "preprint"}}]}},"aggregations": {"sources": {"terms": {"field": "sources.raw","size": 200}}}}';
 
 var filterMap = {
@@ -158,7 +160,7 @@ export default Ember.Controller.extend({
         }
         let query = {
             query_string: {
-                query: this.get('queryString') || '*'
+                query: elasticEscape(this.get('queryString')) || '*'
             }
         };
 
--- a/app/utils/elastic-query.js
+++ b/app/utils/elastic-query.js
@@ -1,0 +1,23 @@
+import Ember from 'ember';
+
+/**
+ * @module ember-preprints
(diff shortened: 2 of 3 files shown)


abought:
This has been hanging a while while waiting for additional product feedback on precisely that point- I'd be curious whether there was any clarity there?

Separately- per @chrisseto , it appears that whatever wraps ES queries isn't returning the error code when ES fails. (hence the 200 response code with a 400 in the response body) 

If the correct error code were returned we could show an error... but since we search as the user types, it would need some timing adjustments to avoid blitzing the user with error messages.

chrisseto:
@abought That's been fixed for a while
screen shot 2016-09-23 at 11 40 14
i made this for you

abought:
Good to hear! (this PR has been hanging for 21 days, so I'm not terribly surprised to see things have changed a bit in the interim)

caneruguz:
@abought Can you say more about what we need clarity on from product? I'll try to get it resolved.

abought:
This fix prevents elasticsearch from choking on a malformed query, by turning off the ability to use symbols that trigger special behavior in elasticsearch.

Essentially, this is a question of:
a) Whether we want the search box to be a simple text search field vs expose raw elasticsearch power
b) If we want powerful searches, should we tell the user? (so that if they get error messages they understand why)
c) Manual testing to verify that the error messages they receive for a search are clear, informative, and not too frequent

PR #154, opened 2016-09-13
Make sure filters actually have a value (even if the value might be bad)

hmoco:
## Purpose:
When pasting a link something like `subject=` gets converted to `subject`, which Ember reads as a value of ` true ` to the Subject query param. When trying to convert query params to filters, it is read as string ` true `. This prevents that behavior (but it does not validate the value of the query param if something is actually passed).

## Changes:
If a queryParam's value is ` true`, do not use it as a filter.


hmoco:
@chrisseto, is this still wanted? The errors seem to be unrelated to the PR, making it a feature and pointing it to develop will fix them.

chrisseto:
@hmoco still wanted. Ping @caneruguz

PR #157, opened 2016-09-14
Internationalization

binoculars:
Add i18n support and default english translation


app/locales/en/translations.js
```diff
@@ -0,0 +1,233 @@
+export default {
```
abought: Incidentally, the `ember-osf` addon (which provides some core features) has some translations of its own. Although the addon has too few translations to be noticeable yet, the goal was that the addon and consuming app would seamlessly combine their translations. 

https://github.com/CenterForOpenScience/ember-osf/tree/develop/addon/locales/en-us

In the addon, we use `en-us` for the locale name. It might be useful to pick a single consistent locale name going forward, so that all parts of the app can respect a single locale setting.

binoculars: The [ember-i18n wiki covers this](/jamesarosen/ember-i18n/wiki/Doc:-Defining-Translations#location-fallback)--see location fallback and region subtags

abought: Very nice! Thanks.

binoculars: I'd like to see us build up a common dictionary, then modularize the translations eventually. That would reduce the workload on the people actually doing the translations.

caneruguz:
@binoculars as we are talking about conventions for COS, you should try not to send a PR from your own ```develop``` branch. There is a lot of ways it gets messed up if it doesn't get merged in quickly. Can you close this PR and send again from a branch like ```feature/internationalization```

caneruguz:
@binoculars don't worry about resolving conflicts until initial code review is done, chances are there will be many changes that affect this work.

binoculars:
:ok: Closing this and sending from a feature branch

PR #160, opened 2016-09-15
Feature/i18n [PREP-182]

binoculars:
Reopening PR for feature branch instead of #157


app/locales/en/translations.js
```diff
@@ -0,0 +1,233 @@
...
+        add_preprint: `Add a preprint`,
+        title: `Title`,
+        search_preprints: `Search preprints...`,
+        added_on: `Added on`,
+        add: `Add`,
+        restart: `Restart`,
+        no_results_found: `No results found.`,
+        authors: `Authors`
+    },
+    _404: {
```
icereval: I could live with this either way but just to make a note you can also quote a key to handle numbers, spaces and special characters.

app/locales/en/translations.js
```diff
@@ -0,0 +1,233 @@
+export default {
+    global: {
+        share: `Share`,
```
icereval: See AirBnB's rules on template literals.
https://github.com/airbnb/javascript#strings--quotes

app/routes/application.js
```diff
@@ -4,4 +4,18 @@ import OSFAgnosticAuthRouteMixin from 'ember-osf/mixins/osf-agnostic-auth-route'
...
+
+        // Works in Chrome and Firefox (editable in settings)
+        if (navigator.languages && navigator.languages[0])
+            locale = navigator.languages[0];
+        // Backup for Safari (uses system settings)
+        else if (navigator.language)
+            locale = navigator.language;
+
+        if (locale)
+            this.set('i18n.locale', locale);
```
icereval: What does this do if there is no translation present?

config/environment.js
```diff
@@ -36,6 +36,9 @@ module.exports = function(environment) {
             // The name of the OSF campaign used to track signups
             campaign: 'osf-preprints'
         },
+        i18n: {
+            defaultLocale: 'en'
```
binoculars: The default locale is set here

app/locales/en/translations.js
binoculars: A bit of refactoring can be done to reduce duplication with string interpolation in the translation file--which is why I opted for template strings (AirBnB 6.4). Once we have a second translation, this will be easier to determine. Right now, it's a 1-for-1 substitution.

I was also originally planning for multi-line strings (didn't happen though).

Reducing escaping double or single quotes was another motivation. (AirBnB 6.6)

app/locales/en/translations.js
binoculars: Good point. I will change it to a quoted key. It's also worth pointing out that the 404 and page-not-found templates are identical.

app/routes/application.js
binoculars: The browser should have a default language setting enabled, if not, it will default to the English translation. Right now it's moot, because we only have English. See config/environment.js L40

app/locales/en/translations.js
icereval: good points

app/locales/en/translations.js
icereval: please enter a JIRA ticket to get this corrected.

app/routes/application.js
icereval: Was it verified if the browser does have a locale specified other than English and there is no localization file, it does 100% default to English?

icereval: One would expect it to, but I'd like to be certain before merge.

binoculars: Yes, I tested it locally with different languages stubbed out and changing the settings on Chrome, Firefox, and Safari.

[The ember-i18n wiki documents this feature](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Setting-the-Locale/c2e2a1177ab7b6ac9e6c0e07301a164482d90855)

PR #161, opened 2016-09-15
Add whitelist and put new name for research paper in economics [PREP-162]

hmoco:
Add a white list of providers that can be used as a filter. Also add filter for OSF -> OSF Preprints, Research Papers in Economics -> RePEc


caneruguz:
@hmoco can you confirm that this is prep-162?

PR #162, opened 2016-09-16
Add whitelist and put new name for research paper in economics [PREP-162] 

hmoco:
Add a white list of providers that can be used as a filter. Also add filter for OSF -> OSF Preprints, Research Papers in Economics -> RePEc

### DIFF
--- a/app/controllers/discover.js
+++ b/app/controllers/discover.js
@@ -78,13 +78,7 @@ export default Ember.Controller.extend({
             var hits = results.aggregations.sources.buckets;
             var providers = [];
             var whiteList = _this.get('whiteListedProviders');
-            hits.map(function(each) {
-                each = each.key;
-                console.log(each);
-                if (whiteList.indexOf(each) !== -1) {
-                    providers.push(each);
-                }
-            });
+            var providers = hits.map(each => each.key).filter(each => whiteList.indexOf(each) !== -1);
             _this.get('osfProviders').slice().map(function(each) {
                 if (providers.indexOf(each) === -1) {
                     providers.push(each);
--- a/app/controllers/discover.js
+++ b/app/controllers/discover.js
@@ -35,6 +35,7 @@ export default Ember.Controller.extend({
     activeFilters: { providers: [], subjects: [] },
     osfProviders: ['OSF', 'PsyArXiv', 'SocArXiv', 'engrXiv'],
 
+    whiteListedProviders: ['OSF', 'arXiv', 'bioRxiv', 'Cogprints', 'engrXiv', 'PeerJ', 'PsyArXiv', 'Research Papers in Economics', 'SocArXiv'],
     page: 1,
     size: 10,
     numberOfResults: 0,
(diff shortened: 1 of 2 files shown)


app/controllers/discover.js
```diff
@@ -76,15 +77,23 @@ export default Ember.Controller.extend({
         }).then(function(results) {
             var hits = results.aggregations.sources.buckets;
             var providers = [];
+            var whiteList = _this.get('whiteListedProviders');
             hits.map(function(each) {
-                providers.push(each.key);
+                each = each.key;
+                console.log(each);
```
caneruguz: Remove console.log's

app/controllers/discover.js
```diff
@@ -76,15 +77,23 @@ export default Ember.Controller.extend({
         }).then(function(results) {
             var hits = results.aggregations.sources.buckets;
             var providers = [];
+            var whiteList = _this.get('whiteListedProviders');
             hits.map(function(each) {
-                providers.push(each.key);
+                each = each.key;
```
caneruguz: is this optimization really necessary? looks brittle

app/controllers/discover.js
```diff
@@ -76,15 +77,23 @@ export default Ember.Controller.extend({
         }).then(function(results) {
             var hits = results.aggregations.sources.buckets;
             var providers = [];
+            var whiteList = _this.get('whiteListedProviders');
             hits.map(function(each) {
```
caneruguz: Is map the right thing to use here? how about ```.filter``` to filter out the results list? 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

app/helpers/filter-replace.js
```diff
@@ -2,7 +2,9 @@ import Ember from 'ember';
 
 var filters = {
     'Open Science Framework': 'OSF',
-    'Cognitive Sciences ePrint Archive': 'Cogprints'
+    'Cognitive Sciences ePrint Archive': 'Cogprints',
+    OSF: 'OSF Preprints',
```
caneruguz: Keep style consistent

app/controllers/discover.js
```diff
@@ -76,15 +77,23 @@ export default Ember.Controller.extend({
...
-                providers.push(each.key);
+                each = each.key;
+                console.log(each);
+                if (whiteList.indexOf(each) !== -1) {
+                    providers.push(each);
+                }
             });
             _this.get('osfProviders').slice().map(function(each) {
```
caneruguz: I know this is not part of PR but maybe you can join these two provider lists and do this step only once?

app/helpers/filter-replace.js
hmoco: This fails jshint, so the alternative would be:
```
'OSF': 'OSF Preprints'   // jshint ignore: line
```

app/controllers/discover.js
hmoco: Not really, because they have different behaviors. Whitelist shows things that should show if they exist, while the osfProviders list represents things that should be there regardless of whether there are documents for it.

app/helpers/filter-replace.js
caneruguz: that's funny, let's keep jshint happy.

app/helpers/filter-replace.js
```diff
@@ -2,7 +2,9 @@ import Ember from 'ember';
...
+    'Research Papers in Economics': 'ResPEc'
```
icereval: RePEc

PR #173, opened 2016-09-28
Feature/i18n fixes [PREP-182]

binoculars:
* Fix bug when changing browser language preferences to ensure that the change is only applied if the language is available in the locales (supposed to default to English if the browser's preferred language is not present).
* Fix references in popovers.


app/routes/application.js
```diff
@@ -6,16 +6,24 @@ import AnalyticsMixin from '../mixins/analytics-mixin';
...
+        }
         // Backup for Safari (uses system settings)
-        else if (navigator.language)
+        else if (navigator.language && availableLocales.contains(navigator.language)) {
             locale = navigator.language;
+        }
 
         if (locale)
-            this.set('i18n.locale', locale);
+            this.set('i18n.locale', navigator.language);
```
icereval: `locale`

PR #183, opened 2016-10-06
[Hotfix] Reset userNodesLoaded Field

pattisdr:
# Purpose

When userNodes initially loading, there is a spinner.  When you navigate away from page, and then return to add preprints page, state wasn't reset, so error message displayed that there were no user nodes that you could turn into a preprint.

# Changes
Adds userNodesLoaded field to the list of fields that are reset when navigating away and returning to Add Preprints page.


caneruguz:
I merged this through gitflow so hotfix is properly released. I'll ping Michael to deploy. Thanks @pattisdr

PR #184, opened 2016-10-11
Feature/Preprints Changes for GUID work [PREP-192]

pattisdr:
# Purpose

Preprints changes to work with @mfraezz's GUID work https://github.com/CenterForOpenScience/osf.io/pull/6346.  Includes @hmoco's work https://github.com/pattisdr/ember-preprints/pull/1.
# Changes
- when creating a preprint, send primaryFile, node, and provider as relationships. 
- subjects in the request are sent as a list of lists.
- Loads both node and preprint on the view page for preprints. (node loaded in the afterModel  hook - promise returned from model  to block the transition until loaded.  Need more work to get embeds working in the future).
- sets isPublished to true when creating preprint
- Henrique's GUID changes.  You can load /:preprint-id and it will load in the ember app.
- loads all providers and puts in store on submit page.
- Changes to still show discipline as flattened list.


pattisdr:
Both requested changes made.

PR #189, opened 2016-10-25
Feature/Edit Preprints [PREP-137]

pattisdr:
## Ticket

https://openscience.atlassian.net/browse/PREP-137

❗️ Dependent on: 
- ~~https://github.com/CenterForOpenScience/ember-preprints/pull/184 (GUID work, front-end)~~ Merged
- ~~https://github.com/CenterForOpenScience/osf.io/pull/6346, (GUID work, back-end)~~ Merged
- ~~https://github.com/CenterForOpenScience/ember-osf/pull/142 (companion PR, ember-osf)~~ Merged
- https://github.com/CenterForOpenScience/osf.io/pull/6475 (companion PR, osf)
## Purpose

Add ability to edit preprints.
## Changes
- [x] Edit buttons on view project and preprint pages if admin
- [x] Non admins attempting to access edit mode for a preprint will be redirected to Forbidden 
- [x] Changes to the way 'save' works in the Add/Edit form, all sections except authors save on 'Save and continue'.
- [x] Collapsed form sections on Add/Edit form reflect what is currently on the server.  When section is expanded, you see the pending, unsaved changes.
- [x] Discard changes buttons added to every section except Authors
- [x] Checkmark/x changed to red and green saved/unsaved icons.
- [x] Changes to add form so add/edit save similarly.  Preprint is actually creating in step 1 (upload) section of add form.  It is published on step 5 (submit).  When you add a discipline and hit save and continue, preprint is modified at that point.  If you add a doi, tag, and description and hit save and continue, two requests are sent, one to add the doi on the preprint, and the other to add tag/description on node.
- [x] User can only edit preprint file by uploading a new version of the file.  User can also edit title, abstract, tags, DOI, disciplines, authors, etc.  However, the actual location of the project is locked.
- [x] If there is an abandoned preprint (draft state, meaning user started the form and didn't finish, they can still use this node in the Add preprint form.  The abandoned preprint will be deleted in the background as they start a new one.  We are not supporting draft state at this time.
- [x]  Newly added contributors to preprint on Edit mode receive an email on 'Add'
- [x] Preprint/taxonomy/preprint-provider ember models, adapters, and serializers are moved to ember-osf out of ember-preprints.
- [x] `preprint_id/?edit` puts you in edit mode for a preprint (note query param).


app/templates/content.hbs
```diff
@@ -1,9 +1,12 @@
-{{title model.title}}
+{{title node.title}}
 
 <div id="view-page">
     <div class="p-t-xl m-t-xl p-b-md m-b-md dark-overlay-header-background header-row">{{!HEADER ROW}}
         <div class="container">{{!CONTAINER DIV}}
-            <h1 id="preprintTitle" class="p-b-md">{{model.title}}</h1>
+            <h1 id="preprintTitle" class="p-b-md">{{node.title}}</h1>
+            {{#if isAdmin}}
+                <a type="button" class="pull-right btn btn-success m-h-lg" href={{concat model.links.html '?edit'}}  {{action "click" 'link' "Content - Edit Preprint" (concat model.links.html '?edit')}}>{{t "content.project_button.edit_preprint"}}</a>
```
binoculars: Should use link-to with query params instead of a tag (see [Ember docs](https://guides.emberjs.com/v2.7.0/routing/query-params/#toc_link-to-helper) example)

app/controllers/submit.js
```diff
@@ -75,87 +106,188 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
...
+        if (node) {
+            return node.get('parent.id') ? false : true;
+        } else {
+            return null;
+        }
     }),
+
+    hasFile: Ember.computed('file', 'selectedFile', function() {
+        // True if file has either been preuploaded, or already uploaded file has been selected.
+        return this.get('file') !== null || this.get('selectedFile') !== null;
```
pattisdr: if there is both a file and a selectedFile, this will return false.

app/templates/components/preprint-navbar.hbs
```diff
@@ -9,7 +9,7 @@
         </div>
         <div id="preprint-navbar-links" class="navbar-collapse collapse navbar-right">
             <ul class="nav navbar-nav">
-                <li>{{#link-to "submit" class="" invokeAction=(action "click" "link" "Navbar - Add preprint")}}{{t "global.add_preprint"}}{{/link-to}}</li>
+                <li><a href='/preprints/submit' class="" invokeAction={{action "click" "link" "Navbar - Add preprint"}}>{{t "global.add_preprint"}}</a></li>
```
pattisdr: This seemed to alleviate some of the weirdness caused by the ?edit query param. I am open to other suggestions.  Add/Edit are sharing same template and controller.  But Add is submit route and Edit is content route.  If "Add Preprint" in navbar is a link-to helper, and you follow these steps, you will see weird behavior. 1) Add a preprint as normal.  Press 'Share' at bottom of page and then confirm in modal. You will be taken to Content page 2) Press "edit" button at top of content page. You will be taken to Edit page 3) Press Add a Preprint button in navbar (button in question).  Complete form as normal.  Click share at bottom. You will be in Edit template instead of content page.

Also back button behavior was weird with this Add Preprint button.  You would get `http://localhost:5000/g5yw4?edit=true` and then hitting back again would give you the correct `http://localhost:5000/g5yw4/?edit`

app/templates/content.hbs
pattisdr: @binoculars I had tried to use a link-to helper initially, but couldn't figure out how to make it work with the weird routing.

The issue is that we are on the content page here, and "edit" template also uses the content route.  So if it was something like this... `                {{#link-to "content" (query-params edit="true") class="pull-right btn btn-success m-h-lg" }}{{t "content.project_button.edit_preprint"}}{{/link-to}}, this button would just tack ?edit=true to the url but wouldn't reload the models and controllers, because we are already on the content page.  I need to force a reload so the controller/template for the submit page will be loaded instead.

app/components/file-uploader.js
```diff
@@ -145,45 +196,72 @@ export default Ember.Component.extend({
...
                 this.get('store')
                     .findRecord('file', resp.data.id.split('/')[1])
-                    .then(file => {
-                        this.set('osfFile', file);
-                        this.sendAction('finishUpload');
-                    });
+                        .then(file => {
+                            this.set('osfFile', file);
+                            if (this.get('nodeLocked')) { // Edit mode
+                                this.get('toast').info('Preprint file updated!');
```
binoculars: If you can set the text via handlebars template, that would be ideal e.g.

```hbs
{{if nodeLocked (t 'path.to.translation.obj')}}
```

Otherwise, as long as it can be changed based on the language (i.e. not just importing the english translation and using that) then it's all the same.

Looking at the docs, [Ember-i18n provides a macro to do it](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Translating-Text#translation-computed-property-macro).

app/templates/components/preprint-navbar.hbs
pattisdr: @binoculars and I looked at this today.  We haven't figured out a great way around it yet, so moving forward.  It technically works - but might be slower. The disadvantage to using an html 'a' tag is that everything is reloaded so you lose some of the benefits of a single page app.

app/templates/content.hbs
pattisdr: Again, from above, this issue was looked at today.  Moving forward as-is.  

"editing" preprints involves using the content route, but forcing the submit controller/template.  Having difficulties forcing setup, renderTemplate, and setupController to re-run with the query-param route.  Using <a> tag instead of link-to helper gets around this issue because it forces everything to reload.

app/controllers/submit.js
```diff
@@ -75,87 +106,188 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
...
     }),
+    abstractChanged: Ember.computed('basicsAbstract', 'node.description', function() {
+        // Does the pending abstract differ from the saved abstract in the db?
+        var basicsAbstract = this.get('basicsAbstract');
+        var nodeDescription = this.get('node.description');
+        var changed = false;
+        if (basicsAbstract !== null) {
+            changed = basicsAbstract.trim() !== nodeDescription;
+        }
+        return changed;
```
pattisdr: slight tweak with basicsAbstract !== null, so changing an abstract to an empty string will register as a change.

pattisdr:
Responded to comments and/or made changes.

PR #190, opened 2016-10-26
Feature/branding [PREP-185] [PREP-216] [PREP-217] [PREP-218] [PREP-219] [PREP-220] [PREP-221]  [PREP-222]  [PREP-223] [PREP-224] [PREP-243] [PREP-253]

binoculars:
[Branded Preprints Specification](https://docs.google.com/document/d/1NdoGVz0L6TQHn3kbXjaWUTzcp_dbG3mmvnNPdf2cEoU)
[UI Tweaks/Changes](https://docs.google.com/document/d/1FqInyDpkDreH-LkJ4KFzpL2bDeOZg_x8t5zEMuu6ncw)

Depends on:
- [x] GUID refactor CenterForOpenScience/osf.io#6346
- [x] API Changes CenterForOpenScience/osf.io#6532
- [ ] Branded campaign logins CenterForOpenScience/osf.io#6545
- [x] Ember-OSF Navbar Auth Dropdown Component CenterForOpenScience/ember-osf#143
- [x] Update to OSF Navbar CenterForOpenScience/ember-osf#149
- [x] Add ability to have a redirect url after logout CenterForOpenScience/ember-osf#151
- [x] GUID refactor for Ember #184 
- [x] Edit preprints #189


app/routes/content.js
```diff
@@ -25,8 +25,43 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
...
+        const findPreprint = this
+            .store
+            .findRecord('preprint', params.preprint_id);
+
+        // If we're already on a branded page, stay there.
+        if (this.get('theme.isProvider'))
+            return findPreprint;
+
+        // If we're on the OSF
+        return findPreprint
```
pattisdr: This redirect process from a `/:preprint_id` to a ` preprints/:provider/:preprint_id/ briefly flashes a `Page Not Found`.

app/routes/content.js
```diff
@@ -16,16 +17,51 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
...
```
pattisdr: This redirect process from a /:preprint_id to a preprints/:provider/:preprint_id/ briefly flashes a Page Not Found`.

pattisdr:
When you manually type some URL's and they redirect to the proper form, you can't hit the back button.  Maybe a replaceWith needed or similar? For example: typing in `preprints/preprint_id/?edit=true redirects to `/preprint_id/?edit=true` but you can't hit the back button because you get redirected right back.

app/components/subject-picker/component.js
```diff
@@ -57,16 +58,44 @@ export default Ember.Component.extend({
...
+                parents
             },
             page: {
                 size: 100
             }
-        }).then(results => this.set('_tier1', results));
+        }).then(results => {
+            const flatSubjects = this.get('flatSubjects');
+
+            if (flatSubjects.size) {
```
pattisdr: Might not be relevant now that the redirect would keep the user from viewing the preprint under a different provider.

app/templates/index.hbs
```diff
@@ -24,9 +24,10 @@
             <div class="row">
                 <div class="text-center col-xs-12">
                     <p class="lead"> or </p>
-                    {{#link-to "submit" type="button" class="btn btn-success btn-lg" invokeAction=(action "click" "link" "Index - Add preprint")}}{{t "global.add_preprint"}}{{/link-to}}
-                    <div class="m-t-md">{{#link-to 'content' 'khbvy' invokeAction=(action "click" "link" "Index - See example")}}{{t "index.header.example"}}{{/link-to}}</div>
-
+                    {{#link-to (if theme.isProvider "provider.submit" "submit") type="button" class="btn btn-success btn-lg" invokeAction=(action "click" "link" "Index - Add preprint")}}
+                        {{t "global.add_preprint"}}
+                    {{/link-to}}
+                    <div class="m-t-md">{{#link-to (if theme.isProvider "provider.content" "content") 'khbvy' invokeAction=(action "click" "link" "Index - See example")}}{{t "index.header.example"}}{{/link-to}}</div>
```
binoculars: It's just hidden because the example link is hard coded to an OSF preprint, so this would take you out of the branded site, which I don't believe is the desired behavior. It should still be there for OSF Preprints. Need feedback from product on this one.

app/templates/components/preprint-footer-branded.hbs
```diff
@@ -0,0 +1,42 @@
...
+        <span>
+            <a href="mailto:{{if model.emailContact model.emailContact (concat 'support+' model.id '@osf.io')}}"
+               onbeforeclick={{action 'click' 'link' 'Branded Footer - Contact'}}
+            >
+                {{t "components.preprint-footer-branded.contact"}}
+            </a>
+        </span>
+    </div>
+    <div style="display: inline; margin-left: 8px;">
+        <a href="https://twitter.com/{{model.socialTwitter}}"
```
binoculars: I talked about this with Nici, and I think it was a requirement for the providers to have all three, but either way, I like your approach better. It should not be displayed if the model doesn't have the field.

pattisdr:
@binoculars, one other point I neglected to mention.  The Add form (submit page), allows you to connect a preprint to an existing OSF project.  

screen shot 2016-11-04 at 3 44 43 pm

The logic [here](https://github.com/CenterForOpenScience/ember-preprints/pull/190/files#diff-372903a5a43d23d620e2beffc5caa0b6R30) is going to have to change.  Prior to branded, the only nodes that could connected to preprints were nodes without published preprints (because there was only one provider).  So to populate the node dropdown, `v2/users/me/nodes/?filter[preprint]=false` , was sufficient.

However, I double-checked with @mfraezz who says `Node` - `PreprintService` is one to many, `PreprintService -PreprintProvider` is one to one.  So there can be multiple preprints on a node, each one associated with a different preprint provider.  The current code will not allow you to create multiple preprints on a node.

@mfraezz suggests "embedding preprints on nodes and calculating the providers that a preprint can be made on that node for".

app/templates/components/preprint-footer-branded.hbs
```diff
@@ -0,0 +1,42 @@
...
+        <span>
+            <a href="mailto:{{if model.emailSupport model.emailSupport (concat 'contact+' model.id '@osf.io')}}"
+               onbeforeclick={{action 'click' 'link' 'Branded Footer - Support'}}
+            >
+                {{t "components.preprint-footer-branded.support"}}
+            </a>
+        </span>
+        <span> | </span>
+        <span>
+            <a href="mailto:{{if model.emailContact model.emailContact (concat 'support+' model.id '@osf.io')}}"
```
binoculars: Nice catch, I also forgot to i18n-ize the aria-labels as well. Will change both.

app/components/subject-picker/component.js
```diff
@@ -57,16 +58,44 @@ export default Ember.Component.extend({
     selection2: null,
     selection3: null,
 
-    init() {
-        this._super(...arguments);
+    flatSubjects: Ember.computed('theme.provider.subjectsAcceptable', function() {
```
binoculars: Good idea, I'll move into the theme service and reference it from there

pattisdr:
Some things still look weird with the edit button but I don't think this should hold up the demo.  Maybe just some more UI polish after it's on staging.
screen shot 2016-11-07 at 3 12 53 pm
screen shot 2016-11-07 at 3 12 40 pm

app/templates/components/preprint-navbar-branded.hbs
```diff
@@ -0,0 +1,18 @@
+<div class="preprint-navbar navbar">
+    <div class="container">
+        <div class="navbar-header">
```
pattisdr: The hover looks very nice.

app/templates/components/preprint-navbar-branded.hbs
```diff
@@ -0,0 +1,18 @@
...
+                <span class="navbar-image"></span>
+                <span class="navbar-title">{{model.name}} {{t "global.preprints"}}</span>
+            {{~/link-to}}
+        </div>
+        <div id="preprint-navbar-links" class="navbar-collapse collapse navbar-right">
+            <ul class="nav navbar-nav">
+                <li><a href="/myprojects/" class="" onbeforeclick={{action "click" "link" "Navbar - My OSF Projects"}}>{{t "components.preprint-navbar-branded.my_projects"}}</a></li>
+                <li>{{#link-to "provider.submit" class="" invokeAction=(action "click" "link" "Navbar - Add preprint")}}{{t "global.add_preprint"}}{{/link-to}}</li>
+                <li>{{#link-to "provider.discover" class="" invokeAction=(action "click" "link" "Navbar - Search")}}{{t "global.search"}}{{/link-to}}</li>
+                {{navbar-auth-dropdown loginAction=(action 'login') headline=(t "components.preprint-navbar-branded.headline")}}
```
pattisdr: Changes look great 
screen shot 2016-11-07 at 3 20 15 pm

app/routes/content.js
```diff
@@ -48,31 +48,35 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
...
+                const {origin, search} = window.location;
+
+                const urlParts = [
+                    origin
+                ];
+
+                if (!isOSF)
+                    urlParts.push('preprints', providerId);
+
+                urlParts.push(`${preprint.get('id')}${search}`);
```
pattisdr: @binoculars I would put a slash in between the preprint id and the search item.  Otherwise I am seeing `/preprints/socarxiv/m86jw/?edit` redirect to `/m86jw?edit` redirect to `m86jw/?edit`.
```js
urlParts.push(`${preprint.get('id')}/${search}`);
```

pattisdr:
@binoculars - changes look good to me, nice work. No remaining blockers on my end.

package.json
```diff
@@ -58,7 +58,7 @@
     "ember-link-action": "0.0.35",
     "ember-load-initializers": "^0.5.1",
     "ember-metrics": "0.6.3",
-    "ember-osf": "git+https://github.com/CenterForOpenScience/ember-osf.git#a3b972d0537cfed5d18f6bfc67feb6e33b68a64a",
+    "ember-osf": "git+https://github.com/CenterForOpenScience/ember-osf.git#687500f51a18fa8b829105126c12310fc7ed6f1a",
```
icereval: b8d2122ea296d2113b2ed98de9b8c2568e313bc2

icereval:
👍  - merged to feature/staging-2 and deployed

app/routes/content.js
```diff
@@ -100,8 +135,24 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
...
+            } else {
+                const path = ['', 'preprints'];
+
+                if (this.get('theme.isProvider'))
+                    path.push(this.get('theme.id'));
+
+                path.push(slug);
+
+                window.history.replaceState({}, 'preprints', path.join('/'));
+                this.intermediateTransitionTo('page-not-found');
```
binoculars: This route is reused in routes/provider/content.js, and we have to handle a few different cases with this.

1. `/:guid` redirects to 3, if the provider is not OSF
2. `/preprints/:guid` always redirects to 1
3. `/preprints/:provider/:guid` redirects to 1, if the provider is OSF or to the correct provider if the it's on the wrong provider (e.g. `/preprints/socarxiv/abc12` -> `/preprints/engrxiv/abc12`)

I was running into cases in testing bad subroutes wouldn't be handled correctly, so this was necessary. I really just wanted to make sure we handle it in case a weird edge case comes up where the slug is not 5 characters.

app/routes/provider.js
```diff
@@ -0,0 +1,35 @@
+import Ember from 'ember';
+import config from 'ember-get-config';
+
+export default Ember.Route.extend({
+    theme: Ember.inject.service(),
+
+    providerIds: config.PREPRINTS.providers.map(provider => provider.id),
```
binoculars: Before, I had was using findAll in the `init` function on the route, but that created a race condition where the promise in `init` wouldn't resolve before `beforeModel` was called. Since we have to statically build the CSS anyway for each provider, it only makes sense to just add them into the config. This way, they're available immediately to the ember app instead of waiting for a promise to resolve (or error) if the network is slow. In phase 2, we're going to have to detect which domain we're on, and the only way to really do that to determine which routes to use is to have a config with the domains.

app/routes/provider.js
```diff
@@ -0,0 +1,35 @@
...
+
+        if (this.get('providerIds').includes(slug)) {
+            this.set('theme.id', slug);
+        } else {
+            this.set('theme.id', config.PREPRINTS.provider);
+
+            if (slug.length === 5) {
+                this.transitionTo('content', slug);
+            } else {
+                this.replaceWith('page-not-found');
```
binoculars: What's the expected behavior?

pattisdr: I am not sure - there is just inconsistent behavior.  A lot of your transitions to `page-not-found` keep the problematic url in the location bar, which I personally like, because I can then look at the url and see what I typed in incorrectly.  If this line is hit, the location bar has the `/preprints/page-not-found` there.

app/routes/provider.js
pattisdr: Makes sense.

app/routes/provider.js
pattisdr: @binoculars and I discussed - we don't know a good way to resolve, and because this is minor, leaving as-is.

pattisdr:
Thanks for addressing my comments @binoculars - all good on my end.

binoculars:
Ready to merge @icereval any objections?

brianjgeiger:
Should be ready once the last changes to CenterForOpenScience/osf.io#6545 are done and merged.

binoculars:
Adding BePress Taxonomies per [PREP-185](https://openscience.atlassian.net/browse/PREP-185) (BePress migration is a blocker to branding)

PR #192, opened 2016-11-02
[hotfix] Reroute to page-not-found when non-guid-like string is passed to reroute-guid.

hmoco:
## Purpose
Currently when attempting to access `osf.io/preprints/not-guid`, (assuming it is not an existing page such as discover), the app loads the redirect route, which contains nothing. This redirects to the default page-not-found in the ember app.

## Changes
When trying to view `/preprints/not-a-guid/ `, the ember app will transition to page-not-found.


caneruguz:
@hmoco can you add description to the PR please? Is there a JIRA ticket for it?

hmoco:
This was something that came up on flowdock, I don't believe a ticket was ever made for it.

PR #193, opened 2016-11-04
Feature/Edit Preprint Improvements [PREP-137] 

pattisdr:
# Purpose

Issues found by QA for Edit Preprints
https://openscience.atlassian.net/browse/PREP-137

# Changes

- Collapsed state shows file version
- If user uploads unchanged version of paper, on 'Save and continue' version will revert back to proper version.
- DOI can be submitted as either the DOI or the DOI link
- If inputting a DOI link, when the input box loses focus, the displayed input will be stripped to just the DOI itself.
- Restore handling of orphans just in case old orphan slips through.

### DIFF
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -73,6 +73,17 @@ function subjectIdMap(subjectArray) {
     return subjectArray.map(subjectBlock => subjectBlock.map(subject => subject.id));
 }
 
+function doiRegexExec(doi) {
+    //Strips url out of inputted doi, if any.  For example, user input this DOI: https://dx.doi.org/10.12345/hello. Returns 10.12345/hello.
+    const doiRegex = /\b(10\.\d{4,}(?:\.\d+)*\/\S+(?:(?!["&\'<>])\S))\b/;
+    if (doi) {
+        const doiOnly = doiRegex.exec(doi);
+        return doiOnly !== null ? doiOnly[0] : null;
+    }
+    return doi;
+
+}
+
 /**
  * "Add preprint" page definitions
  */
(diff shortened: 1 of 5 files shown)


app/controllers/submit.js
```diff
@@ -262,7 +273,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
     doiChanged: Ember.computed('model.doi', 'basicsDOI', function() {
         // Does the pending DOI differ from the saved DOI in the db?
         // If pending DOI and saved DOI are both falsy values, doi has not changed.
-        return (this.get('basicsDOI') || this.get('model.doi')) ? this.get('basicsDOI') !== this.get('model.doi') : false;
+        return doiRegexExec(this.get('basicsDOI') || this.get('model.doi')) ? doiRegexExec(this.get('basicsDOI')) !== this.get('model.doi') : false;
```
binoculars: This should probably be simplified a bit. (Let's discuss possibilities).

app/controllers/submit.js
```diff
@@ -540,7 +551,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
             node.save()
                 .then(() => {
                     if (this.get('doiChanged')) {
-                        model.set('doi', this.get('basicsDOI'));
+                        model.set('doi', doiRegexExec(this.get('basicsDOI')));
```
binoculars:
```js
model.set('doi', doiRegexExec(this.get('basicsDOI')) || null);
```
and remove the if statement on L555-L557

app/controllers/submit.js
```diff
@@ -73,6 +73,17 @@ function subjectIdMap(subjectArray) {
     return subjectArray.map(subjectBlock => subjectBlock.map(subject => subject.id));
 }
 
+function doiRegexExec(doi) {
+    //Strips url out of inputted doi, if any.  For example, user input this DOI: https://dx.doi.org/10.12345/hello. Returns 10.12345/hello.
+    const doiRegex = /\b(10\.\d{4,}(?:\.\d+)*\/\S+(?:(?!["&\'<>])\S))\b/;
```
binoculars: Why not just use string.prototype.replace with regex here instead of matching?

app/controllers/submit.js
binoculars:
```js
        const basicsDOI = doiRegexExec(this.get('basicsDOI'));
        const modelDOI = this.get('model.doi');
        
        return (basicsDOI || modelDOI) && basicsDOI !== modelDOI;
```

app/controllers/submit.js
binoculars:
```js
return (doi || '').replace(/^https:\/\/dx\.doi\.org\//, '');
```

pattisdr: Used your idea, to replace the input when the box loses focus.  I fire it `onchange`. Still using `regexObj.exec(str)` because of the different combinations the url could take (http, https..).  However, if there is no match, I don't change the errored DOI, I leave as is.  If there is a match, I replace the input.

pattisdr:
@binoculars Changes made.

PR #195, opened 2016-11-10
[HOLD] Feature/Multiple Preprints Per Node [PREP-242]

pattisdr:
# Purpose
❗️ HOLD discussion of requirements needed.
❗️ Requires companion ember-osf PR https://github.com/CenterForOpenScience/ember-osf/pull/148

The backend permits one preprint per provider per node.  This PR updates the UI to match.

https://openscience.atlassian.net/browse/PREP-242
# Changes
- Node dropdown formerly uses query users/nodes/?filter[preprint]=false, looking for nodes that don't have preprints.  New dropdown will allow an existing node to be chosen as long as it doesn't have a published preprint for that provider.  Done with query users/nodes/?embed=preprints and then looking through the embedded preprints to make sure the current provider doesn't have a published preprint on that node.
- Warning messages when in Add or Edit mode if a published preprint from another provider currently exists.  Messages show on fields that are stored on the node and will therefore be reflected on all preprints stored there (abstract, keywords, file, contributors, title).  Not discipline/doi.
- File is locked from changing if another published preprint exists on that node.


pattisdr:
Closing - looks like we are going in a different direction. Can reopen if things change.

PR #200, opened 2016-11-28
Feature/Rewrite get-ancestor-descriptor helper [PREP-257]

pattisdr:
# Ticket 
https://openscience.atlassian.net/browse/PREP-257
❗️ Requires that https://github.com/CenterForOpenScience/ember-osf/pull/148 is merged.

# Purpose

On the `Add Preprint` page, you have the option to connect your preprint to an existing project.   The `get-ancestor-descriptor` helper allows us to format the projects in the project dropdown similarly to how they are formatted in the dashboard, with a hierarchy, so users can distinguish between similarly-named projects.  However, there are some errors in the logic, particularly when a project in the hierarchy is private.

screen shot 2016-11-28 at 8 21 06 am

# Changes
- Rewrites get-ancestor-descriptor so logic is easier to follow.
  - If parentId, rootId, parentparentId's are undefined  (case if private or does not exist), attempts to pull id from relationship link (in which case it is private). If link DNE, then relationship DNE.
  - Defines parentId, rootId, parentParentId, parentTitle, and rootTitle first.  If parentTitle or rootTitle DNE, changes to 'Private'
  - Then checks if node is one level, two levels, three levels, or four+ levels deep and the returns the appropriate ancestor descriptor accordingly.
- Better handles private projects
- Adds better tests to get-ancestor-descriptor

# For QA

The way dropdown is populated was rewritten, so you should re-test whatever you tested before for this.  Most behavior should be the same, but there were errors in the way that 'private' projects in the hierarchy were being displayed, mostly in the number of levels you would see.

Errors that have been fixed:

- Create a project "Parent", and then a component inside this project, "Child".  Then add another admin contributor to `Parent` and remove yourself.  So you only have access to "Child", but not it's ancestor, "Parent".  Project dropdown should read: ` Private / Child `.  Prior to this fix, it read only ` Child `, leading you to think "Child" was top-level.

- Create four levels of projects "Apple" > "Banana" > "Cantaloupe" > "Date".  Remove yourself as a contributor from "Apple" and "Cantaloupe".  Because you are still a contributor on "Banana", and therefore can access 'Cantaloupe',  dropdown for 'Date', reads `Private / ... / Cantaloupe / Date`.  Old version was `Private / Cantaloupe / Date`, which is incorrect, because grandparent of 'Date' is not private, and items in the hierarchy are missing.

### DIFF
--- a/app/helpers/get-ancestor-descriptor.js
+++ b/app/helpers/get-ancestor-descriptor.js
@@ -1,42 +1,51 @@
 import Ember from 'ember';
 
-// Formats titles similar to the way they're displayed in the dashboard.  For example, Root Name / ... / Parent Name / Node Name.
+function fetchIdFromRelationshipLink(node, relationship) {
+    // Private node ids can be accessed under initializedRelationships.
+    if (node) {
+        let relationships = node._internalModel._relationships.initializedRelationships[relationship];
+        if (relationships && relationships.link) {
+            return relationships.link.split('nodes')[1].replace(/\//g, '');
+        }
+    }
+   return undefined;
+}
+
+function fetchTitle(node, relationship) {
+    // Fetches parent or root title.  If null, marks 'Private'.
+    let title = node.get(`${relationship}.title`);
+    if (typeof title === 'undefined') {
+        title = 'Private';
+    }
+    return title;
+
+}
+
 export function getAncestorDescriptor(params/*, hash*/) {
-    var node = params[0];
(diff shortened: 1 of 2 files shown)


brianjgeiger:
Oh, and CenterForOpenScience/ember-osf#148  is merged, so this (or another) PR will require that the version being called matches the latest EOSF commit.

pattisdr:
@binoculars ^ comment made my day :)

PR #201, opened 2016-11-29
Branding Phase 2 - Provider Domains [PREP-259]

binoculars:
See [the spec](https://docs.google.com/document/d/1NdoGVz0L6TQHn3kbXjaWUTzcp_dbG3mmvnNPdf2cEoU/) for details.

Companion to:
- CenterForOpenScience/osf.io#6793 (Adds domain field, redirects)
- CenterForOpenScience/ember-osf#173 (Adds domain field to the model)

## Ticket

https://openscience.atlassian.net/browse/EOSF-327

## TODO:
- [x] Make the routes under `/preprints/:provider/:whatever` redirect to `provider.domain/:whatever`
- [x] Update all redirects to go to the correct place, the first time (no multiple redirects, where possible).
- [x] Get tests passing
- [x] Provide guidance for staging and prod (Update: DevOps ticket in Jira)

## Notes for Reviewers:
If testing this PR without the changes from OSF, set environment variable `ENABLE_PROVIDER_DOMAINS='true'`

See the [README section](https://github.com/binoculars/ember-preprints/tree/f2261222e090eb04ec1cca1cb59b82a8d90a4c87#provider-domains) on how to populate the domains locally. The domains are defined in the ember config (config/environment.js). TLDR: `sudo ./scripts/add-domains.js`

- You should be able to go to a provider such as http://local.socarxiv.org:4200/ (preprints container or ember serve must be running.
- Going to http://localhost:5000/preprints/socarxiv should redirect you to http://local.socarxiv.org:4200/
- If you create a preprint under a provider and try to go to it from http://localhost:5000/abc12, you should be redirected to http://local.socarxiv.org:4200/abc12. Note: GUID redirects will be handled by osf.io in CenterForOpenScience/osf.io#6793 and /preprints/socarxiv should be redirected by the proxy in staging/production. The redirects here are primarily to handle anything that falls through the cracks.
- Please click through each portion of the site and verify that the links work correctly.
- If you comment out one of the domains (other than osf.io) in config/environment.js, you should be able to go back to using http://localhost:5000/preprint/provider style routing. Please verify that everything still works as expected here.

## Notes for QA
- This is mostly a code change with the exception of how the routing is handled. Once this is up on staging, all of the redirects should work as expected.


app/templates/components/taxonomy-top-list.hbs
```diff
@@ -2,7 +2,7 @@
     <div class="subject row m-b-sm">
         {{#each pair as |subject|}}
             <div class="col-xs-12 col-sm-6 subject-item {{if (get subjectTooltips subject.text) 'hint--bottom hint--large'}}" aria-label={{get subjectTooltips subject.text}}>
-                {{#link-to (if theme.isProvider 'provider.discover' 'discover') (query-params subjectFilter=subject.text) class='btn btn-primary p-sm' invokeAction=(action "click" "link" (concat 'Preprints - Index - Browse ' subject.text))}}
+                {{#link-to (route-prefix 'discover') (query-params subjectFilter=subject.text) class='btn btn-primary p-sm' invokeAction=(action "click" "link" (concat 'Preprints - Index - Browse ' subject.text))}}
```
abought: This is a somewhat gnarly requirement. Have you taken a look at `Route#renderTemplate`? It seems to provide a way for one route to use the template or behavior of another. I'm not totally sure if it applies here, but might spark some ideas.

http://emberjs.com/api/classes/Ember.Route.html#method_renderTemplate

scripts/add-domains.js
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env node
```
abought: No review performed on this script in initial round.

app/router.js
```diff
@@ -1,12 +1,37 @@
...
+    .PREPRINTS
+    .providers
+    // Exclude OSF
+    .slice(1)
+    // Filter out providers without a domain
+    .filter(p => p.domain)
+    .find(p =>
+        // Check if the hostname includes: the domain, the domain with dashes instead of periods, or just the id
+        hostname.includes(p.domain) ||
+        hostname.includes(p.domain.replace(/\./g, '-')) ||
```
hmoco: When is this relevant, out of curiosity?

app/router.js
```diff
@@ -25,17 +50,25 @@ const Router = Ember.Router.extend({
...
-    this.route('index', {path: 'preprints'});
-    this.route('page-not-found', {path: 'preprints/page-not-found'});
-    this.route('submit', {path: 'preprints/submit'});
-    this.route('discover', {path: 'preprints/discover'});
-    this.route('content', {path: '/:preprint_id' });
-    this.route('provider', {path: 'preprints/:slug'}, function() {
-        this.route('content', {path: '/:preprint_id'});
-        this.route('discover');
+
+    if (provider) {
```
hmoco: Does provider get recalculated properly when window.location changes? I would assume it does, since it works, but it seems weird that a const value changed would cause this Router.map to be rerun, assuming the preprint app is still a singleton across the different providers

app/services/theme.js
```diff
@@ -16,10 +16,15 @@ export default Ember.Service.extend({
     store: Ember.inject.service(),
     session: Ember.inject.service(),
 
+    // If we're using a provider domain
+    isDomain: false,
```
hmoco: Not the biggest fan of this, could we possibly change it to something like `hasDomain`, `isProvider`, 'onProvider`, `onProviderDomain`? Don't feel strongly about any of these, but was initially confused by isDomain (although the comment is super helpful)

app/services/theme.js
```diff
@@ -31,21 +36,54 @@ export default Ember.Service.extend({
...
+    }),
+
+    pathPrefix: Ember.computed('isProvider', 'isDomain', 'id', function() {
+        let pathPrefix = '/';
+
+        if (!this.get('isDomain')) {
+            pathPrefix += 'preprints/';
+
+            if (this.get('isProvider')) {
+                pathPrefix += `${this.get('id')}/`;
```
hmoco: wait, just realized that there's also something called `isProvider`. When would `isDomain` and `isProvider` not both be the same?

app/templates/forbidden.hbs
```diff
@@ -1,27 +1,4 @@
...
-                    {{/link-to}}
-                {{/if}}
-            </div>
-        </div>
-    </div>
-</div>
+{{error-page
+    label='Forbidden'
+    translateKey='page-forbidden'
+}}
```
hmoco: Really like this!

scripts/add-domains.js
```diff
@@ -0,0 +1,48 @@
...
+
+const {providers} = config.PREPRINTS;
+const hostsFileName = '/etc/hosts';
+const hostIP = '127.0.0.1';
+const hostsFile = fs.readFileSync(hostsFileName, {encoding: 'utf8'});
+const sectionHeader = '## EMBER-PREPRINTS ##\n';
+const sectionFooter = '\n## /EMBER-PREPRINTS ##';
+const rgx = new RegExp(`(?:${sectionHeader})(.|\\s)*(?:${sectionFooter})`, 'm');
+const domainProviders = providers
+    .slice(1)
```
hmoco: Throughout this PR, there seems to be a lot of `providers.slice(1)`. Is there any use for a separate config var that defines non-OSF providers?

hmoco:
Functional testing looked good, although I was unable to test the redirect from `/preprints/<provider>` to provider domain.

app/router.js
binoculars: Since changing the provider requires a full page reload, the entire app, including the router, will be evaluated each time.

hmoco:
Also, after reading through this, not really a request for a change, but a quick note for both future development and potential refactoring: our usage of the word `provider` seems to be incredibly broad, including OSF at times. It leads to slightly confusing code to do it being not descriptive or precise enough.

app/services/theme.js
binoculars: Some of the providers will not use their own domain, and will stay under osf.io/preprints/whatever

`isProvider` is used to determine if the provider is OSF (for our default case) or not
`isDomain` is whether the provider is using a domain or not

scripts/add-domains.js
binoculars: I was trying to avoid modifying the original config too much, which would have downstream changes--not ideal, but neither is creating a ton of merge conflicts if it can be avoided.

app/services/theme.js
hmoco: Is there a case in which the provider being used has a domain but is not using it? 

`isProvider` is true if, and only if, the current provider is not OSF?

app/router.js
binoculars: I had a conversation with @icereval a while back about how we might deploy this on staging/test. `agrixiv-org.staging.osf.io` was a likely naming scheme. Putting a dot in a subdomain creates another subdomain.

app/services/theme.js
binoculars: If the domain exists in the config (and domains are enabled), then using the domain is used. It has to be one or the other due to the requirements.

Yes, `isProvider` basically means not OSF.

binoculars:
I got set up locally and updated my osf.io PR. I would like avoid updating that again until we're serious about merging, because it needs a migration.

My local copy is redirecting, so I'm not sure what's causing it to not redirect. It could be a container sync issue.

I would like to come up with better semantics for provider. At this point, we've been using it in branded preprints since that was launched, so a refactor would need to be planned out.

brianjgeiger:
I think in the OSF world, a `Provider` is a third party that has information, code, or service that is being served via the OSF but is not generally available to everyone. So `Users` aren't `Providers` because anyone can be a user, but `PreprintProvider`, `AddonProvider`, `RegistryProvider`, and so on would be valid Providers (whether or not all of those things currently exist). Do we have uses that are not currently following that convention?

binoculars:
I'm not aware of any, but it would good to codify that somwhere

PR #202, opened 2016-11-29
Feature/Redirect to Resource Deleted if 410 [PREP-260]

pattisdr:
# Ticket 

https://openscience.atlassian.net/browse/PREP-260

# Purpose

If you delete a node that contains a preprint, when you try to navigate to that preprint it will get stuck in an infinite loop.

# Changes

- Adds resource-deleted route (with GA tracking and branding-specific)
- Redirects to resource-deleted route if the request to fetch the node returns a 410 Gone (actually checking error message details, can't get the status code Ember-side)
 - Does not modify url in location bar when redirects - want to see which url had error

# Screenshots - OSF and Psyarxiv

screen shot 2016-11-29 at 4 20 16 pm

screen shot 2016-11-29 at 4 20 23 pm

### DIFF
--- a/app/routes/content.js
+++ b/app/routes/content.js
@@ -135,23 +135,27 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
     },
     actions: {
         error(error, transition) {
-            const slug = transition.params[transition.targetName].preprint_id;
-
-            if (slug.length === 5) {
-                window.location.href = [
-                    window.location.origin,
-                    slug
-                ].join('/');
+            if (error && error.errors && Ember.isArray(error.errors) && error.errors[0].detail === 'The requested node is no longer available.') {
+                this.intermediateTransitionTo('resource-deleted'); // Node containing preprint has been deleted. 410 Gone.
             } else {
-                const path = ['', 'preprints'];
+                const slug = transition.params[transition.targetName].preprint_id;
 
-                if (this.get('theme.isProvider'))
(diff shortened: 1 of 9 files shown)


app/routes/content.js
```diff
@@ -135,23 +135,27 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
...
-                    window.location.origin,
-                    slug
-                ].join('/');
+            if (error && error.errors && Ember.isArray(error.errors) && error.errors[0].detail === 'The requested node is no longer available.') {
+                this.intermediateTransitionTo('resource-deleted'); // Node containing preprint has been deleted. 410 Gone.
             } else {
-                const path = ['', 'preprints'];
+                const slug = transition.params[transition.targetName].preprint_id;
+
+                if (slug.length === 5) {
```
brianjgeiger: Is this how we're determining it's a GUID?

pattisdr: I think so - @binoculars can explain more. I just moved this logic into an else branch, I didn't change anything else.

brianjgeiger: We'll probably re-visit that one day, but should be fine for now.

pattisdr:
Fixing tests...

app/routes/content.js
binoculars: It should already be a preprint by the time it hits this route, but we're doing very crazy things with the router on both the front-end and back-end to make it fit various requirements.

PR #203, opened 2016-11-30
Updates for PsyArXiv and smaller logos [PREP-265]

binoculars:
Purpose
---------
Respond to PsyArXiv requests for visual changes

Changes
--------
PsyArXiv wanted some changes and logos were blurry in the navbar, so I created smaller versions that look better

Side effects
-----------
None, unless transparent logo backgrounds don't work in some places

Ticket
-------
https://openscience.atlassian.net/browse/PREP-265

### Before:
screen shot 2016-11-30 at 14 07 12
### After:
screen shot 2016-11-30 at 14 08 13


brianjgeiger:
For tickets like this, before an after screenshots are super helpful. Could you add that really quickly?

PR #206, opened 2016-11-30
Add license-picker to the preprints add/edit. [PREP-235]

hmoco:
## Purpose
Add the ability to display, modify and add licenses to a preprint, taking in the provider's default and subset of available licenses

## Changes
Addition of license widget on the edit/add preprint flows, and a show license field on the Preprint view page

## Depends on:
https://github.com/CenterForOpenScience/ember-osf/pull/152

## Tickets
https://openscience.atlassian.net/browse/PREP-235

## QA Consideration
For right now, the discard changes and save indicator are behaving weirdly. Fixing it ASAP

### DIFF
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -599,7 +599,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
                     year: this.get('model.licenseRecord') ? this.get('model.licenseRecord').year : false || date.getUTCFullYear().toString(),
                     copyrightHolders: this.get('model.licenseRecord') ? this.get('model.licenseRecord').copyright_holders.join(', ') : false || ''
                 });
-            })
+            });
         },
         stripDOI() {
             // Replaces the inputted doi link with just the doi itself
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -305,7 +305,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
         if (this.get('model.licenseRecord') || this.get('model.license.content')) {
             changed = changed || (this.get('model.license.name') !== this.get('basicsLicense.licenseType.name'));
             changed = changed || (this.get('model.licenseRecord').year !== this.get('basicsLicense.year'));
-            changed = changed || (this.get('model.licenseRecord.copyright_holders.length') ? this.get('model.licenseRecord.copyright_holders').join(',') : this.get('basicsLicense.copyrightHolders') !== '');
+            changed = changed || ((this.get('model.licenseRecord.copyright_holders.length') ? this.get('model.licenseRecord.copyright_holders').join(',') : '') !== this.get('basicsLicense.copyrightHolders'));
         } else {
             changed = changed || ((this.get('availableLicenses').toArray().length ? this.get('availableLicenses').toArray()[0].get('name') : null) !== this.get('basicsLicense.licenseType.name'));
             let date = new Date();
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -201,8 +201,9 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
     uploadValid: Ember.computed.alias('nodeLocked'), // Once the node has been locked (happens in step one of upload section), users are free to navigate through form unrestricted
     abstractValid: Ember.computed.alias('validations.attrs.basicsAbstract.isValid'),
     doiValid: Ember.computed.alias('validations.attrs.basicsDOI.isValid'),
+    licenseValid: false,
     // Basics fields that are being validated are abstract and doi (title validated in upload section). If validation added for other fields, expand basicsValid definition.
-    basicsValid: Ember.computed.and('abstractValid', 'doiValid'),
+    basicsValid: Ember.computed.and('abstractValid', 'doiValid', 'licenseValid'),
     // Must have at least one contributor. Backend enforces admin and bibliographic rules. If this form section is ever invalid, something has gone horribly wrong.
     authorsValid: Ember.computed.bool('contributors.length'),
     // Must select at least one subject (looking at pending subjects)
@@ -373,8 +374,10 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
     }),
 
     actions: {
-        editLicense(license) {
+        editLicense(license, validates) {
+            debugger;
             this.set('basicsLicense', license);
+            this.set('licenseValid', validates);
         },
         next(currentPanelName) {
             // Open next panel
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -192,7 +192,11 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
             basicsTags: Ember.A(),
             basicsAbstract: null,
             basicsDOI: null,
-            subjectsList: Ember.A()
+            basicsLicense: null,
+            subjectsList: Ember.A(),
+            availableLicenses: Ember.A(),
+            applyLicense: false,
+            newNode: false
         }));
     },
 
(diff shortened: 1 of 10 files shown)


app/controllers/submit.js
```diff
@@ -556,22 +604,51 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
...
+            let currentNodeLicenseType = node.get('license');
+            let currentNodeLicenseRecord = node.get('nodeLicense');
+            let newCopyrightHolders = [''];
+            if (this.get('basicsLicense.copyrightHolders') && this.get('basicsLicense.copyrightHolders').length) {
+                newCopyrightHolders = this.get('basicsLicense.copyrightHolders').split(',');
+            }
 
             if (this.get('abstractChanged')) node.set('description', this.get('basicsAbstract'));
             if (this.get('tagsChanged')) node.set('tags', this.get('basicsTags'));
+            if (this.get('licenseChanged') && this.get('applyLicense')) {
```
pattisdr: `licenseChanged` is often true, when I have changed nothing. For example, I save my license, go to authors, and then return to basics and add a tag. I haven't touched the license. licenseChanged is still true, and then you end up unnecessarily putting the license stuff in the PATCH request.  

There was some discussion when I implemented this page over only sending in the relevant fields and worried about the small risk of collisions, hence the checking to make sure if a field has changed before setting it.

app/controllers/submit.js
```diff
@@ -279,9 +284,30 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
...
+            copyrightHolders: record && record.copyright_holders ? record.copyright_holders.join(',') : null,
+            licenseType: license || null
+        };
+    }),
+    licenseChanged: Ember.computed('model.license', 'model.licenseRecord', 'basicsLicense.year', 'basicsLicense.copyrightHolders', 'basicsLicense.licenseType', function() {
+        let record = this.get('model.licenseRecord');
+        let license = this.get('model.license') || null;
+        let changed = false;
+        if (record) {
+            changed = (record.year !== this.get('basicsLicense.year')) || (this.get('basicsLicense.copyrightHolders') !== record.copyright_holders.join(','));
```
pattisdr: When you load a preprint in Edit Mode, changed can be true here, which is incorrect.  This means licenseChanged is true, and thus basicsChanged is true.  It looks like `(this.get('basicsLicense.copyrightHolders') !== record.copyright_holders.join(','))` is comparing an empty array to an empty string and saying they are not equal.
screen shot 2016-12-01 at 5 21 30 am

pattisdr: Even after saving changes in Edit Mode, Discard Changes is still enabled.

app/mixins/setup-submit-controller.js
```diff
@@ -21,7 +22,11 @@ export default Ember.Mixin.create({
                 controller.set('providers', providers);
             }
         );
-
+        this.get('theme.provider').then(provider => {
+            provider.query('licensesAcceptable', {'page[size]': 20}).then(licenses => {
```
pattisdr: The spec says `The default license will be chosen by the service (OSFPreprints will start with CC0)`. My default license for the OSF is showing up as "Academic Free License (AFL) 3.0", the first alphabetically. But maybe the data isn't properly populated yet.

app/templates/submit.hbs
```diff
@@ -156,6 +166,18 @@
                                     </div>
                                 </div>
                                 <div class="row">
+                                    <div class="col-md-6">
+                                        <h4>{{if newNode 'Apply' 'Also apply'}} this license to my OSF Project.</h4>
+                                        {{#unless newNode}}
+                                            <p>Selecting a license for your preprint does not automatically apply the license to your OSF project.</p>
+                                        {{/unless}}
```
pattisdr: Also, spec says edit form has different language, don't know if this is still applicable: Apply this license to my OSF Project. Updating the license for your preprint does not automatically update the license on your OSF project.  The flag to know if you are on edit a preprint or adding a preprint (as both pages use this controller), is `editMode`.

app/controllers/submit.js
```diff
@@ -105,6 +105,10 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
     userNodes: Ember.A(),
     userNodesLoaded: false,
 
+    availableLicenses: Ember.A(),
+    applyLicense: false,
+    newNode: false,
```
pattisdr: These three fields need to be added to the `clearFields` method which restores submitForm defaults. For example, if you submit a preprint and then hit the back button, the clearFields method makes sure you don't have lingering data.

app/templates/submit.hbs
```diff
@@ -156,6 +166,18 @@
                                     </div>
                                 </div>
                                 <div class="row">
+                                    <div class="col-md-6">
+                                        <h4>{{if newNode 'Apply' 'Also apply'}} this license to my OSF Project.</h4>
```
pattisdr: Also, the spacing seems a little off here.
screen shot 2016-12-01 at 5 57 36 am

app/controllers/submit.js
```diff
@@ -350,6 +376,13 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
     }),
 
     actions: {
+        editLicense(license, validates) {
```
hmoco: Will use Ember.run.once on _setNodeLicense to ensure that the bouncing doesn't happen.

app/controllers/submit.js
```diff
@@ -350,6 +376,13 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
...
+            this.set('basicsLicense', license);
+            this.set('licenseValid', validates);
+        },
+        applyLicenseToggle(apply) {
+            this.set('applyLicense', apply);
```
hmoco: Because, a use case is that newNode is false, but applyLicense is true.

hmoco:
@pattisdr I'm adding +1 reactions to anything I need to go back to, to keep track of them! Thanks for the review so far!

app/controllers/submit.js
```diff
@@ -556,22 +604,51 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
             let currentAbstract = node.get('description');
             let currentTags = node.get('tags').slice(0);
             let currentDOI = model.get('doi');
+            let currentLicenseType = model.get('license');
+            let currentLicenseRecord = model.get('licenseRecord');
+            let currentNodeLicenseType = node.get('license');
+            let currentNodeLicenseRecord = node.get('nodeLicense');
+            let newCopyrightHolders = [''];
+            if (this.get('basicsLicense.copyrightHolders') && this.get('basicsLicense.copyrightHolders').length) {
+                newCopyrightHolders = this.get('basicsLicense.copyrightHolders').split(',');
```
hmoco: I believe that is intended behavior. It allows previous selections to be cached if the license ends up changing to one that requires those fields. Those are also not displayed when viewing the preprint's license if they are not required.

app/templates/submit.hbs
hmoco: Good point!

app/templates/submit.hbs
hmoco: Nevermind, I'm not sure if I was looking at something different or the spec has changed slightly, but you are right!

app/controllers/submit.js
pattisdr: But you only care about the state of applyLicense though, right? So that case, set applyLicense to true.

app/mixins/setup-submit-controller.js
hmoco: Yep, the preprint provider populate script doesn't contain license information yet

app/templates/content.hbs
```diff
@@ -88,6 +88,23 @@
...
+                        <h4 class="p-v-md f-w-md">{{t "global.license"}}</h4>
+                        <p class="p-v-md p-b-sm">
+                            <span style='cursor: pointer'>
+                                {{#if showLicenseText }}
+                                    {{fa-icon 'caret-down' click=(action 'toggleLicenseText')}}
+                                {{else}}
+                                    {{fa-icon 'caret-right' click=(action 'toggleLicenseText')}}
+                                {{/if}}
+                            </span>
+                            {{model.license.name}}
```
pattisdr: Can you move the license name back to before the caret? Sorry I wasn't clear, i just meant change the caret directions, but I think the caret still looks better after the license name.

pattisdr:
@brianjgeiger, @hmoco
I don't seem to be able to update the node license on the project overview page anymore (outside of this PR). I don't know if it's my environment, or if there were any recent changes to set_node_license or related.

brianjgeiger:
I'm seeing that, too. Checking Sentry.

brianjgeiger:
I'm having Casey look into that. Good catch.

pattisdr:
@hmoco  - the changes improve the page a lot, thanks for the detailed comments in the code as well.  The important thing is that changing the license on the preprint and the node, or one and not the other is working great.  

The `licenseChanged` behavior is improved, but not perfect. I am finding several instances where this is still broken, but shouldn't hold up the release.  Recommend that this go to staging, and we can work more on this for a follow-up release.

PR #214, opened 2016-12-02
[PREP-272]UI clean up, set default license.

hmoco:
## Purpose
Clean up licenses work found by QA.

## Changes
 - Wrapping on license text for preprints now matches the Node license view
 - Only shows the license field on content page of a preprint if that preprint has a license
 - If license is still set to be the default when creating a new preprint, this now applies the license to the request (as it should have been done)

## Ticket:
[PREP-272]


brianjgeiger:
Since you have matched code in EOSF, you'll need to update the pinned commit to `c69e10128f14edc82c35ebfe79f4a2a7b89d3ec2`

app/controllers/submit.js
```diff
@@ -792,6 +780,19 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
             model.set('isPublished', true);
             node.set('public', true);
 
+            if (this.get('applyLicense')) {
+                let newCopyrightHolders = [''];
+                if (this.get('basicsLicense.copyrightHolders') && this.get('basicsLicense.copyrightHolders').length) {
+                    newCopyrightHolders = this.get('basicsLicense.copyrightHolders').split(',');
+                }
+                if (node.get('nodeLicense.year') !== this.get('basicsLicense.year') || node.get('nodeLicense.copyrightHolders') !== newCopyrightHolders) {
```
pattisdr: Question about a particular case for this line.  Say my original node has a copyrightHolders Dawn.  The preprint created on this node has no copyrightHolders, nor did the user ever enter any because the license that they selected didn't require it. applyLicense is true.  It seems strange to me that we would then erase the copyrightHolders on the node in this instance.

pattisdr:
A few other comments:

1) In my opinion, we should still save the license to the node if `applyLicense` is true on the Edit Preprint form (same template, editMode).  The form currently behaves like several mini-forms.  Each section saves on `Save and Continue`.  I agree that in Add mode, we shouldn't save the license to the node until the end, but I think we should restore to former behavior for Edit Mode.  As is, a user could visit the preprint in edit mode, change the license, change abstract, tags, DOI, and check yes for applyLicense, hit 'Save and Continue', and then navigate away from the page. Every change that the user made would persist to the server, except for the license change on the node.  At some point, we've talked about moving everything on the form to save at the end, but we currently save by section, and should be consistent with that if we can. 

2) In basics, where year and copyright holders are required, lack of copyright holders will prevent the user from hitting 'Save and continue', but it might take the user awhile to figure that out.  We have red warning text on other inputs. Could we add a warning text if year missing?

3)  Licenses used to be broken into categories in dropdown. Are we not doing that anymore?


screen shot 2016-12-02 at 12 12 48 pm

pattisdr:
@hmoco pass done, I have a few questions, and am requesting that we still apply the license to the node in editMode in the basics section (open to discuss).

app/controllers/submit.js
hmoco: I think that would make sense. If applyLicense is true, then the license type will be applied to the node. So, either the user will be forced to have a copyright holders (replacing the node's) or copyright holders is not used for that license type. You can't have a license that shows copyright holders without filling that field.

hmoco:
1. This was a decision by @brianjgeiger and Nici on flowdock. I would agree with you however. @brianjgeiger, what should be the final behavior?
2. The copyright Holders should be showing a (required) next to the field name. Is it not? Or do you believe the warning should be more noticeable (red)
3. We are not doing categories on the preprints license picker. This is because all providers will be using a subset of licenses to be available.

brianjgeiger:
@hmoco @pattisdr It should be consistent with other node changing behavior. I _believe_ Nici was only asking for a change when adding a new item. But I think the short is: if it modifies an existing node, it should happen at the same time that other node-modifying things happen.

pattisdr:
Thanks @hmoco.

Did you see this comment? https://github.com/CenterForOpenScience/ember-preprints/pull/214#pullrequestreview-11200837

1) Yes, thanks, I just don't know if they considered editMode behavior.  I agree with everyone for Add mode.

2) Yes, I saw the (required)  I just think being more noticeable (red) would be helpful.

hmoco:
@brianjgeiger, ok! For clarification, we don't want the node being saved until the entire form is completed if we are not on edit mode, but if we are, then it should be saved with the rest of basics? Because the node gets saved in both places

brianjgeiger:
That sounds correct, yes.

saradbowman:
I think this should work the way everything else works in terms of saving.

pattisdr:
@hmoco was this change in response to @Sara's last comment? This undoes what @Nici suggested.   To reiterate, I think on Add Preprint, node license should save on 'Submit'.  In Edit Preprint, node License should save in 'Basics'.

hmoco:
@pattisdr, after talking to @brianjgeiger again, the suggested behavior is for the license to be saved when the basics form is submitted no matter what, as we are already updating other node fields there, on both add/edit modes.

pattisdr:
Got it, thanks.

brianjgeiger:
I had thought before that the rest of the node information was saved after, in the case of adding, or during, in the case of edit. But apparently that's not the case, so yeah, I disagreed with Nici and checked with Sara.

pattisdr:
OK, so just waiting on a response to https://github.com/CenterForOpenScience/ember-preprints/pull/214#pullrequestreview-11200837

app/controllers/submit.js
hmoco: This seems to be fine. If the copyright holders are changed, so is the license type. Then, either a different license that does not require copyright holders will be applied to the Node, so that would not be shown either way, or a license that does, in which case the form can't be submitted without putting somehting down for copyright Holders. There is no case where you are clearing node information that would be used.

pattisdr: Ok thanks

pattisdr:
@hmoco all good here.

app/controllers/submit.js
hmoco: oops not sure how I responded twice.

pattisdr:
Just saw Nici's comment on the JIRA ticket - waiting to see if there need to be changes for this.

brianjgeiger:
I think it's good. I'm going to merge this as is and if QA find anything else, we'll make a new PR.

PR #218, opened 2016-12-05
[PREP-278] Fix inconsistensies on copyrightHolders

hmoco:
## Purpose
Fix licenseChanged behavior to be more accurate.

## Changes
Fix inconsistencies on copyrightHolders

### DIFF
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -600,7 +600,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
                 this.set('basicsLicense', {
                     licenseType: license || this.get('availableLicenses').toArray()[0],
                     year: this.get('model.licenseRecord') ? this.get('model.licenseRecord').year : false || date.getUTCFullYear().toString(),
-                    copyrightHolders: this.get('model.licenseRecord') ? this.get('model.licenseRecord').copyright_holders.join(',') : false || ''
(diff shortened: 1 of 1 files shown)


app/controllers/submit.js
```diff
@@ -600,7 +600,7 @@ export default Ember.Controller.extend(BasicsValidations, NodeActionsMixin, Tagg
                 this.set('basicsLicense', {
                     licenseType: license || this.get('availableLicenses').toArray()[0],
                     year: this.get('model.licenseRecord') ? this.get('model.licenseRecord').year : false || date.getUTCFullYear().toString(),
-                    copyrightHolders: this.get('model.licenseRecord') ? this.get('model.licenseRecord').copyright_holders.join(', ') : false || ''
+                    copyrightHolders: this.get('model.licenseRecord') ? this.get('model.licenseRecord').copyright_holders.join(',') : false || ''
```
pattisdr: Can you explain why you need `false || ''` ? Why not just `''`?

pattisdr:
@hmoco - Looks good, just one question.

app/controllers/submit.js
hmoco: hah, good catch! Should be just `''`

pattisdr:
Looks good to me!  The combination of this and the ember-osf PR seems to do the trick. The save button / discard changes are behaving how I would expect. Good work.

PR #220, opened 2016-12-05
Case insensitive providers whitelist [PREP-279]

binoculars:
Makes whitelist for Providers case insensitive

# Ticket
https://openscience.atlassian.net/browse/PREP-279


pattisdr:
Changes look good to me.

PR #221, opened 2016-12-06
[PREP-280] Use osf regex if not an outside provider

hmoco:
Problem
---------
On branded preprint search, clicking on the title of a preprint that's got an article DOI directs the user to the DOI. It should send the user to the branded view of the preprint. 
An example is on PsyArXiv - search for "go/no go association task"
This particular example is ALSO broken because the DOI seems invalid- that's unrelated.

Changes
--------

Link to ticket
--------------
https://openscience.atlassian.net/browse/PREP-280

### DIFF
--- a/app/components/search-result.js
+++ b/app/components/search-result.js
@@ -29,29 +29,20 @@ export default Ember.Component.extend(Analytics, {
 
     osfID: Ember.computed('result', function() {
         let re = /osf.io\/(\w+)\/$/;
-        let re2 = /osf.io\/preprints\/(\w+)\/$/;
         // NOTE / TODO : This will have to be removed later. Currently the only "true" preprints are solely from the OSF
         // socarxiv and the like sometimes get picked up by as part of OSF, which is technically true. This will prevent
         // broken links to things that aren't really preprints.
-        if (this.get('result.providers.length') !== 1) {
-            return false;
-        }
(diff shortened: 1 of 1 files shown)


app/components/search-result.js
```diff
@@ -42,10 +42,9 @@ export default Ember.Component.extend(Analytics, {
     hyperlink: Ember.computed('result', function() {
         let re = null;
         for (let i = 0; i < this.get('result.providers.length'); i++)
```
binoculars: Assign the providers outside of the loop (using property look-ups within a loop is expensive)
```js
const resultProviders = this.get('result.providers'); // May need to call .toArray() because ember
```

Not sure why this needs to be a loop since the last iteration will win. There's no if statement with break inside the loop. I realize you didn't write that part, but this is a good time to revisit it.
```js
const re = this.providerUrlRegex[resultProviders[resultProviders.length - 1] || this.providerUrlRegex.OSF;
```
⬆️  Should be the same result for `re`. If it does need to be in a loop for a valid reason, please add a comment explaining that reason so we can have it documented.

hmoco:
@binoculars, please take a look at the changes I made. Your comment made sense, and the last one in the iteration would win even if it returned a falsy value. So I changed so that if earlier iterations did return a value, that would not get blanked by a falsy one in a later iteration, with OSF being used if none of the providers return anything.

PR #222, opened 2016-12-06
Feature/Project Dropdown Filter both Projects and Components [PREP-254]

pattisdr:
# Ticket 
https://openscience.atlassian.net/browse/PREP-254
❗️ Depends on https://github.com/CenterForOpenScience/ember-osf/pull/148

# Purpose

Searching for a project title to add a preprint to should also return all components under that project in the results.

# Changes

Pass a custom matcher into the power-select component that looks through the root title, node title, and parent title for a match, instead of just the node title.

### DIFF
--- a/app/components/preprint-form-project-select.js
+++ b/app/components/preprint-form-project-select.js
@@ -55,4 +55,13 @@ export default Ember.Component.extend({
      * @property {boolean} fileSelect
      */
     fileSelect: false,
+
+    titleMatcher(node, term) {
+        // Passed into power-select component for customized searching.
+        const nodeTitle = (node.get('title') || '').toLowerCase();
+        const rootTitle = (node.get('root.title') || '').toLowerCase();
+        const parentTitle = (node.get('parent.title') || '').toLowerCase();
+        const lowerTerm = term.toLowerCase();
+        return nodeTitle.includes(lowerTerm) || rootTitle.includes(lowerTerm) || parentTitle.includes(lowerTerm) ? 1 : -1;
+    }
 });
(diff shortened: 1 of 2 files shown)


app/components/preprint-form-project-select.js
```diff
@@ -55,4 +55,13 @@ export default Ember.Component.extend({
      * @property {boolean} fileSelect
      */
     fileSelect: false,
+
+    titleMatcher(node, term) {
+        // Passed into power-select component for customized searching.
+        const nodeTitle = (node.get('title') || '').toLowerCase();
```
binoculars: The default matcher for ember-power-select includes using [stripDiacritics](/cibernox/ember-power-select/blob/master/addon/utils/group-utils.js#L961-L972), which converts special characters. Since we allow those characters in project/component titles, we ought to follow that convention here as well. Here's some (untested) reference code:

```js
import {stripDiacritics} from 'ember-power-select/utils/group-utils';
```

…

```js
const sanitizedTerm = stripDiacritics(term).toLowerCase();

const fields = [
    'title',
    'root.title',
    'parent.title'
];

for (const field of fields) {
    const value = node.get(field) || '';

    if (!value)
        continue;

    const sanitizedValue = stripDiacritics(value).toLowerCase();

    if (sanitizedValue.includes(sanitizedTerm))
        return 1;
}

return -1;
```

brianjgeiger:
For client side filtering, stripping diacriticals on both the filter term and the source fields should be fine.

pattisdr:
Thanks for pointing this out @binoculars - I completely missed it.

PR #223, opened 2016-12-07
Add redirect for urls with uppercase chars, remove duplicate metrics [PREP-286]

binoculars:
## Purpose
Make provider names case insensitive (redirects to correct location if it is the correct provider name, but wrong case)

Also, get rid of duplicate trackEvent for sharing link buttons on the content page (Google Analytics).

## Ticket
https://openscience.atlassian.net/browse/PREP-286

### DIFF
--- a/app/controllers/content.js
+++ b/app/controllers/content.js
@@ -153,12 +153,6 @@ export default Ember.Controller.extend(Analytics, {
         shareLink(href, network, action, label) {
             const metrics = Ember.get(this, 'metrics');
 
-            metrics.trackEvent({
-                category: network,
-                action,
-                label: window.location.href
-            });
-
             if (network.includes('email')) {
                 metrics.trackEvent({
                     category: 'link',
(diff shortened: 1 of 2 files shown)


app/controllers/content.js
```diff
@@ -153,12 +153,6 @@ export default Ember.Controller.extend(Analytics, {
         shareLink(href, network, action, label) {
             const metrics = Ember.get(this, 'metrics');
 
-            metrics.trackEvent({
```
brianjgeiger: Is this related to this PR?

brianjgeiger: Why yes it is, right there in the description.

app/routes/provider.js
```diff
@@ -10,8 +10,17 @@ export default Ember.Route.extend({
 
     beforeModel(transition) {
         const {slug} = transition.params.provider;
+        const slugLower = (slug || '').toLowerCase();
+
+        if (this.get('providerIds').includes(slugLower)) {
+            if (slugLower !== slug) {
+                const {pathname} = window.location;
+                window.location.pathname = pathname.replace(
+                    new RexExp(`^/preprints/${slug}`),
```
sloria: Typo here: `RegExp`

PR #225, opened 2016-12-07
Hotfix/Contributor List Order in Search Results [PREP-287]

pattisdr:
# Ticket
https://openscience.atlassian.net/browse/PREP-287

# Purpose
The contributor list on the Discover Preprints page is in reverse order of the contributor list on the Preprint detail page.  The preprint detail page is correct.  The cause is we were sorting contributors from SHARE to make it look like APIv2 preprints data, we were sorting descending instead of ascending.

# Changes
Reverses the contributor array returned from SHARE.

### DIFF
--- a/app/components/search-result.js
+++ b/app/components/search-result.js
@@ -27,6 +27,14 @@ export default Ember.Component.extend(Analytics, {
         return result.description.slice();
     }),
 
+    contributors: Ember.computed('result', function() {
+        let result = this.get('result');
+        if (result.contributors) {
+            return result.contributors.reverse();
+        }
+        return;
+    }),
+
     osfID: Ember.computed('result', function() {
         let re = /osf.io\/(\w+)\/$/;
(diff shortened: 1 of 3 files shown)


app/components/search-result.js
```diff
@@ -27,6 +27,14 @@ export default Ember.Component.extend(Analytics, {
         return result.description.slice();
     }),
 
+    contributors: Ember.computed('result', function() {
+        let result = this.get('result');
+        if (result.contributors) {
+            return result.contributors.reverse();
```
brianjgeiger: So did switching a and b on https://github.com/CenterForOpenScience/ember-preprints/blob/develop/app/controllers/discover.js#L179 not work?

pattisdr: Switching a and b does work, I was just being dumb :)

pattisdr: I forgot we were doing this sorting preprints-side and thought it was coming from SHARE already in this order.

PR #228, opened 2016-12-08
Feature/Private Preprint Should Redirect to Page Not Found[PREP-288]

pattisdr:
## Purpose

Currently if a user tries to navigate to a preprint that has been made private and the user is not authenticated/not authorized, the page will get stuck in an infinite loop.

## Changes

Uses the specific error detail messages returned by the APIv2 to determine if 401 or 403 because we can't access the raw json in Ember and do not see the status code (looking at an Ember Error that has some of the info from the API v2 error but not all).  If error message is `You do not have permission to perform this action.` or `Authentication credentials were not provided.` on the preprints content page, redirects both to Page Not Found.

Otherwise, runs through existing redirect logic (sometimes we want to redirect back to OSF side and show the Page Not Found there, for example.)

## Ticket

https://openscience.atlassian.net/browse/PREP-288

### DIFF
--- a/app/routes/content.js
+++ b/app/routes/content.js
@@ -143,8 +143,11 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
     },
     actions: {
         error(error, transition) {
-            if (error && error.errors && Ember.isArray(error.errors) && error.errors[0].detail === 'The requested node is no longer available.') {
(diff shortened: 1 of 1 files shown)


pattisdr:
Closing in favor of https://github.com/CenterForOpenScience/ember-preprints/pull/247

PR #232, opened 2016-12-08
[EOSF-397] Submit pieces of the submit/edit form by pressing enter on input fields.

hmoco:
## Purpose
Allows moving onto the next piece of the submit/edit form by pressing enter on fields where it makes sense to do so.

## Changes
Validated-input, if not a text-area, allows for sending a pressSubmit event when pressing enter.

## Ticket
https://openscience.atlassian.net/browse/PREP-158

## Testing Considerations:
Fields to attempt pressing Enter when the form is valid and the cursor is on the edit field:
 - Making a new preprint -> upload file -> put title -> press enter: Save and continue to subjects
 - Connect to an existing project -> upload file -> make new component -> insert title -> press enter: Save and continue to subjects
 - Connect to an existing project -> upload file -> use existing project -> edit title -> press enter: Save and continue to subjects
 - Connect to an existing project -> choose existing file -> make new component -> insert title -> press enter: Save and continue to subjects
 - Connect to an existing project -> choose existing file -> use existing project -> edit title -> press enter: Save and continue to subjects
 -  On basics section, if and only if all requirements are completed (valid DOI, abstract and license fields), pressing enter on Year (of license widget if available), copyright Holders (of license widget if available), on DOI if empty or a valid doi.


abought:
The original JIRA ticket also required enter-on-submit behavior for the various modals throughout the site. Here is one modal as an example (but also look through the code for wherever you see `bs-modal` used):

screen shot 2017-01-26 at 4 15 56 pm

If this is a scope change agreed on with product, please update the ticket and then return to me. Thanks!

hmoco:
@abought a new ticket has been created for modals (https://openscience.atlassian.net/browse/EOSF-478). Requesting that this moves forward!

app/controllers/submit.js
```diff
@@ -688,6 +691,9 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
                     label: `Preprints - ${this.get('editMode') ? 'Edit' : 'Submit'} - Save and Continue Basics Section`
                 });
             // Saves the description/tags on the node and the DOI on the preprint, then advances to next panel
+            if (!this.get('basicsValid')) {
```
abought: Confirmed that this is the only section where a user will be hitting enter in a single-line text field, so none of the other validators need to be checked. 

If we add fields to these forms in the future that may change, in which case "enter to submit" would need to be added to those fields later.

app/templates/components/preprint-form-project-select.hbs
```diff
@@ -92,7 +92,7 @@
                                             {{/if}}
                                         {{/if}}
                                         <form onchange={{action 'track' 'input' 'onchange' 'Preprints - Submit - Add title, Connect preprint to existing project'}}>
-                                            {{preprint-title-editor nodeTitle=nodeTitle titleValid=titleValid}}
+                                            {{preprint-title-editor nodeTitle=nodeTitle titleValid=titleValid pressSubmit=(action (if (eq convertOrCopy 'convert') existingNodeExistingFile createComponentCopyFile) name)}}
```
abought: This seems like a bit of a subtle edge case- in your testing notes, you might want to call this conditional scenario out to QA specifically so they know to check it.

hmoco:
@abought update PR with testing considerations.

app/templates/components/preprint-form-project-select.hbs
abought: I see... this copies logic from line 99. I'd say let this pass, but note the copy-pasted logic it as a potential maintainability issue for the future.

PR #233, opened 2016-12-08
Feature/Preprint Search Results Date [PREP-282]

pattisdr:
## Purpose

1) You have to expand the caret on the preprints search result to view the date
2) Preprints search is displaying the date the preprint was added to the branded service, not added to OSF

## Changes

1) Move Added on so you don't have to expand caret to view
2) Change search result to look at `date` instead of `date_created`.  `date` is the most relevant field, either` (data['date_published'] or data['date_updated'] or data['date_created'])`
3) Have preprint content page show datePublished instead of dateCreated

screen shot 2016-12-08 at 4 36 00 pm

## Side effects
On the content page, changing date_created to date_published should mostly be the same.  A preprint is created when they start the Add Preprint process - as they work through the form, they are really editing a draft.  The datePublished field is different than the dateCreated field.  However, because we don't allow them to resume the draft, these are always going to be close together.  But datePublished seems more accurate - the time in which the preprint is available for others to see. dateCreated is the preprint draft.  

## Ticket

https://openscience.atlassian.net/browse/PREP-282


app/routes/content.js
```diff
@@ -104,7 +104,8 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
                 ['og:description', node.get('description')],
                 ['og:site_name', this.get('theme.provider.name')],
                 ['og:type', 'article'],
-                ['article:published_time', new Date(preprint.get('dateCreated') || null).toISOString()]
+                ['article:published_time', new Date(preprint.get('datePublished')).toISOString()],
```
abought: Can you clarify why this extra conversion step is necessary? (lines 107-108)

Within the preprints front-end, is there ever a situation where `datePublished` and/or `dateModified` would be null? If not, then any DS.attr of type `date` will [already be a date object](https://github.com/emberjs/data/blob/master/addon/-private/transforms/date.js#L27). (1)


If I'm reading this right, that means:
- The only reason to convert the field to a `Date` would be if Date was null- eg if you weren't guaranteed that the field value had a `toUTCString` method already.
- If you create a Date by passing in null, then the date displayed might not be optimal for the use case. (2)


[1] Example command line exploration by inserting a debugger statement into content.js...
```
datepub = preprint.get('datePublished')
Wed Jan 18 2017 18:17:56 GMT-0500 (EST)
datepub instanceof Date
true
```

[2] Sample date creation from null
```
new Date(null).toISOString()
"1970-01-01T00:00:00.000Z"
```

app/routes/content.js
```diff
@@ -104,7 +104,8 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
...
+                ['article:modified_time', new Date(preprint.get('dateModified')).toISOString()]
```
abought: Confirmed that meta tags are being added to the page with correct date!

app/routes/content.js
pattisdr: When you say conversion step, you're talking about .toISOString()? It seems like SHARE results don't always have dates there.

abought: > When you say conversion step, you're talking about .toISOString()? It seems like SHARE results don't always have dates there.

This is defined in the `afterModel` hook, where `preprint` is the result of an API call- it doesn't seem like this specific usage of field data will ever be a SHARE object...? (see line 31) I might be missing something, so wanted to ask.

By conversion step, I mean the act of converting this field value into a date. By default, it seems that `DS.DateTransform` already converts things into a date for us.
`new Date(preprint.get('dateModified'))`

pattisdr: You're right, I was getting it confused with the search-results, also addressed in this PR, which are using SHARE results.  This is a preprint from the API, so I think it would always have a date.

pattisdr: So would you like me to remove the Date(...) @abought?

abought: Yes- or identify why it's there....?

Sometimes complex code points to an edge case, so I try to ask before just calling for its removal.... but in this case, the most likely edge case would cause the published date to show "1970". 

If we can simplify the code and get the same result, prefer the simpler version.

pattisdr: Okay, I didn't write the Date(...) originally, but I can't think of a reason myself why we need it.

brianjgeiger: @binoculars ^^^

binoculars: If you don't add, `|| null`, we and dateCreated is not defined, the page throws an error and refuses to render.

pattisdr: So I need to restore || null, and the Date(...) needs to stay because we don't want to do `.toISOString` of null.

abought: Interesting. Under what circumstances can a preprint be used with this widget but not have a creation date?

If you create a date object without having a date, it seems this would default to Jan 1 1970, which is probably not an accurate choice of date. Is that the desired behavior?

binoculars: We were seeing it in production. This only applies to the meta tags. If dateCreated isn't there, it's not really Preprint's job to figure out what the desired behavior is. We the API contract dictates that it will always be there. Having a bad date for a scraper is better than throwing an error for the entire page. I was planning on verifying that #239 works once it's up on staging. We'll probably have merge conflicts.

abought: An API bug? Interesting....

Since this deals with a known edge case, it's ok to leave in and approve for move forward as written.

As long as the API behavior is tracked and ticketed, we're good here. Thanks for verifying. A fix is outside the scope of this PR and we can wait until the dust settles.

pattisdr: Okay, restored || null and Date conversion.

PR #235, opened 2016-12-12
Feature/Helper tests, Submit Controller Tests, & Documentation [PREP-293]

pattisdr:
## Purpose

Add some tests and documentation for preprints!

## Changes
- [x] [Audit](https://docs.google.com/spreadsheets/d/1YwJ70e3DFzgdT3t96UVTIwhX5PqBIzHY6EejdTuxjE8/edit#gid=0) what tests already exist for completeness (this is mainly just a checklist of our routes, controllers, helpers, etc.)
- [x] Complete testing of helpers
- [x] Add tests for submit controller (actions and computed properties)
- [x] TODO point to latest ember-osf commit once related PR merged https://github.com/CenterForOpenScience/ember-osf/pull/165
- [x] Add some preprints documentation - add yuidoc.json, and some yuidoc syntax so docs building properly (basing off @abought's work in ember-osf).

### Current issues:  
1. Some tests skipped because they are testing controller actions that require the promise be resolved before asserting values.  Calling controller actions is done via `controller.send` so promise is not returned. Don't know a way to wait until promise has resolved?
2. Some tests skipped failing with message, "You can only unload a record which is not inFlight." action testing on submit controller

## Notes for Reviewers
Just attempting to start on tests here, lots of tests commented out and lots more tests to write.

## Ticket

https://openscience.atlassian.net/browse/PREP-293


app/helpers/safe-markup.js
```diff
@@ -1,6 +1,12 @@
 import Ember from 'ember';
 
-export function safeMarkup(params/*, hash*/) {
+/**
+  * safeMarkup helper. Returns a string that will not be HTML escaped by Handlebars.
+  *
+  * @method safeMarkup
+  * @param {string} string
+  * @return {Object} Returns Handlebars.SafeString
+  */export function safeMarkup(params/*, hash*/) {
```
abought: Thanks for adding a docstring! Can you confirm that this actually shows up in the generated docs when you run `yarn run docs`?

(you don't need to commit the changes, but please verify they show up)

1. Can the export statement please be on a separate line?
2. For convention sake, I've been trying to write the helper name as it would appear when used (eg `safe-markup`). Not sure if this causes problems for yuidoc- thoughts?

app/helpers/safe-markup.js
```diff
@@ -1,6 +1,12 @@
 import Ember from 'ember';
 
-export function safeMarkup(params/*, hash*/) {
+/**
+  * safeMarkup helper. Returns a string that will not be HTML escaped by Handlebars.
```
abought: A good start! Can you please add a sentence that discourages the use of this helper for user-entered text?

This was created for a very specific niche need (branding), and for security reasons should almost never be used in general development. This came up recently or I wouldn't impose on you.

tests/unit/helpers/number-format-test.js
```diff
@@ -1,10 +1,15 @@
...
 
-module('Unit | Helper | number format');
+// module('Unit | Helper | number format');
 
-// Replace this with your real tests.
-test('it works', function(assert) {
-  let result = numberFormat([42]);
-  assert.ok(result);
-});
+// Works in Chrome but doesn't work in terminal
```
abought: Interesting. There are two alternative paths here:

1. If a test doesn't work in the terminal, that's often because phantomJS is based on older and more outdated browser code than Chrome. Sometimes this is because code relies on es6 features, and we can get around it via babel, our transpiler.

2. Rather than commenting out a test, we should explicitly mark it as skipped. That helps us to not lose track of the test later.
https://github.com/emberjs/ember-qunit/issues/144#issuecomment-75179707

app/helpers/safe-markup.js
pattisdr: hmm, trying to get `yarn run docs` to work - seeing `error Command "docs" not found.`

abought: Ah. Apparently this command is defined in `package.json#scripts` for ember-osf, but not for preprints.

pattisdr: oh, right, thanks.

pattisdr:
Ok, I have skipped tests that I don't have time to address at this time, rather than commenting them out.  Added boilerplate for documentation and added some docstrings to helpers, components, routes, and controllers.

app/components/file-uploader.js
```diff
@@ -12,10 +16,43 @@ import Analytics from '../mixins/analytics';
  *  either become the newly created project or component, or the existing node.  After file is uploaded to the designated 'node',
  *  'osfFile' is set to the uploadedFile.
  *
- *  NOTE: file-uploader is used in two places in this application - on the submit page and inside the preprint-form-project-select component.
+ *  NOTE: file-uploader is used in two places in the preprints application - on the submit page and inside the preprint-form-project-select component.
  *  If new properties need to be passed to this component, be sure to update in both places.
  *
- * ```
+ * ```handlebars
+ * {{file-uploader
```
abought: I do not envy you the experience of writing this usage example. Wow.

app/helpers/filter-replace.js
```diff
@@ -7,6 +7,19 @@ var filters = {
...
+/**
+ * @module ember-preprints
+ * @submodule helpers
+ */
+
+/**
+ * filterReplace helper. Replaces long provider names without messing with search filter logic
+ *
+ * @class filterReplace
+ * @param {String} filter Filter
```
abought: Can a `class` have `params`? Or does this need to be a `method`? (Applies to various places in these docs)

It appears that ember's builtin template helpers (like hash) use `@for`, which could be an option if we don't want to declare things as both classes *and* methods all over the place:
http://yui.github.io/yuidoc/syntax/index.html#for
https://github.com/emberjs/ember.js/blob/v2.11.0/packages/ember-glimmer/lib/helpers/hash.js#L31

app/controllers/application.js
```diff
@@ -1,6 +1,15 @@
...
 import OSFAgnosticAuthControllerMixin from 'ember-osf/mixins/osf-agnostic-auth-controller';
 
+/**
+ * @module ember-preprints
+ * @submodule controllers
+ */
+
+/**
+ * @class Application Controller
+ * @extends Ember-OSF.OSFAgnosticAuthControllerMixin
```
abought: Technically, this extends Ember.Controller. For mixin, a better option might be `@uses`.

Here is a usage example from ember showing extends vs uses in action:
https://github.com/emberjs/ember.js/blob/v2.11.0/packages/ember-runtime/lib/system/array_proxy.js#L74

app/controllers/content.js
```diff
@@ -40,6 +41,14 @@ function queryStringify(queryParams) {
     return query.join('&');
 }
 
+/**
+ * @module ember-preprints
+ * @submodule controllers
```
abought: I *think* declaring the module once at the top of the file is sufficient and might be preferred?

app/controllers/content.js
```diff
@@ -7,11 +7,12 @@ import permissions from 'ember-osf/const/permissions';
 /**
  * Takes an object with query parameter name as the key and value, or [value, maxLength] as the values.
  *
+ * @method queryStringify
  * @param queryParams {!object}
  * @param queryParams.key {!array|!string}
  * @param queryParams.key[0] {!string}
  * @param queryParams.key[1] {int}
- * @returns {string}
+ * @return {string}
```
abought: Not requesting a change in scope of this PR, but as a general note this docstring isn't quite clear- it describes the inputs but not the intended output or behavior. In particular, how this differs from `$.param`. (since Ember bundles jquery by default)

app/helpers/if-filter.js
```diff
@@ -1,20 +1,29 @@
...
+ * @submodule helpers
  */
+
+/**
+  * ifFilter helper. To be used within an {#if} block, to only display an item if it matches a filter.
+  *
+  * @class ifFilter
+  * @param {String} element
+  * @param {Object} filter
+  * @param {String} list intersection ??
```
abought: We can, uh, improve this one later...? ;)

app/helpers/number-format.js
```diff
@@ -1,5 +1,16 @@
...
+/**
+ * @module ember-preprints
+ * @submodule helpers
+ */
 
+/**
+  * numberFormat helper. Transforms 3500 into 3,500 for example, if in US English locale.
+  *
+  * @class numberFormat.
+  * @param {Integer} element
```
abought: In the utterly most pedantic sense, Javascript doesn't have integers; just numbers.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number
http://www.2ality.com/2014/05/is-integer.html

This is mostly only relevant in that YUIDoc can be configured to automatically link to built-in types it knows about. Most people should know what an integer is without looking it up, though.

abought:
Oh, spoke to soon- I guess this will need developer attention... This PR has developed a merge conflict. Busy afternoon!

pattisdr:
Addressing merge conflict now...

app/helpers/filter-replace.js
pattisdr: I am not sure. I had originally used `method` here, but it wasn't showing up in the docs.

abought: Yeah, in the past I've had to declare both `class` and `method`. Otherwise, the method isn't part of anything. There may be a more elegant way.

YUIDoc writes out a file called `docs/data.json`, and the `warnings` section of that file can sometimes (but not always) help identify doc build errors.

PR #236, opened 2016-12-13
[PREP-149] Clear url params on link-to transition

hmoco:
## Purpose
When leaving the page for any reason, the filters should be cleaned up.

## Changes
Makes it so link-tos clear the urlparams which dictate filters by setting them on willTransition.

## Ticket

https://openscience.atlassian.net/browse/PREP-149


app/routes/discover.js
```diff
@@ -14,4 +14,13 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, {
             .findAll('preprint-provider', { reload: true })
             .then(result => result.filter(item => item.id !== 'osf'));
     },
+    actions: {
+        willTransition() {
+            //Don't want to send clearFIlters action as that will trigger
+            //behavior specifc to the clearFIlters button.
+            let controller = this.controllerFor('discover');
```
binoculars: L22-L23 don't actually clear out the filters. We need to use the [functionality in the controller](/CenterForOpenScience/ember-preprints/blob/develop/app/controllers/discover.js#L327-L330) instead. This will actually reset the filters and trigger the query params to change. I suggest moving those lines to a function on the controller (maybe called `_clearFilters`) and then calling that inside the `clearFilters` action.

We should then be able to use `this.controllerFor('discover')._clearFilters()` here in the route's `willTransition` action. The reason I'd like to separate them out in the controller is so that the metrics trackEvent doesn't get called incorrectly (i.e. it should only be called when the correct user action is fired.

I also noticed that the `q` query parameter doesn't get reset either. That should be addressed in this PR as well.

binoculars:
It's working as expected now. I pulled in the latest develop, and ran the tests in docker. They're all passing. Good work! @hmoco, if you can pull develop into this branch, the Travis check should pass as well.

PR #238, opened 2016-12-16
Fix for GUID URL infinite redirects [PREP-300]

binoculars:
## Ticket
https://openscience.atlassian.net/browse/PREP-300

## Notes for reviewers
Create a new preprint, logout, then try to access that url. It should work correctly now.

PR #237, opened 2016-12-13
[PREP-295] Add google scholar and more meta tags

hmoco:
## Purpose

 Add google scholar and more meta tags

## Ticket

https://openscience.atlassian.net/browse/PREP-295

PR #239, opened 2016-12-17
Feature/fix url redirect merge [PREP-300] [EOSF-362]

binoculars:
Companion to #238 (this targets develop)

## Purpose
Fixes conflicts with #237 and #238 and fixes ordering of meta tags

## Changes
Updates meta tags for contributors, fixes error handling in redirects

## Side effects
N/A

## Notes for Reviewers
Look at meta tags in `<head>` on a preprint page e.g.:
screen shot 2016-12-17 at 15 46 25

## Notes for QA / testing
We'll need to ensure that the tags are scraped properly from Facebook, LinkedIn, and Google Scholar once this is up on staging

## Ticket
https://openscience.atlassian.net/browse/PREP-300
https://openscience.atlassian.net/browse/EOSF-362


abought:
The associated JIRA ticket is closed- is there something available to track this work?

binoculars:
This would be a better candidate for [acceptance testing](https://guides.emberjs.com/v2.11.0/testing/acceptance/). If we decide to add new tags or the requirements change in the future, it would be difficult to anticipate. We could have the list of current tags in the correct order and then fail the test if any new ones were added without updating the test. That would be reasonable.

brianjgeiger:
That seems like a good way to do that.

PR #243, opened 2017-01-05
[Hotfix] Permission language for the branded preprint partners for using the -rXiv extension [EOSF-403]

sheriefvt:
## Purpose

The purpose is to improve preprints UI by adding the permission language to the branded preprint sites. 

The branded preprint partners are using the -rXiv extension with permission from Cornell and they need a language on their pages stating such.
engrXiv: "arXiv is a trademark of Cornell University, used under license. This license should not be understood to indicate endorsement of content on engrXiv by Cornell University or arXiv."
SocArXiv: "arXiv is a trademark of Cornell University, used under license."
PsyArXiv: "arXiv is a trademark of Cornell University, used under license. This license should not be understood to indicate endorsement of content on PsyArXiv by Cornell University or arXiv."

The language is showing now below the COS copyright on the UI. See the attached figures.
screen shot 2017-01-05 at 11 01 19 am

screen shot 2017-01-05 at 11 04 27 am

screen shot 2017-01-05 at 11 05 28 am

## Changes
- Create a new component permission-language.
- Create a helper function “eq” to check provider name.
- Update the application scss.
- Add the new component in the proper place within the larger component preprint-branded-footprint.

## Side effects
None expected.

## Ticket

https://openscience.atlassian.net/browse/EOSF-403


abought:
The screenshot shows copyright 2011-2016. Should that be updated?

app/helpers/eq.js
```diff
@@ -0,0 +1,4 @@
+import Ember from 'ember';
+
+const eq = (params) => params[0] === params[1];
```
brianjgeiger: This isn't necessary. Looking through the package.json, we already have ember-truth-helpers installed https://github.com/CenterForOpenScience/ember-preprints/blob/develop/package.json#L68 so you can just use that instead of defining your own.

sheriefvt: Thanks..will work on this

brianjgeiger:
Also, be sure to run tests locally before pushing. Your new test was failing.

app/templates/components/permission-language.hbs
```diff
@@ -0,0 +1,18 @@
+{{yield}}
+{{#if (eq model.name "engrXiv")}}
+    <div class="text-center permission-footer-language ">
+        <span>arXiv is a trademark of Cornell University, used under license. This license should not be understood to indicate endorsement of content on engrXiv by Cornell University or arXiv.</span>
```
brianjgeiger: We are using i18n to enable us to do localization in the future, which means that we keep language in a separate file. This will let us translate into different languages easily. I don't know the exact method for dealing with this, but you'll want to put these strings into https://github.com/CenterForOpenScience/ember-preprints/blob/develop/app/locales/en/translations.js and pull from there.

tests/integration/components/permission-language-test.js
```diff
@@ -0,0 +1,25 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('permission-language', 'Integration | Component | permission language', {
+  integration: true
+});
+
+test('it renders', function(assert) {
```
brianjgeiger: Would it be possible to add tests to ensure that, if we're using a particular model, we're getting the proper text? We don't have a lot of tests in ember yet, and we're not quite all-in with knowing the best way to test, but that seems like a worthwhile set of tests. It might also be worth verifying, if possible, that for every model we have copyright text to go with. Maybe ask Nici if that's an assumption we should make.

app/templates/components/permission-language.hbs
sheriefvt: I did this change. Thanks for your feedback

tests/integration/components/permission-language-test.js
sheriefvt: I just added tests to ensure the component is returning the proper language for each provider.

brianjgeiger: Hmmm. I think those tests are probably not what we need. Go ahead and take them out, and then I'll approve the PR and get you working on something else.

app/locales/en/translations.js
```diff
@@ -248,6 +248,11 @@ export default {
             support: `Support`,
             contact: `Contact`
         },
+        'permission-language':{
```
binoculars: This needs to be collapsed. We don't want the translations to be duplicated. We can add a configuration flag in config/environment.js to conditionally include the second sentence.

app/templates/components/permission-language.hbs
```diff
@@ -0,0 +1,18 @@
+{{yield}}
```
binoculars: There's no need to create a component for this. The markup should just go in preprint-footer-branded.hbs.

app/locales/en/translations.js
brianjgeiger: We don't know what the other providers are going to want as far as trademark info. They're not all going to be arxiv based, but it is likely they're going to want some sort of IP statement if they are branding somehow. We can roll it into some more arbitrary text location if Product approves that, but it's otherwise having this be flexible is better.

binoculars: I agree with that, but the translation file should still never contain duplicate sentences. The conditional logic should be in the template. There is a way to pass the name of the provider from the template to the translation.

app/locales/en/translations.js
```diff
@@ -1,5 +1,6 @@
 const preprints = `Preprints`;
 const brand = `OSF ${preprints}`;
+const permission_sentence1 = 'arXiv is a trademark of Cornell University, used under license.';
```
brianjgeiger: Maybe `arxiv_trademark_license` for this and `arxiv_non_endorsement` for the other.

app/locales/en/translations.js
sheriefvt: thanks for your feedback. Much better now.

PR #247, opened 2017-01-11
[hotfix] [EOSF-179] Handle API errors

binoculars:
## Purpose
401, 403, and 404 were causing the page to break

## Changes
Handles API errors in content route and transitions to the appropriate page

## Side effects
N/A

## Ticket
https://openscience.atlassian.net/browse/EOSF-179

### DIFF
--- a/app/routes/content.js
+++ b/app/routes/content.js
@@ -6,6 +6,15 @@ import config from 'ember-get-config';
 import loadAll from 'ember-osf/utils/load-relationship';
 import permissions from 'ember-osf/const/permissions';
 
+// Error handling for API
+const handlers = new Map([
+    // format: ['Message detail', 'page']
+    ['Authentication credentials were not provided.', 'forbidden'], // 401
+    ['You do not have permission to perform this action.', 'forbidden'], // 403
+    ['Not found.', 'page-not-found'], // 404
+    ['The requested node is no longer available.', 'resource-deleted'] // 410
+]);
+
 export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitControllerMixin, {
     theme: Ember.inject.service(),
     headTagsService: Ember.inject.service('head-tags'),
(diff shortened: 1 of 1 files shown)


app/routes/content.js
```diff
@@ -6,6 +6,15 @@ import config from 'ember-get-config';
 import loadAll from 'ember-osf/utils/load-relationship';
 import permissions from 'ember-osf/const/permissions';
 
+// Error handling for API
+const handlers = new Map([
+    // format: ['Message detail', 'page']
+    ['Authentication credentials were not provided.', 'forbidden'], // 401
+    ['You do not have permission to perform this action.', 'forbidden'], // 403
```
pattisdr: Suggest having both 401 and 403 errors redirect to Not Found instead of Forbidden. 

This issue was previously raised in this ticket: https://openscience.atlassian.net/browse/EOSF-179.
That ticket said, "Expected behavior for a preprint project made private is that the preprint link will show a "Page not found". 

Forbidden is more accurate, but do we want to tell the user this?

binoculars: Really, a 401 should have a separate unauthorized page and tell the user that they need to log in in order to get either the correct page, or the forbidden page. With 403s going to not found, we're just obfuscating the fact that the project was made private. The API will still return a 403 code. It should be fixed in the API to be correct. I can change it quickly, but just want to make sure (since it's a hotfix) that we get it right.

pattisdr: You can just leave it as forbidden to match the API, then. I was mainly bringing this up since the original ticket said to make it Not Found, but I don't have strong feelings either way.

brianjgeiger: Definitely long-term 401 should bounce to CAS login, which has some implied functionality for this helper. But for now, the above is okay. For the 403, since the original ticket did have Not Found and it was approved in a Product Demo, either get Product to approve this as is (pointing out the original ticket) or go with the original version. I'm good either way. It should probably be demoed to Product regardless.

brianjgeiger:
Also I'm setting this to be linked to [EOSF-179] and closing the other issue.

app/routes/content.js
```diff
@@ -9,8 +9,8 @@ import permissions from 'ember-osf/const/permissions';
 // Error handling for API
 const handlers = new Map([
     // format: ['Message detail', 'page']
-    ['Authentication credentials were not provided.', 'forbidden'], // 401
-    ['You do not have permission to perform this action.', 'forbidden'], // 403
+    ['Authentication credentials were not provided.', 'page-not-found'], // 401
```
pattisdr: Looks good.

PR #249, opened 2017-01-13
[EOSF-447] Feature/yarn

binoculars:
## Ticket

https://openscience.atlassian.net/browse/EOSF-447

## Purpose

Updates the node package installation process to use yarn instead
Update ember-cli for the project

## Changes

Install yarn https://yarnpkg.com/

To get tests passing locally, you may have to do the setup steps for ember-cli in https://github.com/ember-cli/ember-cli/releases/tag/v2.10.0


brianjgeiger:
Is there anything in this that might have to cause a deployment configuration change for devops?

abought:
Good point.

We should probably use Yarn for deployment (as well as just Travis). Pinging @mattclark to find out which repo would need changing.

binoculars:
Yes, DevOps will need to have yarn installed and then run `yarn` or `yarn install` instead of `npm install`. Assuming, the `NODE_ENV` is properly set to `production`, no additional flags would need to be passed. Otherwise, use flag `--production`. I would also recommend caching yarn's cache directory.

README.md
```diff
@@ -13,12 +13,15 @@ You will need the following things properly installed on your computer.
...
+* [Yarn](https://yarnpkg.com/)
 * [Bower](http://bower.io/)
 * [Ember CLI](http://ember-cli.com/)
 * [PhantomJS](http://phantomjs.org/)
 
 ## Installation
-* `npm install`
+
+* `git clone` this repository
+* `yarn`
```
abought: Notes:

1. For new user facing instructions, we try to avoid vague shorthand in favor of explicit commands fully written out (eg `yarn install` instead of `yarn`)

2. Per flowdock discussion, please note that Travis CI will need to run the command with an optional flag as follows: `yarn install --pure-lockfile` ([ref](https://github.com/yarnpkg/yarn/issues/570))

3. When I ran `yarn install` without the above flag, on this branch, I got changes to my lockfile. Can you be sure the file is regenerated against the results of a clean `node_modules` directory?

package.json
```diff
@@ -27,7 +27,7 @@
     "ember-ajax": "^2.0.1",
     "ember-autosave": "2.0.1",
     "ember-bootstrap": "0.7.2",
-    "ember-cli": "^2.7.0",
+    "ember-cli": "^2.10.0",
```
binoculars: ember-cli and ember are versioned separately. We're still using ember (core) 2.7 in bower.json. Once we get to a working state with the current set of node modules and bower components, then I plan on doing another PR to update them to the latest LTS version and fix any issues that arise as a result of those updates.

tests/index.html
```diff
@@ -10,9 +10,9 @@
     {{content-for "head"}}
     {{content-for "test-head"}}
 
-    <link rel="stylesheet" href="assets/vendor.css">
-    <link rel="stylesheet" href="assets/preprint-service.css">
-    <link rel="stylesheet" href="assets/test-support.css">
+    <link rel="stylesheet" href="{{rootURL}}assets/vendor.css">
```
binoculars: Updates from ember-cli to get the tests working correctly. A large portion of the people developing on preprints were having issues because the ember-cli is outdated.

README.md
binoculars: Yarn wants you to use `yarn`, the default command for install (it's more of their convention than a convenience), but I won't bikeshed this. Changed it in the latest commit.

I changed everything to use --pure-lockfile as you suggested.

Dockerfile
```diff
@@ -1,8 +1,11 @@
-FROM node:6
+FROM node:boron
 
 RUN apt-get update \
     && apt-get install -y \
         git \
+        # Next 2 needed for yarn
+        apt-transport-https \
```
abought: I can't comment on these Dockerfile settings. Does this file need additional separate review from devops (MattC or Michael)? (ok if they already signed off, otherwise a blocker for merge)

README.md
```diff
@@ -55,10 +58,16 @@ There are a few scripts to run to populate your local preprint providers list, a
...
 ### Running Tests
+You can run tests either with ember installed on your machine or by using [Docker](https://docs.docker.com/engine/getstarted/step_one/)
 
+#### On your local machine
 * `ember test`
 * `ember test --server`
 
+#### With Docker
+* `docker build --tag preprints .`
+* `docker run preprints`
```
abought: Confirmed this method of running tests works for me locally.

Dockerfile
```diff
@@ -27,8 +39,8 @@ RUN cd /tmp \
 RUN mkdir -p /code
 WORKDIR /code
 
-COPY ./package.json /code/package.json
-RUN npm install
+COPY ./package.json ./yarn.lock /code/
```
binoculars: @icereval explained this to me, the intermediate containers get built after each command, so to keep them as small as possible, it is necessary to separate them out.

Dockerfile
binoculars: Michael actually wrote this section on my machine when we were getting me updated to work with docker. These steps are necessary, because the packages are not included in the default node container to add apt repo with https. We could install yarn via curl piped through bash, but it's probably less secure.

abought: Fair enough. As long as DevOps has reviewed it already this is ok by me.

PR #252, opened 2017-01-18
Fix Special Char for Preprints on the front end [EOSF-202]

sheriefvt:
## Purpose

Fix bugs related to two tickets EOSF-202 and EOSF-178

See attached screenshots
image

image

EOSF-202:
- Ampersands don't render properly on preprints. This affects preprint title, abstract, and tags.
- The ampersands are not rendering properly in the email title when use share via email.

EOSF-178
- Fix the contact us button.
- Fix the share via email button.

## Changes

EOSF-202:
- Created a new util function fixSpecialChar to handle the ampersand. 
- Used an existing helper function safeMarkup.

EOSF-178
- Change action event from onclick to onbeforeclick.

## Side effects 

For now other characters < or > are not handled in the edit mode. 
See the attached screenshot. 
image

This can be fixed by going back to previous commit fe3b9185f1287719e16c84b1a213477c27a541fb.
image

## Ticket

https://openscience.atlassian.net/browse/EOSF-202.
https://openscience.atlassian.net/browse/EOSF-178


app/controllers/submit.js
```diff
@@ -525,11 +541,12 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
                     action: 'click',
                     label: 'Preprints - Submit - Save and Continue, New Component, Copy File'
                 });
-            node.addChild(this.get('nodeTitle'))
+            node.addChild(fixSpecialChar(this.get('nodeTitle')))
```
abought: Is this really necessary? Bleach lets existing escape sequences pass untouched. There's a chance that setting the new node title to the old title will give you back the desired result regardless...

app/templates/components/preprint-form-header.hbs
```diff
@@ -26,10 +26,10 @@
 {{#if (and (not isOpen) (eq name 'Upload'))}}
     <div class="m-t-md preprint-header-preview">
         {{#if preprintNode}}
-            <p><em class="m-r-md"> {{t "components.preprint-form-header.location"}}: </em> {{get-ancestor-descriptor preprintNode}} {{preprintNode.title}}</p>
+            <p><em class="m-r-md"> {{t "components.preprint-form-header.location"}}: </em> {{get-ancestor-descriptor preprintNode}} {{safe-markup preprintNode.title}}</p>
```
abought: If the value ever comes back from the server with HTML in it*- wouldn't this cause the node title to be displayed as raw html, rather than displayed as written?

The `safe-markup` helper should be used in highly restricted, very limited cases where there is absolutely no user-entered input involved. I would tend to discourage its use for displaying values in this way. If you have any questions I'd be happy to show examples or discuss the idea; just stop by anytime.

Two sample cases:
1. If we remove bleach (an HTML sanitizer) from the backend in the future
2. If we get a filename from a source like google drive, where the rules for a filename are outside our control and we do not run any transformations/ sanitizing on these values

sheriefvt: Yes sure, thanks. Will finish fixing the styling and come to you.

abought: Per previous discussion, remember that this PR **cannot** be merged so long as this line is present (see also line 32 below, etc). 

The `safe-markup` helper should not be used with user-entered text.

app/controllers/submit.js
```diff
@@ -699,7 +714,7 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
             let model = this.get('model');
             // Saves off current server-state basics fields, so UI can be restored in case of failure
             let currentAbstract = node.get('description');
-            let currentTags = node.get('tags').slice(0);
+            let currentTags = node.get('tags').slice(0).map(fixSpecialChar);
```
abought: Since this is a display concern, I might suggest fixing tag display in the template rather than the controller. You'd be using the template helper instead of the util function.

app/controllers/submit.js
```diff
@@ -725,6 +740,7 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
 
             node.save()
                 .then(() => {
+                    this.set('node.description', this.get('node.description'));
```
abought: Can you explain the purpose of this line? It seems strange to set/mutate a value that you already know. I'm not sure this is necessary and might have side effects.

app/controllers/submit.js
```diff
@@ -733,6 +749,7 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
                         }
                         model.save()
                             .then(() => {
+                                this.set('node.description', this.get('node.description'));
```
abought: See above comment- setting a value we already have seems unnecessary, and may have side effects.

app/controllers/submit.js
```diff
@@ -746,6 +763,7 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
                         model.set('license', this.get('basicsLicense.licenseType'));
                         model.save()
                             .then(() => {
+                                this.set('node.description', this.get('node.description'));
```
abought: See previous notes.

app/controllers/submit.js
```diff
@@ -779,7 +797,8 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
                 });
             let tags = this.get('basicsTags').slice(0);
             Ember.A(tags);
-            tags.pushObject(tag);
+
+            tags.pushObject(fixSpecialChar(tag));
```
abought: This seems unnecessary. Special characters get messed up because of the server. User input that has not yet been to the server should not need to be altered/ "fixed". We should send the server exactly what the user intended to write.

app/mixins/setup-submit-controller.js
```diff
@@ -61,7 +63,7 @@ export default Ember.Mixin.create({
         controller.set('filePickerState', 'existing'); // In edit mode, dealing with existing project
         controller.set('existingState', 'new'); // In edit mode, only option to change file is to upload a NEW file
         controller.set('node', node);
-        controller.set('nodeTitle', node.get('title'));
+        controller.set('nodeTitle', fixSpecialChar(node.get('title')));
```
abought: I'm not sure this is necessary: the `fixstring` transform means that data from the server should just naturally look ok when we access it; we should not need to "fix" it further for display.

Also, in general, I'd encourage us to keep display logic in the template when feasible, eg using the template helper instead of modifying the property on the controller.

app/templates/components/search-result.hbs
```diff
@@ -81,7 +81,7 @@
                     {{#if result.tags}}
                         <div class="m-t-sm">
                             {{#each result.tags as |tag index|}}
-                                <span class="badge">{{tag}}</span>
+                                <span class="badge">{{fix-special-char tag}}</span>
```
abought: This should work. Thanks!

app/templates/index.hbs
```diff
@@ -95,7 +95,7 @@
                             {{t "index.services.bottom.div.line3"}}
                         </div>
                     </p>
-                    <a href="mailto:contact@osf.io" class="btn btn-info btn-lg" onclick={{action "click" "link" "Preprints - Index - Contact"}}>{{t "index.services.bottom.contact"}}</a>
+                    <a href="mailto:contact@osf.io" class="btn btn-info btn-lg" onbeforeclick={{action "click" "link" "Preprints - Index - Contact"}}>{{t "index.services.bottom.contact"}}</a>
```
abought: What is the reason for binding the `click` action to the `onbeforeclick` event?

app/controllers/submit.js
```diff
@@ -268,15 +275,19 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
     basicsTags: Ember.computed('node', function() {
         // Pending tags
         let node = this.get('node');
-        return node ? node.get('tags') : Ember.A();
+        let newTags = null;
+        if (node != null) {
+            newTags = node.get('tags').slice(0).map(fixSpecialChar);
```
abought: Per our discussion: can we stick to consistently transforming values in one place, eg fixing display of characters in the template instead of the controller?

Having a consistent standard keeps display logic in the display layer, and makes it easier to find/guess where a value is being changed later.

A major exception to that rule would be if this value were being passed to the tags widget- there is some special logic for the `jquery.tagsinput` widget, and in that case it would be ok to modify the value in the controller.

app/controllers/submit.js
```diff
@@ -246,7 +248,12 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
     }),
     titleChanged: Ember.computed('node.title', 'nodeTitle', function() {
         // Does the pending title differ from the title already saved?
-        return this.get('node.title') !== this.get('nodeTitle');
+        if (this.get('nodeTitle') != null) {
+            return this.get('node.title') !== this.get('nodeTitle').toString();
```
abought: Can you clarify when nodeTitle needs to be converted to a string? I thought it was bound to the value of a text box?

app/controllers/submit.js
```diff
@@ -531,6 +543,7 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
                     this.set('parentNode', node);
                     this.set('node', child);
                     this.set('basicsAbstract', this.get('node.description') || null);
+                    this.set('nodeTitle', this.get('node.title'));
```
abought: Is this related to special characters, or something else?

app/controllers/submit.js
sheriefvt: will take care of this

app/templates/index.hbs
sheriefvt: This was a temporary fix. It is addressed now in pull request for ticket EOSF-178.

abought:
Can you please move the JIRA ticket back to CR and assign me when this is ready for review?

PR #253, opened 2017-01-20
[EOSF-445] Link to download file uses GUID

hmoco:
## Purpose
Google Scholar needs the file download link to be hosted in the same domain of the page it is linked, so, to allow it to download our preprints, the Download preprint links to the file page with the `action=download` query param, hosted under osf.io. 

## TODO
https://openscience.atlassian.net/browse/OSF-7375, will make it so the url does not need to be built on Ember. 

## Ticket
https://openscience.atlassian.net/browse/EOSF-445


hmoco:
@binoculars do you know why this might be failing tests?

binoculars:
@hmoco yes, we found out earlier today that --prefer-offline is broken on travis, just remove it from the .travis.yml

binoculars:
Tested it locally and it's working.

bitmoji

Let's try to use the unit tests more so we can get our code coverage % 🆙

bitmoji

PR #256, opened 2017-01-26
Fix contact us button and share via email button

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-178

## Purpose

Fix the contact us button

## Changes

change onclick action to onbeforeclick

## Notes for QA
From reviewer: "Any time there are events with requests involved, I encourage QA to test this with multiple browsers, and look at whether analytics events are sent as intended in addition to core functionality."


pattisdr:
Can you confirm that the google analytics event is still fired with this change?

sheriefvt:
@pattisdr I restored the google analytics event.

pattisdr:
Thanks - can you verify that it's actually firing? I think I have seen actions not firing on onbeforeclick.

sheriefvt:
Ok sure..will look closely at this and get back to you.

sheriefvt:
I think google analytics event is not firing. Will try to find a solution by tomorrow.

sheriefvt:
Hopefully, google analytics event is working now. Thanks @binoculars, @pattisdr for help.

app/controllers/index.js
```diff
@@ -60,4 +60,23 @@ export default Ember.Controller.extend(Analytics, {
...
+            metrics.trackEvent({
+                category,
+                action,
+                label
+            });
+
+            if (label.includes('email'))
+              return;
+
+            window.open(href, '', 'menubar=no,toolbar=no,resizable=yes,scrollbars=yes,width=600,height=400');
```
abought: Hmm.... In general, MDN recommends against using this method:
https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Avoid_resorting_to_window.open()

It prefers that developers use links directly as intended. What is the purpose of opening a new window for the emails?

It appears that href is filled in here; there's a chance of user-entered text (node.title) becoming part of the link url. That's not great, though it appears the characters are escaped in this case:
https://github.com/abought/ember-preprints/blob/381fbb8e6017a0553d66a77d217825721a4c46c8/app/controllers/content.js#L96-L96

We should avoid using `window.open` and prefer an anchor link directly if possible, for reasons cited by MDN above. If we must use this method, the value of href should be treated carefully, and features such as `noopener` should be very explicitly used.

abought:
Changes look ok, but I didn't have time to functionally test this with GA set up. Any time there are events with requests involved, I encourage QA to test this with multiple browsers, and look at whether analytics events are sent as intended in addition to core functionality.

PR #257, opened 2017-01-26
Improved pagination for preprints search results [EOSF-378]

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-378 

## Purpose

Initially, users can only go forward and backward in the search results. There is a need to improve the pagination for preprints.

## Changes

Adopt and re-use a page controls component similar to SHARE.


abought:
Haven't run the code or seen a demo- but conceptually, this sounds like the osf-paginator widget in ember OSF.

The code in our shared library of widgets was pulled from the old OSF as a temporary fix, and perhaps this may become a more ember-friendly alternative. (there's no guarantee we have to keep the existing ember-osf widget, but I don't think we should have two widgets)

Perhaps @brianjgeiger  could weigh in on the best approach to using a single pagination widget, with code living in `ember-osf`?

See existing widget documentation for comparison:
http://centerforopenscience.github.io/ember-osf/classes/osf-paginator.html

brianjgeiger:
We should not have two widgets. We should have one widget with some options.

pattisdr:
We also will have Ember-SHARE's pagination widget, page-controls in Ember-OSF. I pulled in a lot of code from ember-share here, into ember-osf. http://github.com/CenterForOpenScience/ember-osf/pull/169. The purpose was to be able to build the entire discover page with a component. 

FYI the osf-paginator component is being used in the authors section of preprints.

I agree that consolidation is good, but keep in mind this will affect multiple applications so it would need to work for all cases.

abought:
> I agree that consolidation is good, but keep in mind this will affect multiple applications so it would need to work for all cases.

Agreed!

But on the other hand- given that list, I think that indicates the high value of making a decision and consolidating sooner, rather than later. I don't think we should all be reimplementing bootstrap paginator widgets.

app/controllers/discover.js
```diff
@@ -213,9 +213,19 @@ export default Ember.Controller.extend(Analytics, {
             return this.set('results', results);
         });
     },
+    totalPages: Ember.computed('numberOfResults', 'size', function() {
+        return Math.ceil(this.get('numberOfResults') / this.get('size'));
```
abought: I *think* this page uses SHARE search results, which are a separate network request not made to the OSF APIv2. In that case this is fine.

But if you ever need to paginate a search from the OSF API, we historically would automatically calculate a "total number of pages" value in the serializer. I mention it here in case it's useful, but also because I vaguely recall hearing someone say that the API changed, and I'm not sure the below is still relevant:
https://github.com/abought/ember-osf/blob/9bfb62aae2fc18728fea5546ab64fa4ed0aea886/addon/serializers/osf-serializer.js#L123-L123

app/controllers/discover.js
```diff
@@ -213,9 +213,19 @@ export default Ember.Controller.extend(Analytics, {
...
+    }),
+
     maxPages: Ember.computed('numberOfResults', function() {
         return ((this.get('numberOfResults') / this.get('size')) | 0) + (this.get('numberOfResults') % 10 === 0 ? 0 : 1);
     }),
+
+    clampedPages: Ember.computed('size', 'totalPages', function() {
+        let maxPages = Math.ceil(10000 / this.get('size'));
```
abought: Where does this number come from? What is a `clampedPage`? Why do we impose a maximum at all, instead of letting the API decide how many pages there really are?

app/controllers/discover.js
```diff
@@ -323,6 +333,11 @@ export default Ember.Controller.extend(Analytics, {
             }
         },
 
+        loadPage(pageNumber) {
+            this.set('page', pageNumber);
+            this.loadPage();
```
abought: Having an action with the same name as a method can be very confusing. Consider calling the latter `this._loadPage`, or somesuch.

app/components/preprint-form-authors.js
```diff
@@ -236,16 +236,17 @@ export default CpPanelBodyComponent.extend(Analytics, {
...
             let query = this.get('query');
             if (query) {
                 this.attrs.findContributors(query, current).then(() => {
                     this.set('addState', 'searchView');
-                    this.set('currentPage', current);
                 }, () => {
                     this.get('toast').error('Could not perform search query.');
                     this.highlightSuccessOrFailure('author-search-box', this, 'error');
+                }).then(() => {
+                   this.set('currentPage', current);
```
abought: It's not necessary to use a separate `.then`. But if it recommended to use a separate `.then` vs `.catch`. The specific line affected by this is Line 245.

A general recommendation of good habit is to rewrite statements like: 
`findContributors().then( success, failure )`

to
`findContributors().then( success).catch(failure )`

app/controllers/discover.js
```diff
@@ -213,9 +213,20 @@ export default Ember.Controller.extend(Analytics, {
...
     },
+    totalPages: Ember.computed('numberOfResults', 'size', function() {
+        return Math.ceil(this.get('numberOfResults') / this.get('size'));
+    }),
+
     maxPages: Ember.computed('numberOfResults', function() {
         return ((this.get('numberOfResults') / this.get('size')) | 0) + (this.get('numberOfResults') % 10 === 0 ? 0 : 1);
     }),
+
+    // TODO update this property if a solution is found for the elastic search limitation.
```
abought: Note to self: Sherif will update this with ticket number.

abought:
This PR has developed merge conflicts.

Will pick up for review once those are addressed. Don't forget to push the ticket number in comment + promise refactor; this PR is very close to being able to go forward.

sheriefvt:
I am working on it now..thanks

yarn.lock
```diff
@@ -2135,9 +2135,9 @@ ember-new-computed@^1.0.2:
   dependencies:
     ember-cli-babel "^5.1.5"
 
-"ember-osf@git+https://github.com/pattisdr/ember-osf.git#c4e87ff18b3f9125b87fbd3ce05e2c41eebdc3b6":
+"ember-osf@git+https://github.com/CenterForOpenScience/ember-osf.git#97a4f00dd86f360b561f3ee63b77297b3e6e0f1e":
```
abought: Confirmed this is just making us more consistent with latest package.json; thanks.

PR #259, opened 2017-01-31
[EOSF-435] Add "see more" for long abstracts on view preprint page.

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-435

## Purpose
Allow citations and other non-abstract pieces of the view preprint page to be seen right away.

## Changes
Add "see more" to the abstract that expands it if it is too long.

### DIFF
--- a/app/controllers/content.js
+++ b/app/controllers/content.js
@@ -129,12 +129,15 @@ export default Ember.Controller.extend(Analytics, {
     }),
 
     useShortenedDescription: Ember.computed('node.description', function() {
-        return this.get('node.description').length > 350;
+        return this.get('node.description') ? this.get('node.description').length > 350 : false;
     }),
 
-    shortenedDescription: Ember.computed('node.description', function() {
+    description: Ember.computed('node.description', 'expandedAbstract', function() {
         // Get a shortened version of the abstract, but doesnt cut in the middle of word by going
         // to the last space.
+        if (this.get('expandedAbstract')) {
+            return this.get('node.description');
+        }
         let text = this.get('node.description').slice(0, 350).split(' ');
         text.pop();
         return text.join(' ') + ' ...';
--- a/app/controllers/content.js
+++ b/app/controllers/content.js
@@ -45,6 +45,7 @@ export default Ember.Controller.extend(Analytics, {
     fullScreenMFR: false,
     expandedAuthors: true,
     showLicenseText: false,
+    expandedAbstract: false,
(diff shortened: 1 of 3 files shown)


app/templates/content.hbs
```diff
@@ -79,7 +79,17 @@
                     </div>
                     <div class="p-t-xs">
                         <h4 class="p-v-md f-w-md"><strong>{{t "global.abstract"}}</strong></h4>
-                        <p class="abstract">{{node.description}}</p>
+                        {{#if useShortenedDescription}}
+                            {{#if expandedAbstract}}
+                                <p class="abstract">{{node.description}}</p>
+                                <a href='#' {{action 'expandAbstract'}}>See less</a>
```
abought: Accessibility note: we try to avoid using `<a>` tags for interactive elements not related to navigation.

To better communicate the intent of this element to screen readers, consider a `<button class="btn-link">` instead.

app/controllers/content.js
```diff
@@ -127,6 +128,18 @@ export default Ember.Controller.extend(Analytics, {
         return text;
     }),
 
+    useShortenedDescription: Ember.computed('node.description', function() {
+        return this.get('node.description').length > 350;
+    }),
+
+    shortenedDescription: Ember.computed('node.description', function() {
```
abought: It seems that `useShortenedDescription` could be removed, in favor of a single property (`description`) that just chose what text to display based on length. This would let you consolidate some redundant markup in `content.hbs` as well.

twsacco:
Thanks, @abought ! QA appreciates it!

app/templates/content.hbs
```diff
@@ -79,7 +79,12 @@
                     </div>
                     <div class="p-t-xs">
                         <h4 class="p-v-md f-w-md"><strong>{{t "global.abstract"}}</strong></h4>
-                        <p class="abstract">{{node.description}}</p>
+                        {{#if useShortenedDescription}}
+                            <p class="abstract">{{description}}</p>
+                            <button class='btn-link' {{action 'expandAbstract'}}>{{if expandedAbstract 'See less' 'See more'}}</button>
```
binoculars: In the future, these strings need to go into the translations file, not in the template.

brianjgeiger: It'd be cool if we could lint for that.

binoculars: The i18n linter does some of this, but doesn't catch things inside handlebars expressions, unfortunately

PR #261, opened 2017-01-31
[EOSF-437] Use ember markdown-it to render abstract

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-437

## Purpose
Render markdown on abstract content, both in the edit/submit page and the content page.

## Changes
Use of `ember-markdown-it` package.


laurenbarker:
Closing since this went back to product (and is possibly handled on the SHARE side)

PR #263, opened 2017-02-01
[EOSF-172] Truncate author list if it will exceed 2 lines

baylee-d:
## Ticket

https://openscience.atlassian.net/browse/EOSF-172

## Purpose

Restricts the amount of authors shown under preprints search to 10 per preprint.

## Changes

Instead of looping through the original collection, loops through a new ember function that pre-filters the collection and restricts the amount returned to 10.

### DIFF
--- a/app/components/search-result.js
+++ b/app/components/search-result.js
@@ -34,8 +34,8 @@ export default Ember.Component.extend(Analytics, {
         for (let i = 0; i < filtered.length; i++) {
             nameArr[i] = filtered[i];
             numNames++;
-            if(numNames === 10 || i === filtered.length - 1) {
-                if (numNames < filtered.length) 
+            if (numNames === 10 || i === filtered.length - 1) {
+                if (numNames < filtered.length)
                     nameArr[i].originalLength = filtered.length;
                 return nameArr;
             }
--- a/app/components/search-result.js
+++ b/app/components/search-result.js
@@ -34,7 +34,7 @@ export default Ember.Component.extend(Analytics, {
         for (let i = 0; i < filtered.length; i++) {
             nameArr[i] = filtered[i];
             numNames++;
-            if(numNames === 6 || i === filtered.length - 1) {
+            if(numNames === 10 || i === filtered.length - 1) {
                 if (numNames < filtered.length) 
                     nameArr[i].originalLength = filtered.length;
                 return nameArr;
--- a/app/components/search-result.js
+++ b/app/components/search-result.js
@@ -28,28 +28,12 @@ export default Ember.Component.extend(Analytics, {
     }),
 
     shortContributorList: Ember.computed('result', function() {
-        let nameArr = [];
-        let numNames = 0;
-        let filtered = this.get('result.contributors').filter(this.isBibliographic);
-        for (let i = 0; i < filtered.length; i++) {
-            nameArr[i] = filtered[i];
-            numNames++;
-            if (numNames === 10 || i === filtered.length - 1) {
-                if (numNames < filtered.length)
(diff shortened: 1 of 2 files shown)


app/components/search-result.js
```diff
@@ -27,6 +27,30 @@ export default Ember.Component.extend(Analytics, {
...
+    shortContributorList: Ember.computed('result', function() {
+        let nameArr = [];
+        let numNames = 0;
+        let filtered = this.get('result.contributors').filter(this.isBibliographic);
+        for (let i = 0; i < filtered.length; i++) {
+            nameArr[i] = filtered[i];
+            numNames++;
+            if (numNames === 10 || i === filtered.length - 1) {
+                if (numNames < filtered.length)
+                    nameArr[i].originalLength = filtered.length;
```
abought: It seems like you're using the presence of this (numeric) property as a boolean flag to indicate that this item is the last item in the list, *and* that there are more results?

Is there a more explicit way to do this, for clarity? Perhaps a separate (boolean) computed that compared the lengths of the two arrays?

app/components/search-result.js
```diff
@@ -27,6 +27,14 @@ export default Ember.Component.extend(Analytics, {
         return result.description.slice();
     }),
 
+    shortContributorList: Ember.computed('result', function() {
+        let filtered = this.get('result.contributors').filter(item => !!item.users.bibliographic);
+        let nameArr = filtered.slice(0, Math.min(10, filtered.length));
+        if (nameArr.length < filtered.length)
+            nameArr[nameArr.length - 1].originalLength = filtered.length;
```
abought: Hmm. I would still prefer tracking `hasMoreContributors` as a separate Ember.computed property. 

It's a bit confusing to use the presence or absence of a numeric property on the last item as a boolean flag, especially when the flag is controlling display of something not directly related to that item.

Here is an example (pseudocode; haven't run it yet) demonstrating the idea. It's a little annoying having so many properties, but computeds are cached, so this only runs once every time the dependent keys change- so you're not having to duplicate work to reuse logic.

search-result.js:
```
hasMoreContributors: Ember.computed('justContributors', 'shortContributorsList', function () {
  return this.get('shortContributorsList').length < this.get('justContributors').length ;
});


search-result.hbs
```
{{#each shortContributorList as |contributor|}}
  {{author-link contributor=contributor}}
{{/each}}
{{if hasMoreContributors "..."}}
```

Comment (abought) [inline app/components/search-result.js]: That said, this is a step in the right direction; much appreciated on the work so far. :)

File: app/components/search-result.js [hunk 3]
```diff
@@ -27,6 +27,16 @@ export default Ember.Component.extend(Analytics, {
         return result.description.slice();
     }),
 
+    shortContributorList: Ember.computed('result', function() {
+        let filtered = this.set('justContributors', this.get('result.contributors').filter(item => !!item.users.bibliographic));
```
abought: In general, we try to avoid creating a computed property that has side effects (changes other things)- this can lead to odd behavior that's hard to debug.

Instead, `justContributors` should be a separate computed that `shortContributorsList` watches instead of `results`.

Perhaps it would be more efficient to stop by and pair program to make sure we nail this down? I'm free most of the afternoon.

app/components/search-result.js
```diff
@@ -27,6 +27,16 @@ export default Ember.Component.extend(Analytics, {
...
+        let nameArr = this.set('shortContributors', filtered.slice(0, Math.min(10, filtered.length)));
```
abought: What is the advantage of having both `shortContributorsList` and `shortContributors`? They seem to be storing the same variable?

PR #266, opened 2017-02-03
[EOSF-487] add permission language to agrixiv

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-487

## Purpose

Add Permission language for -rxiv in agrixiv page and keep Sciello from showing an error

### DIFF
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -252,7 +252,8 @@ export default {
         },
         'permission-language':{
             arxiv_trademark_license,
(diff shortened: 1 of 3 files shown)


brianjgeiger:
@hmoco Could you include a screenshot of the new language in place?

brianjgeiger:
Thinking about this just a bit more, make a `no_trademark` language item, make it blank, assign it to permissionLanguage, and then make a test to ensure that  all of the fields are properly filled out, please.

brianjgeiger:
In general, before and after pictures are good for these because they will give QA a good idea of what to look for.

PR #267, opened 2017-02-06
Preliminary support for code coverage in preprints [#EOSF-485]

abought:
## Ticket

https://openscience.atlassian.net/browse/EOSF-485

🦑  This PR has been updated a few times- feel free to merge-and-squash.

## Purpose

Add code coverage measurements to the ember-preprints repo.

This is a developer facing change. It will require coordination with `coveralls.io` to activate the service for this repo.

## Changes
Similar to those added for ember-osf.
See: https://github.com/CenterForOpenScience/ember-osf/pull/175

## Side effects
- This is a developer facing change and is not expected to affect users
- Side effects may include dizziness, chattier pull requests that guilt people into writing tests, and occasional red "your coverage went down" warnings that can be appeased through coveralls settings and/or writing tests


package.json
```diff
@@ -26,13 +27,15 @@
   "devDependencies": {
     "autoprefixer": "^6.3.7",
     "broccoli-asset-rev": "^2.4.2",
+    "coveralls": "^2.11.16",
```
brianjgeiger: Aren't we moving towards pinned versions?

abought: Hmm. Strictly speaking the yarn lockfile will pin this rather explicitly in practice, if someone chooses to install that way. 

Have we made an official decision to pin *all* dependencies going forward, then? I knew we were versioning EOSF, but would be interested to know the details of any larger policy change.

brianjgeiger:
https://openscience.atlassian.net/browse/EOSF-454 And in general, yes, we should know what we're using and should make a conscious decision to use it. Which will mean regularly checking for updated versions and verifying that they do what we want them to do.

abought:
> https://openscience.atlassian.net/browse/EOSF-454 And in general, yes, we should know what we're using and should make a conscious decision to use it. Which will mean regularly checking for updated versions and verifying that they do what we want them to do.

Good to know. Worth announcing to the entire team for all our projects.

I will make the change right now, but long term I advise looking into a service such as [GreenKeeper](https://greenkeeper.io/), as this is a highly labor-intensive task that we can and should automate. I'll investigate adding it to the demo app as a first trial.

The yarn lockfile should already be a record of a known working state, and this service will automatically tell us when specific breaking changes come along (in case someone abuses semver and unintentionally introduces a bug).

abought:
Changes pushed; will wait until CI is finished to move the ticket. This PR is a good candidate for merge-and-squash.

PR #268, opened 2017-02-06
[EOSF-238] Change scielo text to black and only have board container if it exists

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-238

## Purpose & Changes
Fix bug on scielo, changing color to black to fix a lot of visibility issues, and only have the advisory container if the current provider actually has an advisory board.


brianjgeiger:
@hmoco Please do screenshots for these things so we can see what has changed.

PR #271, opened 2017-02-09
[EOSF-495] Add new logos for PsyArXiv

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-495

## Purpose
Update PsyArxiv Logos

## Changes
Before:
screen shot 2017-02-09 at 9 29 20 am
After (ignoring the html on text):
screen shot 2017-02-09 at 9 29 27 am


hmoco:
@brianjgeiger, the images being replaced (and most of the 'small' options) are about 10k, and the new ones are at about 50. If you could downscale these that would be awesome.

PR #275, opened 2017-02-10
[EOSF-285] Styling for Selected Sub-disciplines List

baylee-d:
## Ticket

- https://openscience.atlassian.net/browse/EOSF-285

## Purpose

When creating a preprint, disciplines and sub-disciplines would smash together when the screen became too small to fit on one line.  This will fix the styling of the disciplines and sub-disciplines when they break onto new lines.

## Changes

- Changed HTML to add another container around each listed item and changed the styling to match. 
- Moved 'x' icon to be within the new div containers.
 
## Side effects

- None expected

## QA Notes

- When a user adds a new sub-discipline to the list, all listed items already selected will have an arrow at the end unless they are the last item in the list.

- In the case that the screen size is too small to fit all items in the list on one line, they will break up and move to the next line in order to fit.
screen shot 2017-02-21 at 10 21 58 am

- If the screen is too small to fit the individual elements on one line, the internal text will wrap to another line to fit.
screen shot 2017-02-21 at 10 22 26 am

- When this happens, the side arrow _will not_ increase it's size to match the new height of the container.  This is expected behavior.

### DIFF
--- a/app/components/subject-picker/template.hbs
+++ b/app/components/subject-picker/template.hbs
@@ -2,7 +2,7 @@
     {{#each selected as |subject|}}
         <span class="subject">
             {{#each subject as |segment|}}
-                <div class="subject-container"><b class="subject-container-text">{{segment.text}} <button class="cancel-subject" aria-label="Remove subject" title="Remove" {{action 'deselect' subject}}><i class="fa fa-close"></i></button></b></div>
+                <div class="subject-container"><span class="subject-container-text">{{segment.text}}<button class="cancel-subject fa fa-close" aria-label="{{t 'submit.body.remove_subject_aria'}}" title="Remove" {{action 'deselect' subject}} /></span></div><span class="right-arrow"></span>
             {{/each}}
         </span>
     {{/each}}
--- a/app/components/subject-picker/template.hbs
+++ b/app/components/subject-picker/template.hbs
@@ -2,7 +2,7 @@
     {{#each selected as |subject|}}
         <span class="subject">
             {{#each subject as |segment|}}
-                <b>{{segment.text}}</b>
+                <div class="subject-container"><b>{{segment.text}}</b></div>
             {{/each}}
             {{fa-icon 'times' click=(action 'deselect' subject)}}
         </span>
--- a/app/components/subject-picker/template.hbs
+++ b/app/components/subject-picker/template.hbs
@@ -2,9 +2,8 @@
     {{#each selected as |subject|}}
         <span class="subject">
             {{#each subject as |segment|}}
-                <div class="subject-container"><b>{{segment.text}}</b></div>
+                <div class="subject-container"><b>{{segment.text}} <span class="cancel-subject" {{action 'deselect' subject}}></span></b></div>
             {{/each}}
-            {{fa-icon 'times' click=(action 'deselect' subject)}}
         </span>
     {{/each}}
 </div>
--- a/app/components/subject-picker/template.hbs
+++ b/app/components/subject-picker/template.hbs
@@ -2,7 +2,7 @@
     {{#each selected as |subject|}}
         <span class="subject">
             {{#each subject as |segment|}}
-                <div class="subject-container"><b>{{segment.text}} <span class="cancel-subject" {{action 'deselect' subject}}></span></b></div>
+                <div class="subject-container"><b class="subject-container-text">{{segment.text}} <button class="cancel-subject" aria-label="Remove subject" title="Remove" {{action 'deselect' subject}}><i class="fa fa-close"></i></button></b></div>
             {{/each}}
         </span>
     {{/each}}
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -739,8 +739,10 @@ hr {
         color: white;
         overflow: hidden;
         padding: 0 2rem;
-        margin-right: 2rem;
-        margin-bottom: 1rem;
+        margin: {
+            right: 2rem;
+            bottom: 1rem;
+        }
         &:hover {
             & > .subject-container, & > .fa {
                 background-color: lighten($brand-primary, 10%);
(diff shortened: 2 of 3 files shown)


app/components/subject-picker/template.hbs
```diff
@@ -2,9 +2,8 @@
     {{#each selected as |subject|}}
         <span class="subject">
             {{#each subject as |segment|}}
-                <b>{{segment.text}}</b>
+                <div class="subject-container"><b>{{segment.text}} <span class="cancel-subject" {{action 'deselect' subject}}></span></b></div>
```
abought: This span appears to be a font-awesome icon. Since the only sign of an interactive element is pictorial, how will a blind user accessing the site via screenreader know that there is a clickable icon/button here? (see [FA accessibility guidelines](http://fontawesome.io/accessibility/) for recommendations?

Tips include:
1. Provide aria labels
2. Use markup semantically. Generally, avoid using `span` for something clickable, because to HTML, `span` just means that there is text there. Prefer `button` for something interactive, or `a` for something that triggers navigation to a different page. (if you use a button or a link, the mouse cursor changes will also change without you needing to write extra CSS, which is nice)

In combination, screen readers like Mac OS VoiceOver use these cues to help blind users understand a page. If you ever want, we can arrange a demo to help show how the page looks from that perspective; just let me know.

app/styles/app.scss
```diff
@@ -725,74 +725,79 @@ hr {
...
-            cursor: pointer;
-            transition: color 0.1s linear;
-            &:hover {
-                color: lighten(red, 40%);
-            }
-            &:active {
-                color: lighten(red, 20%);
+                & > b > .cancel-subject {
+                    &:before {
+                        font: {
```
abought: For the most part, we use font awesome's predefined classes for this kind of thing when possible. It's more readable in the markup, and has the additional benefit of not redefining something that is already there. Eg, `<i class="fa fa-close"></i>`

I might be missing something- is there a particular advantage to redefining the CSS here?

app/styles/app.scss
```diff
@@ -725,74 +725,79 @@ hr {
...
-        & > .fa {
-            color: white;
-            cursor: pointer;
-            transition: color 0.1s linear;
-            &:hover {
-                color: lighten(red, 40%);
-            }
-            &:active {
-                color: lighten(red, 20%);
+                & > b > .cancel-subject {
```
abought: I notice that many of these selectors are extremely specific, and could break if the page saw even minor design changes. Is there a way to simplify this?

If I'm reading this correctly, this says:

"Only apply this style if this element is a child of a node with id preprint-form-subjects, which contains one or more elements called subject container, where the last of those contains a bold tag, in which case if that bolded element contains an element with class cancel-subject, only if this is the last subject container should we apply this style"

Some of this hierarchy might be necessary, but I'd encourage limiting the levels of nesting if at all possible. (including dependence on specific markup tags like `b`) We want the class to be unique enough not to accidentally style everything, but not so specific that a small design change breaks it entirely.

app/styles/app.scss
```diff
@@ -725,74 +725,79 @@ hr {
...
-        border-radius: 2px;
-        transition: background-color 0.1s linear;
+        margin: {
+            right: 2rem;
+            bottom: 1rem;
+        }
         &:hover {
-            background-color: lighten($brand-primary, 10%);
-            & > b:after {
+            & > .subject-container, & > .fa {
```
abought: There do not appear to be any children of subject-container with class fa at present, unless I'm missing something... 

If that is still true after any CSS refactoring you do, we generally encourage removing any style rules that are completely unused.

app/components/subject-picker/template.hbs
```diff
@@ -2,9 +2,8 @@
     {{#each selected as |subject|}}
         <span class="subject">
             {{#each subject as |segment|}}
-                <b>{{segment.text}}</b>
+                <div class="subject-container"><b class="subject-container-text">{{segment.text}} <button class="cancel-subject" aria-label="Remove subject" title="Remove" {{action 'deselect' subject}}><i class="fa fa-close"></i></button></b></div>
```
abought: Thanks for adding ARIA labels! Can you please move all hard-coded text into a translations file?

abought:
With the new PR changes, in Chrome 57, I am seeing some small visual anomalies- the arrow seems slightly off to the side and detached from the rectangle. Can you reproduce?
screen shot 2017-04-07 at 11 29 56 am

app/styles/app.scss
```diff
@@ -736,79 +736,77 @@ hr {
...
+    position: absolute;
+    display: inline;
+    &:before {
+        content: '';
+        border-top: 14px solid transparent;
+        border-bottom: 15px solid transparent;
+        border-left: 15px solid $brand-primary;
+        position: absolute;
+        margin-left: -16px;
+        margin-top: -14px;
```
abought: Any reason for the slight variation in border and margin sizes here? (14, 15, 16 px)? Asking in case this relates to the slightly alignment issues I was seeing.

Also, is there any particular reason for the move from `rem` to `px` units? (eg were we deliberately trying to prevent rescaling in this case?)

baylee-d:
@abought I fixed the previous issue with the px/rem but I wasn't able to recreate the issue you were seeing with the arrows (tested it in Chrome 57, as well as Safari and Firefox).  Not entirely sure what was happening there, but these changes might fix it.  Let me know if they still show up for you.

abought:
Baylee- I'm out this week, but if you're comfortable with alignment of arrows based on the sample provided, I'm ok passing this to another reviewer and letting further bug fixes fall under the banner of QA.

(So long as there are no actively known issues)

baylee-d:
I just checked again to make sure that there weren't any issues with this, and everything looked good.  If you're still okay with it, this issue should be alright being passed onto QA.  Thanks for all of the feedback and help with this!

PR #279, opened 2017-02-15
[EOSF-501] Improve Preprint download count wrapping

baylee-d:
## Ticket

https://openscience.atlassian.net/browse/EOSF-501

## Purpose

The download count for a preprint was breaking to the next line before the share icons.  This ticket fixes this issue.

## Changes

Added a new class to both containers in order to remove the original template styling for the width of each container.

## Screenshots 

<b>BEFORE</b>
Before changes, the download count would wrap to the next line before the icons next to it.  This could be confusing to users who expect to see the number of downloads next to the label.
screenshot 2017-02-08 13 52 57

<hr>
<b>AFTER</b><br>
After changes, the download count remains on the same line as the rest of the text. 
screen shot 2017-03-08 at 12 26 05 pm

When the screen becomes too small to fit everything on one line, the icons are the first to move to the next line underneath the "Download preprint" button.
screen shot 2017-03-08 at 12 27 16 pm

## Notes to QA
Because CSS can potentially act differently across browsers, please be sure to test this on different browsers to make sure that there aren't any abnormalities.

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -113,6 +113,16 @@ ul.comma-list {
     color: black;
 }
 
+#downloadCount {
+    display: inline-block;
+}
+
+@media (max-width: 991px) {
+    .download_project, .social-div {
+        width: auto;
+    }
+}
+
 @media (max-width: 768px) {
     #providerCarousel {
         .carousel-control .icon-next, .carousel-control .icon-prev {
@@ -1195,6 +1205,7 @@ label.title-label {
     float: none !important;
     white-space: nowrap;
 }
+
 .social-div {
     margin-top: 5px;
 }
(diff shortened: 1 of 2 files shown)


app/styles/app.scss
```diff
@@ -113,6 +113,24 @@ ul.comma-list {
     color: black;
 }
 
+div[id='download_project clearfix'] {
```
abought: A few notes on CSS here:

1. You can reference an ID as `#name` which is slightly more idiomatic ([this post](http://stackoverflow.com/questions/9285451/handling-css-id-and-classes-with-spaces) claims the selector-based ID syntax is deprecated in newer browsers, but I can't find the evidence to support this claim)

2. Clearfix is usually a class name (from bootstrap)- it's weird to me that it would be part of the element ID, especially since...

3. ...support for element IDs with spaces is spotty at best, and generally should be avoided.

Out of curiosity, what happens if you rename this element so that clearfix is applied as a class to this parent element? Will the child styles still be necessary then?

brianjgeiger:
@baylee-d Can you fix the merge conflict and move back to passed cr when the tests pass?

PR #280, opened 2017-02-15
[hotfix] [EOSF-514] Use providers list from APIv2 instead of hardcoded list

binoculars:
## Ticket

https://openscience.atlassian.net/browse/EOSF-514

## Purpose

Updates the discover route so that Preprint Providers in from the API are used in the whitelist (not hardcoded)

## Changes

/preprints/discover (OSF Preprints only) should now have all of the providers available. Providers in the API, but not in SHARE will be displayed with with a document count of 0.
screen shot 2017-02-15 at 13 55 35

## Side effects

Providers should now load automatically when added to the API

### DIFF
--- a/app/controllers/discover.js
+++ b/app/controllers/discover.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import config from 'ember-get-config';
+import loadAll from 'ember-osf/utils/load-relationship';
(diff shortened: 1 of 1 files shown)


binoculars:
Thought I was going to use loadAll, but realized I needed findAll instead, so this should be g2g once CI finishes

PR #285, opened 2017-02-24
[EOSF-496] Add version selector for preprint files

binoculars:
❗️ Needs:
- [x] CenterForOpenScience/osf.io#6947
- [x] CenterForOpenScience/ember-osf#192

## Ticket

https://openscience.atlassian.net/browse/EOSF-496

## Purpose

Adds the ability to download previous versions of preprint files

## Changes

* Some clean-up on the supplementary-file-browser component
* Dropdown for previous versions--defaults to latest version, and shows up to the last 10 versions
* Add version to the file-download-path utility function
  * Tests for the utility function

## Notes for QA

You'll need to create a preprint and then go to the project and upload with the same filename (or edit the file) to get more versions. I would recommend attaching other files with a different number of versions to be able to see that the select box is actually updating when you click on a different file. The latest version (highest number) should be at the top of the box, and up to 10 previous versions should be show.

screen shot 2017-03-08 at 17 04 45

When the page initially loads, there won't be any versions shown, then when the API request for the file completes, the current (latest) version will be the only option. Once the API request for all of the versions is complete, up to the last 10 should be shown in the dropdown.

The blue "Download preprint" button in the upper right next to the social media sharing buttons will always download the latest version of the primary preprint file. This remains unchanged.

## Side effects

Due the fact that we're paging the API, only the last 10 versions of a file will be selectable, which I think is appropriate for this situation. The user can always visit the project to > the last 10 versions. For preprints, this will likely be rare, so putting too much time into overcoming this will have diminishing returns.


hmoco:
Reran the tests for this. Functional testing looked good, code looks fine, especially with the addition of tests!

brianjgeiger:
Merge conflict

tests/integration/components/supplementary-file-browser-test.js
```diff
@@ -67,16 +112,16 @@ test('has additional files', function(assert) {
     assert.equal(this.$('#rightArrow').length, 1);
     assert.equal(this.$('#downArrow').length, 1);
 
+    // TODO enable when when tests are fixed
```
hmoco: Quick question and then I will return code reviewing: Can you (@binoculars) provide some clarity as to why a lot of tests here are being skipped? What do you mean by when tests are fixed and why is that not within the scope of this pr?

binoculars: The versions are retrieved with a query, and I can't seem to get it to load properly. Removing the observers also caused other things to break like the filename, so not sure how to handle that right now. I know it works in practice, I don't know how it's supposed to be tested. Would appreciate feedback on how to make it work. It's not out of scope, I just spent too much time rewriting the component, which wasn't my original intention.

hmoco: Makes sense. For suggestions, the only thing I can think of is pinging @jamescdavis, as he was telling me about a `wait` method that could be helpful here. However, I am a little worried about this not being testable and causing other things to break and suggest that maybe this gets looked into/fixed prior to it going to code review and getting merged.

jamescdavis: Yes, `wait` should work here. [This article](https://www.mutuallyhuman.com/blog/2016/01/22/component-integration-testing-in-ember) mentions specifically using `wait` for integration testing of components that have asynchronous operations going on under the hood. I tried it here, but I'm not sure the test is set up correctly otherwise, so asserts were still failing. For an example of using `wait` see: https://github.com/CenterForOpenScience/ember-preprints/commit/28208e449136c19ce7556fc17de97109179b03e1#diff-57c6d61e1c420bd038a684e4cb14f95fR554

laurenbarker:
Closing since this is so stale. Kicking the [Jira ticket](https://openscience.atlassian.net/browse/IN-74) back for triage

PR #288, opened 2017-02-27
[EOSF-238] Change share logo to white for Scielo

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-238

## Purpose
Change share logo to white for Scielo

### DIFF
--- a/app/styles/brands/scielo.scss
+++ b/app/styles/brands/scielo.scss
@@ -8,7 +8,7 @@
     white,
     $logo-dir + 'scielo-banner.png',
     $logo-dir + 'scielo-logo.png',
-    true,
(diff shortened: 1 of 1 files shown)


app/styles/brands/scielo.scss
```diff
@@ -8,7 +8,7 @@
     white,
     $logo-dir + 'scielo-banner.png',
     $logo-dir + 'scielo-logo.png',
-    true,
+    false,
```
brianjgeiger: Yeah, this is not going to be a long term maintenance problem, all these unlabelled descriptors. :) I look forward to putting this into a database.

PR #289, opened 2017-02-28
Clear search box text when navigating away from page [EOSF-494]

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-494 

## Purpose
This is a follow-up to EOSF-426, which clears selected filters when navigating away from the preprints search page. However, it does NOT clear text entered into the search box. 

This ticket is to clear the text entered in the search box.

## Changes
 - Defined a new method within the controller to clear the search box.
 - Call the new method from the existing willTransition method inside discover.js

### DIFF
--- a/app/controllers/discover.js
+++ b/app/controllers/discover.js
@@ -330,6 +330,10 @@ export default Ember.Controller.extend(Analytics, {
         });
     },
 
+    _clearQueryString() {
+        this.set('queryString', '');
+    },
+
     otherProviders: [],
     actions: {
         search(val, event) {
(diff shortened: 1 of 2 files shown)


app/controllers/discover.js
```diff
@@ -383,6 +387,10 @@ export default Ember.Controller.extend(Analytics, {
                 });
         },
 
+        clearQueryString() {
```
binoculars: I don't see where this is being called, it looks like it's not needed. Is this actually needed?

PR #291, opened 2017-03-01
[EOSF-470] Add white lettering when needed for bitss subjects on selection

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-470


app/styles/app.scss
```diff
@@ -204,7 +204,7 @@ $logo-dir: 'img/provider_logos/';
     background-image: url($logo-dir + 'bitss-small.png');
 }
 .provider-bitss-dark {
-    background-image: url($logo-dir + 'bitss-banner.png');
+    background-image: url($logo-dir + 'bitss-dark-small.png');
```
abought: As noted in the companion PR, these images may contain more pixels than are needed for where they will be displayed. (bitss-dark-small in particular is 388kb for a small black and white image) Could I trouble you to downscale this a bit?  The best size is "whatever the page actually expects to show"- see this guidance:
https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization#delivering_scaled_image_assets

If there's a black and white /reduced color palette option, that might also lead to a file size reduction.

hmoco:
@abought, logos much smaller!

abought:
> @abought, logos much smaller!

Hmm. Down from 440 kb to ~20kb total for the two images? Mobile users shall sing songs of these deeds.

If QA decides that the image optimization made it blurry, we can always use less aggressive settings- there's a lot of wiggle room now and this is a big improvement.

abought:
Got nondeterministic CI failures, so I went ahead and restarted the build for you.

PR #293, opened 2017-03-03
[EOSF-594] Add example controller unit tests and component integration tests

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-594

## Purpose
Improve test coverage, with integration and controller tests, to allow simpler future tests writing.

## Changes
Use of ember-osf factories


bower.json
```diff
@@ -12,7 +12,8 @@
     "toastr": "^2.1.2",
     "hint.css": "^2.3.2",
     "jquery.tagsinput": "^1.3.6",
-    "loaders.css": "^0.1.2"
+    "loaders.css": "^0.1.2",
+    "jquery-mockjax": "~2.2.1"
```
abought: `jquery-mockjax` sounds like something used only for testing. Could I suggest moving it to `devDependencies` instead of `dependencies`?

See: http://stackoverflow.com/questions/19339227/bower-and-devdependencies-vs-dependencies

tests/unit/controllers/content-test.js
```diff
@@ -1,12 +1,98 @@
...
+        description = 'string stringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstringstring';
+        notExpanded = 'string ';
+        node.set('description', description);
+        ctrl.notifyPropertyChange('description');
+        assert.equal(ctrl.get('description'), notExpanded.trim());
+
+        //Test less than 350 doesn't use description (uses node.description)
+
+        ctrl.set('expandedAbstract', true);
+        // expandedAbstract logic has been moved to the template, making the expanded tests unnescessary here.
```
abought: Per this note: does this mean that any code after the last assertion can be removed?

tests/helpers/module-for-acceptance.js
```diff
@@ -2,22 +2,45 @@ import { module } from 'qunit';
...
-
-      if (options.beforeEach) {
-        return options.beforeEach.apply(this, arguments);
-      }
+        this.application = startApp();
+        FakeServer.start();
+        manualSetup(this.application.__container__);
+        const url = config.OSF.apiUrl;
+        const provider = FactoryGuy.build('preprint-provider');
+        stubRequest('get', url + '/v2/users/me', (request) => {
```
abought: This looks helpful!

I ran the tests in chrome with the network inspector open, and noticed that the page still sends a request to google analytics every time tests run. This may indicate one more thing that needs stubbing. Thoughts?

binoculars:
Probably should revert changes in controllers, as I've made major updates in #319

jamescdavis:
Indeed. It looks like #319 should probably be merged first? There will likely be several merge conflicts.

app/controllers/submit.js
```diff
@@ -684,7 +684,7 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
                 this.set('basicsLicense', {
                     licenseType: license || this.get('availableLicenses').toArray()[0],
                     year: this.get('model.licenseRecord') ? this.get('model.licenseRecord').year : date.getUTCFullYear().toString(),
-                    copyrightHolders: this.get('model.licenseRecord') ? this.get('model.licenseRecord').copyright_holders.join(',') : ''
+                    copyrightHolders: this.get('model.licenseRecord') ? this.get('model.licenseRecord').copyright_holders.join(', ') : ''
```
jamescdavis: Need to add a test for this.

hmoco:
@brianjgeiger, can't get this to fail locally, can't get it to pass here. Any ideas?

jamescdavis:
merge develop, then it will fail locally

jamescdavis:
Also, I wanted to reference https://github.com/ember-cli/ember-cli-qunit/issues/162, which discusses the issue using wait helpers with qunitjs < 2.3.1 and https://github.com/qunitjs/qunit/pull/1148, which was the fix to qunitjs that allows assertions after [assert.async()](https://api.qunitjs.com/assert/async). Switching to ember-cli-qunit 3 was necessary because ember-cli-qunit 2 depends on qunitjs 1.

PR #274, opened 2017-02-10
Update the language for "sort by options" and "added on" field [EOSF-438]

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-438

## Purpose

The "Date Added" dates are the dates the Preprint was created. The Sorting date, however, is really the Date Modified date. There should be consistency to meet user expectations.

## Changes

- Update the translation file. 
- Use the translation service.
- Rename sort buttons to "Sort by Modified Date (newest to oldest)" and "Sort by Modified Date (oldest to newest)"
- Update the "Added on" field to be "Last edited on"

PR #295, opened 2017-03-06
[EOSF-528] Fixed order on Preprints Search sort

baylee-d:
## Ticket

https://openscience.atlassian.net/browse/EOSF-528

## Purpose

The order of sorting in Preprints would change based on what the user selects.  The order of the list should stay consistent regardless of what is selected.  

## Changes and Screenshots

The list of available sources a user can choose from is static and doesn't change regardless of what's chosen.  

The main style aspect that was changed was to make the font of the dropdown better match the style of other dropdowns on the site (mainly the navbar).  The alignment of the dropdown list was changed to be anchored on the right to match its position on the page.  When you hover over the choices, the background of the list element changes to a slightly lighter color, and the font gets a bit darker.
screen shot 2017-03-30 at 2 50 34 pm

And when you select something from the list, the order of the list will stay the same regardless of what you chose:
screen shot 2017-03-30 at 2 51 02 pm

## Notes to QA
Since this has some CSS changes, please be sure to test this in all browsers to make sure that there aren't any unexpected changes.

### DIFF
--- a/app/controllers/discover.js
+++ b/app/controllers/discover.js
@@ -51,15 +51,15 @@ export default Ember.Controller.extend(Analytics, {
     queryBody: {},
     providersPassed: false,
     pageNumbers: [],
-    staticSortOptions: ['Relevance', 'Upload date (oldest to newest)', 'Upload date (newest to oldest)'],
     sortByOptions: ['Relevance', 'Upload date (oldest to newest)', 'Upload date (newest to oldest)'],
+    selectedSortByOption: '',
 
     treeSubjects: Ember.computed('activeFilters', function() {
         return this.get('activeFilters.subjects').slice();
     }),
-    // chosenOption is always the first element in the list
-    chosenSortByOption: Ember.computed('sortByOptions', function() {
-        return this.get('sortByOptions')[0];
+    // chosenSortByOption is going to be the last selected element, or if it's a new page then it's the first in the list
+    chosenSortByOption: Ember.computed('selectedSortByOption', function() {
+        return this.get('selectedSortByOption') === '' ? this.get('sortByOptions')[0] : this.get('selectedSortByOption');
     }),
 
     showActiveFilters: true, //should always have a provider, don't want to mix osfProviders and non-osf
@@ -294,12 +294,11 @@ export default Ember.Controller.extend(Analytics, {
             });
         }
 
-        const sortByOption = this.get('chosenSortByOption');
         const sort = {};
 
-        if (sortByOption === 'Upload date (oldest to newest)') {
+        if (this.get('selectedSortByOption') === 'Upload date (oldest to newest)') {
             sort.date_updated = 'asc';
-        } else if (sortByOption === 'Upload date (newest to oldest)') {
+        } else if (this.get('selectedSortByOption') === 'Upload date (newest to oldest)') {
             sort.date_updated = 'desc';
         }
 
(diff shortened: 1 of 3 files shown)


binoculars:
#274 (not yet merged) will create conflicts with this PR. Can you merge and resolve conflicts on that? After tests pass, just send back to CR.

sheriefvt:
@baylee-d, @binoculars  please reference the new pull request #305 instead of #274

binoculars:
@baylee-d Can you resolve conflicts please? @sheriefvt closed his original PR and opened a new one that is now merged.

app/styles/app.scss
```diff
@@ -340,6 +340,26 @@ ul.preprints-block-list {
...
+#sortBy {
+    min-width: 291px;
+    width: 100%;
+}
+
+#sortByOptionList {
+    background-color: #EFEFEF;
+    min-width: 291px;
+    .listOption {
+        cursor: pointer;
```
binoculars: The original inline style had `cursor: hand`, do we need that still?

app/controllers/discover.js
```diff
@@ -59,13 +59,14 @@ export default Ember.Controller.extend(Analytics, {
...
 
     treeSubjects: Ember.computed('activeFilters', function() {
         return this.get('activeFilters.subjects').slice();
     }),
-    // chosenOption is always the first element in the list
-    chosenSortByOption: Ember.computed('sortByOptions', function() {
-        return this.get('sortByOptions')[0];
+    // chosenSortByOption is going to be the last selected element, or if it's a new page then it's the first in the list
+    chosenSortByOption: Ember.computed('sortByOption', function() {
+        return this.get('sortByOption') === '' ? this.get('sortByOptions')[0] : this.get('sortByOption');
```
binoculars:
```js
return this.get('sortByOption') || this.get('sortByOptions')[0];
```
Is the same thing, but it requires less computation

app/controllers/discover.js
```diff
@@ -300,13 +301,12 @@ export default Ember.Controller.extend(Analytics, {
             });
         }
 
-        const sortByOption = this.get('chosenSortByOption');
         const sort = {};
-
         const i18n = this.get('i18n');
-        if (sortByOption.toString() === i18n.t('discover.sort_oldest_newest').toString()) {
+
+        if (this.get('sortByOption').toString() === i18n.t('discover.sort_oldest_newest').toString()) {
```
binoculars: whenever you're comparing something, multiple times, assign to a variable first

```js
const sortByOption = this.get('sortByOption').toString();

if (sortByOption === i18n ...) {

} else if (sortByOption === i18n ...) {

}
```

app/styles/app.scss
```diff
@@ -340,6 +340,26 @@ ul.preprints-block-list {
     padding-left: 0;
 }
 
+#sortBy {
```
binoculars: Why 291px? Pixels should be the last resort. It doesn't quite look right on the iPhone 5 setting:

screen shot 2017-03-27 at 15 29 16

app/styles/app.scss
```diff
@@ -340,6 +340,26 @@ ul.preprints-block-list {
...
+    min-width: 291px;
+    width: 100%;
+}
+
+#sortByOptionList {
+    background-color: #EFEFEF;
```
binoculars: I'm not sure if this is the only place where we have a dropdown on preprints, but it's worth looking into for consistency. In #285 I added one for selecting preprint file versions, so it should be consistent. I'd prefer to modify the global dropdown styling, instead of having one-off styling just for this particular dropdown.

app/controllers/discover.js
```diff
@@ -303,10 +303,11 @@ export default Ember.Controller.extend(Analytics, {
 
         const sort = {};
         const i18n = this.get('i18n');
+        const sortByOption = this.get('sortByOption').toString();
 
-        if (this.get('sortByOption').toString() === i18n.t('discover.sort_oldest_newest').toString()) {
+        if (sortByOption.toString() === i18n.t('discover.sort_oldest_newest').toString()) {
```
binoculars: You can get rid of the `.toString()` here and on L310, since you already have it as a string from L306.

PR #296, opened 2017-03-06
[EOSF-531] Fix overlap on content view

binoculars:
## Ticket

https://openscience.atlassian.net/browse/EOSF-531

## Purpose

Content on the preprints content view overflows into the citations section when the browser window width is between 990px and 1200px.

## Changes

Updates the bootstrap classes and stylesheet

### Before
screen shot 2017-03-06 at 15 18 01

### After
screen shot 2017-03-06 at 15 18 46

## Side effects

N/A

### DIFF
--- a/app/templates/components/supplementary-file-browser.hbs
+++ b/app/templates/components/supplementary-file-browser.hbs
@@ -3,10 +3,10 @@
 {{/file-renderer}}
 
 <div class="row">
-    <div class="col-sm-8">
+    <div class="col-sm-6">
         <p class="f-w-lg p-b-md">{{selectedFile.name}}</p>
     </div>
-    <div class="col-sm-4 text-nowrap supplemental-downloads">
+    <div class="col-sm-6 text-nowrap supplemental-downloads">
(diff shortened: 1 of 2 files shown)


app/templates/components/supplementary-file-browser.hbs
```diff
@@ -3,10 +3,10 @@
 {{/file-renderer}}
 
 <div class="row">
-    <div class="col-sm-8">
+    <div class="col-sm-6">
         <p class="f-w-lg p-b-md">{{selectedFile.name}}</p>
     </div>
-    <div class="col-sm-4 text-nowrap supplemental-downloads">
+    <div class="col-sm-6 text-nowrap supplemental-downloads">
```
abought: Just as a small note: if we want to preserve the large-device layout while *also* accommodating smaller screens, I believe bootstrap provides a way to specify different column layouts simultaneously. Eg:
https://getbootstrap.com/examples/grid/

(I'll functionally test this next. Unfortunately docker-sync has gone rogue at the moment)

PR #297, opened 2017-03-08
Fix Preprints submit form, create component, text [EOSF-547]

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-547 

## Purpose
One or two letters on the right and the left of the paragraph are cut off. Doesn't matter what browser window width you are viewing with.
There is a need to a fix the text.

## Changes

Changed text style

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -1253,6 +1253,11 @@ nav.branded-navbar {
     line-height: 28px;
   }
 
+.finalize-upload-language {
+    padding:  0 15px;
+    align-content: center;
+}
+
 .branded-footer-links {
     margin-bottom: 1em;
     line-height: 28px;
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -695,6 +695,7 @@ hr {
         }
         &.cp-is-open {
             opacity: 1;
+            padding: 15px;
         }
     }
     // THESE STYLES ARE INTERFERING WITH MY STYLES.
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -1253,11 +1253,6 @@ nav.branded-navbar {
     line-height: 28px;
   }
 
-.finalize-upload-language {
-    padding:  0 15px;
-    align-content: center;
-}
-
 .branded-footer-links {
     margin-bottom: 1em;
     line-height: 28px;
--- a/app/templates/components/file-uploader.hbs
+++ b/app/templates/components/file-uploader.hbs
@@ -88,7 +88,7 @@
                         </form>
 
                         <div class="row text-center m-v-md">
-                            <p> {{t "components.preprint-form-project-select.initiate_preprint_process"}} </p>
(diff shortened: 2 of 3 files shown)


app/templates/components/file-uploader.hbs
```diff
@@ -88,7 +88,7 @@
                         </form>
 
                         <div class="row text-center m-v-md">
-                            <p> {{t "components.preprint-form-project-select.initiate_preprint_process"}} </p>
+                            <p class="finalize-upload-language"> {{t "components.preprint-form-project-select.initiate_preprint_process"}} </p>
```
abought: This seems to fix it- but I'm a little surprised that text was ever falling outside the container at all.

Two notes:
1. I think this might be worth a smidge further investigating as to the root cause; otherwise we're likely to keep one-off fixing small instances.

2. Currently, this repo imports osf styles, which define margin and padding helper classes. For example, the div on line 90 uses `m-v-md` (15 px vertical margins). Consider using an applicable pre-defined style helper rather than adding your own, where possible.
https://centerforopenscience.github.io/osf-style/style.html#layout-helper

sheriefvt:
Thanks for your comments 👍 .

abought:
Just to follow up- what was the root cause of the problem? Why was the text getting clipped at all?

sheriefvt:
I am not sure about the cause for this problem. It might be of using float somewhere or cp-Panel has overflow set to hidden in the app.scss. It might be that collapse/ expand the panel remove or cancel the parent container style settings. This is what possible causes I collected from investigation yesterday.

abought:
> I am not sure about the cause for this problem. It might be of using float somewhere or cp-Panel has overflow set to hidden in the app.scss. It might be that collapse/ expand the panel remove or cancel the parent container style settings. This is what possible causes I collected from investigation yesterday.

Thanks for investigating a few possible causes. Was there any evidence to support one particular possibility, out of those you listed?

Basically, if fixing the root cause would be a really big change, then this is fine to merge as is. But if we think that a small targeted change could let us fix the behavior (without adding a lot of styles), then the simpler solution is always preferred.

So: based on your investigation, do we currently know enough about the cause to believe that a simpler solution is feasible (<2 hrs work)?

sheriefvt:
I tried changing the style of cp-panel body and added padding. Without padding and using overflow hidden might be the cause.

abought:
> I tried changing the style of cp-panel body and added padding. Without padding and using overflow hidden might be the cause.

I'm not sure that this change is better or worse than the previous version per se, but it appears to work. Moving it into the collapsible panel style will at least help all current and future uses of this widget, rather than adding one-off style fixes to every usage.

For future reference, note that `padding` sets padding on all four sides, not just left or right.

app/styles/app.scss
```diff
@@ -695,6 +695,7 @@ hr {
         }
         &.cp-is-open {
             opacity: 1;
+            padding: 15px;
```
binoculars: `margin: 1rem 0;`

Look at the other element's margins

PR #299, opened 2017-03-09
[EOSF-498] Dockerfile adds ember-osf destination folder

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-498

## Purpose
To create a destination for data volume containing local version of ember-osf, for ember-osf/ember-preprints development.

## Changes
Adds /ember-osf folder to root of preprints container.

## Side effects
If not using it for local ember-osf development, creates an empty folder

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,7 @@ RUN cd /tmp \
     && mv watchman /usr/local/bin/watchman \
     && rm -Rf /tmp/watchman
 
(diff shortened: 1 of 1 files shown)


binoculars:
Can you merge hmoco/osf.io#11 for this PR?

binoculars:
We don't actually need to modify the Dockerfile here. That directory will get created when it's mounted in docker-compose. Closing.

PR #300, opened 2017-03-13
[Hotfix] [EOSF-555] Prerender full abstract

binoculars:
## Ticket

https://openscience.atlassian.net/browse/EOSF-555

## Purpose

Needed for Google Scholar to index the page properly

## Changes

* Sets `expandedAbstract` to true by default if the user agent contains `Prerender` (per the [Prerender docs](https://prerender.io/documentation/faq))
* Simplifies the description property
* Use CSS for the ellipses
* Move strings to translation file
* Hides the show more/show less button if the the description the shortened description is less than 350 characters

## Side effects

N/A

### DIFF
--- a/app/controllers/content.js
+++ b/app/controllers/content.js
@@ -56,7 +56,7 @@ export default Ember.Controller.extend(Analytics, {
     expandedAuthors: true,
     showLicenseText: false,
     fileDownloadURL: '',
-    expandedAbstract: false,
+    expandedAbstract: navigator.userAgent.includes('Prerender'),
     isAdmin: Ember.computed('node', function() {
         // True if the current user has admin permissions for the node that contains the preprint
         return (this.get('node.currentUserPermissions') || []).includes(permissions.ADMIN);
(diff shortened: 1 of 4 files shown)


app/controllers/content.js
```diff
@@ -56,7 +56,7 @@ export default Ember.Controller.extend(Analytics, {
     expandedAuthors: true,
     showLicenseText: false,
     fileDownloadURL: '',
-    expandedAbstract: false,
+    expandedAbstract: navigator.userAgent.includes('Prerender'),
```
abought: In ember, controllers are singletons. If this property gets changed once on any preprint page, there's a good chance that this value would be remembered, and all future preprints you visit (until next reload) would obey the same setting.

Based on the assumption that the prerendered page won't go around clicking any buttons on the page, this shouldn't affect crawlers, so I'd say this is not a blocker for this ticket. But it may be a small user-facing bug. (confirmed on staging)

For future reference, to control initial state every time the page loads, methods like `setupController` or [resetController](http://emberjs.com/api/classes/Ember.Route.html#method_resetController) can be helpful.

abought: For the future, we may want to think about doing other things (like meta tags) conditionally on prerender as well, and centralizing the logic. But this is a good start.

app/templates/content.hbs
```diff
@@ -83,12 +83,8 @@
...
                     <div class="p-t-xs">
                         <h4 class="p-v-md f-w-md"><strong>{{t "global.abstract"}}</strong></h4>
-                        {{#if useShortenedDescription}}
-                            <p class="abstract">{{description}}</p>
-                            <button class='btn-link' {{action 'expandAbstract'}}>{{if expandedAbstract 'See less' 'See more'}}</button>
-                        {{else}}
-                            <p class="abstract">{{node.description}}</p>
-                        {{/if}}
+                        <p class="abstract {{unless expandedAbstract 'abstract-truncated'}}">{{if expandedAbstract node.description description}}</p>
+                        <button class='btn-link' hidden={{not hasShortenedDescription}} {{action 'expandAbstract'}}>{{t (if expandedAbstract 'content.see_less' 'content.see_more')}}</button>
```
abought: Something we may want to watch for: will this button text become part of the indexed abstracts for our preprints?

app/controllers/content.js
binoculars: Inital state for prerender is all that matters, because there won't be js that the bot can execute

app/templates/content.hbs
binoculars: No idea, the ticket only specified the paragraph tag.

PR #301, opened 2017-03-14
[EOSF-549] Add query param for selected file on content page

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-549

## Purpose
Allows passing in query param on content page that pre-selects a file to display if the id passed matches that of a file in the preprint


binoculars:
I didn't think to test this, but when editing a preprint, it fails. I think the issue the query parameters from the controller are somehow interfering with the route.

PR #269, opened 2017-02-06
Feature/Keen Tracking Preprint Views and Downloads [EOSF-347]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-347

❗️ Now blocked by discover-page component PR https://github.com/CenterForOpenScience/ember-osf/pull/169
❗️ For Devops - need to setup system variables for Keen private and public projects, process.env.PREPRINTS_PRIVATE_PROJECT_ID, process.env.PREPRINTS_PRIVATE_WRITE_KEY, process.env.PREPRINTS_PUBLIC_PROJECT_ID, process.env.PREPRINTS_PUBLIC_WRITE_KEY
❗️ TODO - REMOVE PINNED EMBER-OSF COMMIT ONCE DISCOVER-PAGE COMPONENT MERGED.
❗️ Download Preprint button not working (issue outside of this PR, seems to be switch to fileDownloadURL.content)
## Purpose

Tracks pageviews of preprints in Keen.  Tracks download counts of preprint file and supplementary files for noncontributors only. Tracks views of preprint file and supplementary files for all users.  Because we're using `ember-metrics`, and ember-osf now has an ember-metrics Keen adapter, events that were being sent just to GA, should start automatically being sent to Keen as well.

## Changes
When primary or supplementary file downloaded, these are logged under the **non-contributor-preprint-downloads** collection.  Sent to both public and private preprints projects, so can be retrieved for later display. File downloads are tracked for non-contribs only. Selected data:
```
 "download_info": {
        "preprint": {
            "type": "preprint",
            "id": "fcrax"
        },
        "file": {
            "primaryFile": true,
            "id": "58763e8963ca2f4790fa8336",
            "version": 1
        }
    },
    "interaction": {
        "action": "click",
        "category": "link",
        "url": "http://localhost:7777/v1/resources/t5hyf/providers/osfstorage/58763e8963ca2f4790fa8336",
        "label": "Content - Download as Non-Contributor"
    },
```
When supplementary or primary preprint files are viewed, these are logged under the **preprint-file-views** collection.   When the content page is loaded along with the primary file that counts as a view of the primary file.  When the user toggles through supplementary files, then those are logged as views of supplementary files. Sent to both public and private preprints projects, so can be retrieved for later display. Selected data: 
```
 "file_views": {
        "preprint": {
            "type": "preprint",
            "id": "fcrax"
        },
        "file": {
            "primaryFile": true,
            "id": "58763e8963ca2f4790fa8336",
            "version": 1
[...]

PR #302, opened 2017-03-15
Feature/Build Preprints Discover Page with Ember-OSF widget [EOSF-558]

pattisdr:
## Ticket

https://openscience.atlassian.net/browse/EOSF-558
❗️ For Devops - need to setup system variables for Keen private and public projects, as well as a google analytics project for preprints: process.env.PREPRINTS_GOOGLE_ANALYTICS_ID, process.env.PREPRINTS_PRIVATE_PROJECT_ID, process.env.PREPRINTS_PRIVATE_WRITE_KEY, process.env.PREPRINTS_PUBLIC_PROJECT_ID, process.env.PREPRINTS_PUBLIC_WRITE_KEY
❗️ TODO - REMOVE PINNED EMBER-OSF COMMIT ONCE DISCOVER-PAGE COMPONENT MERGED.
❗️ Depends on https://github.com/CenterForOpenScience/ember-osf/pull/205
## Purpose

Builds the entirety of the preprints discover page with the Ember-OSF discover-page widget.  A lot of code has been moved to Ember-OSF - mostly components and helpers that can be shared with registries.  A lot of code from the discover controller has been added to the discover-page widget in Ember-OSF.  Accommodates themed pages.  Facets specific to preprints have been componentized and are passed into the discover-page widget (search-facet-provider and search-facet-taxonomy). Analytics is now more generic.

## Changes

The discover page is now built like this:
```
{{discover-page
    activeFilters=activeFilters
    clearFiltersButton=clearFiltersButton
    consumingService=consumingService
    detailRoute=detailRoute
    discoverHeader=discoverHeader
    facets=facets
    fetchedProviders=model
    filterMap=filterMap
    filterReplace=filterReplace
    lockedParams=lockedParams
    page=page
    provider=provider
    q=q
    queryParams=queryParams
    searchButton=searchButton
    searchPlaceholder=searchPlaceholder
    showActiveFilters=showActiveFilters
    sortOptions=sortOptions
    subject=subject
}}
```
## For QA - noticeable changes. There shouldn't be many differences, main change is that page is being built with a component (that also builds registries discover). Needs thorough testing of Discover Page as well as checking navigation from Discover Page to other pages.
- Analytics labels are now more generic - previously an event might be sent to look like this:
```
{
   action:"click",
   category: "link",
   label: "Preprints - Index - Add preprint"
}
```
Now we remove "Preprints" from the label:
```
{
   action:"click",
   category: "link",
   label: "Index - Add preprint"
}
```
The reason is that some analytics are being fired from Ember-OSF now.  Instead of having the label designate which service we are using, we are sending preprints to preprints buckets and registries to registries buckets.
## Other changes
- Query parameters combined with OR instead of AND (since this is how we're actually searching)
- Lucene search turned back on.
- Help button added on search bar for Lucene search
-  If invalid search query like (a:), more lucene search help appears (like SHARE)
- Number of searchable preprints added under search bar
- Sort by button formatting has changed and now says Modified date
- Whitelisted providers
- if no query params, there are no empty query params in URL.
- better margins on search results in xs screens


pattisdr:
Hm, these tests pass locally.

pattisdr:
Oops, I didn't have global gitignores setup on my new machine and missed that.  Thanks - my intent was to remove more code here than I added!

tests/integration/components/search-facet-taxonomy-test.js
```diff
@@ -0,0 +1,29 @@
...
+
+    this.render(hbs`{{search-facet-provider
+        key=key
+        options=facet
+        updateFilters=(action noop)
+        activeFilters=activeFilters
+        filterReplace=filterReplace
+    }}`);
+
+  assert.equal(this.$().text().trim(), '');
```
binoculars: Same as before, need some decent tests

app/components/search-facet-provider.js
```diff
@@ -0,0 +1,114 @@
...
+ * }}
+ * ```
+ * @class search-facet-provider
+ */
+export default Ember.Component.extend(Analytics, {
+    store: Ember.inject.service(),
+    theme: Ember.inject.service(),
+    otherProviders: [],
+    searchUrl: config.OSF.shareSearchUrl,
+    whiteListedProviders: [
```
binoculars: This would be a good opportunity to move these to config/environment

package.json
```diff
@@ -31,6 +31,7 @@
     "ember-ajax": "2.0.1",
     "ember-autosave": "2.0.1",
     "ember-bootstrap": "0.7.2",
+    "ember-browserify": "^1.1.13",
```
pattisdr: We've discussed this [before](https://github.com/CenterForOpenScience/ember-preprints/pull/269#issuecomment-28444624), it seems like ember-browserify is needed to use some Ember-OSF dependencies. It doesn't matter that ember-browserify is installed in ember-osf; you need it in the consuming application.  I think it's a bug with [ember-browserify and addons](https://github.com/ef4/ember-browserify#using-ember-browserify-in-addons)

binoculars: So this is actually a little different. You should move ember-browserify to the dependencies section (instead of devDependencies) of ember-osf. That way it *should* get installed when your run yarn and don't have to explicitly include it in consuming apps that rely on ember-osf, unless they actually need it in that the consuming app itself, which in this case we don't.

pattisdr: i will try that, thanks

pattisdr: Preprints does not seem to be building without ember-browserify now if pointed towards the Ember-OSF PR for the latest discover page changes https://github.com/CenterForOpenScience/ember-osf/pull/205

PR #307, opened 2017-03-22
[EOSF-545] re-add disciplines

binoculars:
## Ticket

https://openscience.atlassian.net/browse/EOSF-545

## Purpose

When trying to add a discipline back after removing it, the current item will remain selected, so it won't be able to be clicked and added again.

## Changes

* Major refactor of the `subject-picker` component
* Simplify the code by using a single object class for the columns.
* Clear out all of the selections (highlighted subjects)
* Clear all but the first column

## Notes for reviewers

* This should work on both the submit page and edit page. The edit page is currently broken (develop), so you'll have to comment out the query parameters in the content controller to get it to load.
* Try adding multiple disciplines with 2 and 3 levels and removing and adding them again in different orders. There are many edge cases around this, and I believe that I've accounted for them, but it's difficult to know for sure.


abought:
When I deselect a subject, the third arrow disappears, but the UI in the third column doesn't update. Is that the intended behavior?

screen shot 2017-04-07 at 5 19 46 pm

app/components/subject-picker/component.js
```diff
@@ -8,6 +8,25 @@ function arrayEquals(arr1, arr2) {
 function arrayStartsWith(arr, prefix) {
     return prefix.reduce((acc, val, i) => acc && val && arr[i] && val.id === arr[i].id, true);
 }
+
+const Column = Ember.Object.extend({
+    sortDefinition: ['text:asc'],
+    filterText: '',
+    selection: null,
+    subjects: [],
+    subjectsFiltered: Ember.computed('subjects.[]', 'filterText', function() {
```
binoculars: It's simple add or remove, so we do want `[]` instead of `@each`

app/components/subject-picker/template.hbs
```diff
@@ -1,47 +1,34 @@
...
     {{/each}}
 </div>
 
 <div class="row">
+    {{#each columns as |column tier|}}
+        <div class="col-md-4">
+            <ul style="overflow: scroll">
+                {{input value=column.filterText class="form-control" placeholder="Search"}}
+                {{#each column.subjectsSorted as |subject|}}
+                    <li {{action 'select' subject tier}} class={{if (eq subject column.selection) 'selected'}}>
```
binoculars: Going with [listbox](https://www.w3.org/TR/wai-aria/roles#listbox) & listitem, since they seem like the best fit

binoculars:
Not sure how that is occurring--I have it set so that when a subject is deselected, all of the current selections in the listbox are removed, then columns 2 and 3 are cleared. Might be a sync / refresh issue.

Latest merge shows some weirdness though:
screen shot 2017-04-10 at 10 40 34

/cc @baylee-d

abought:
> Latest merge shows some weirdness though:

Confirming weirdness in current behavior. Whether we target a fix for this PR or not, I think this is a good question to answer before merging.
screen shot 2017-04-20 at 11 31 56 am

app/components/subject-picker/component.js
```diff
@@ -31,53 +50,11 @@ export default Ember.Component.extend(Analytics, {
...
-        }
-        return items;
-    }),
-    tier3Sorted: Ember.computed.sort('tier3Filtered', 'tierSorting'),
-
-    // Currently selected subjects
-    selection1: null,
-    selection2: null,
-    selection3: null,
+    initialSubjects: () => [],
```
abought: Per flowdock discussion, this might not work as expected.

abought:
> Confirming weirdness in current behavior. Whether we target a fix for this PR or not, I think this is a good question to answer before merging.

Changes look good, but the issue raised by Barrett and myself above is worth at least a glance before this gets merged.

brianjgeiger:
Merge conflicts

app/controllers/submit.js
```diff
@@ -831,31 +804,49 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
...
+            //         .then(() => {
+            //             this.send('next', this.get('_names.1'));
+            //         })
+            //         .catch(() => {
+            //             model.set('subjects', currentSubjects);
+            //             this.get('toast').error(this.get('i18n').t('submit.disciplines_error'));
+            //         });
+            // } else {
+            //     this.send('next', this.get('_names.1'));
+            // }
```
jamescdavis: Can you remove this commented out code?

PR #311, opened 2017-03-29
[EOSF-585] Hotfix/maintentance banner

binoculars:
## Ticket

https://openscience.atlassian.net/browse/EOSF-585

## Purpose

Add maintenance banner to preprints

## Changes

Added a component. This is temporary and will be moved into ember-osf.

### DIFF
--- a/app/components/maintenance-banner/component.js
+++ b/app/components/maintenance-banner/component.js
@@ -1,0 +1,48 @@
+import Ember from 'ember';
+import config from 'ember-get-config';
+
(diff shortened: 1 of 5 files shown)


brianjgeiger:
@binoculars Since this may be a while for tests, can you confirm that all the unit tests and lints are good locally?

binoculars:
Yes, I can confirm tests are working for both PRs. Didn't include the integration tests for the component due to the time constraint, but will follow-up tests once we move it to ember-osf.

PR #312, opened 2017-03-29
Subject tooltip for top-level disciplines by preprint providers (1) [EOSF-491]

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-491 

## Purpose

When hovering over top-level discipline in the index page, a tooltip shows an example of subdisciplines. Currently, the subdisciplines are hard-coded, causing the display of one or more subjects that are not used (allowed) by the provider. 

This PR provides solution#1. Solution#2 is presented in PR #326.
This solution use tooltips that show children subjects chosen from the subset that the provider allows, and the information come from the server API as separate API request per each subject. No changes made to OSF API or ember-OSF.

## Changes

Remove the hard-coded subjects and replace with actual children subjects coming from the API.

## Side effects

For Solution#1
The API calls needed to retrieve the subdisciplines might add burden to the server side. The QA team need to conduct stress/load testing to evaluate the side effects. Another solution is provided through updating the server side API and ember-OSF in PR #326.


app/components/taxonomy-top-list.js
```diff
@@ -18,18 +19,35 @@ import Analytics from '../mixins/analytics';
...
+                        parents
+                    },
+                    page: {
+                        size: 4  //number of subjects to show-up in each tooltip.
+                    }
+                })
+            .then(results =>
+                results.map(result => result.get('text'))
+            )
+            .then(results =>
```
binoculars: No need to put another `.then` here. You can chain the result from `results.map`. A `.then` is only necessary if the previous statement returns a promise.

tests/integration/components/taxonomy-top-list-test.js
```diff
@@ -1,18 +0,0 @@
-import { moduleForComponent, test } from 'ember-qunit';
```
binoculars: Need to write tests for this--especially with this level of complexity.

app/components/taxonomy-top-list.js
```diff
@@ -18,18 +19,35 @@ import Analytics from '../mixins/analytics';
...
+                results.toArray().sort().join(', ')
+            );
+    },
+    getAllChildren(sortedList, i, provider) {
+        if (i >= 0) {
+            const subjectId = sortedList.objectAt(i).get('id');
+            const subjectChildren = this.getSubjectChildren(subjectId, provider);
+            subjectChildren.then(result => {
+                Ember.$('#' + subjectId).attr('data-hint', result); Ember.$('#' + subjectId).attr('aria-label', result);
```
binoculars: Don't do DOM manipulation. Set a property on the component and let Handlebars do the rest.

app/components/taxonomy-top-list.js
```diff
@@ -18,18 +19,35 @@ import Analytics from '../mixins/analytics';
...
+            }).then(() => this.getAllChildren(sortedList, i - 1, provider));
+        }
+    },
+    didInsertElement() {
+        this._super(...arguments);
+        const sortedList = this.get('list').sortBy('text');
+        this.get('theme.provider').then(provider => this.getAllChildren(sortedList, sortedList.get('length') - 1, provider));
```
binoculars: You should be using `.map` here instead of a recursive function

laurenbarker:
Closing since this improvement was resolved by removing tooltips on subjects.

PR #240, opened 2016-12-18
Spanish translation.

DrDub:
## Purpose

This is a Spanish translation. I dedicate it to public domain, use it as
you see fit.

## Changes

Spanish locale files

## Side effects

If the browser is configured to Spanish, the texts should be in Spanish.

PR #324, opened 2017-04-12
[EOSF-488] Add support for Spanish translation

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-488

## Purpose
Continuation of https://github.com/CenterForOpenScience/ember-preprints/pull/240. With the addition of a test for this and future translations. 

## Side Effects
New items included in translation.js in english will need at least a placeholder translation on every other language.


binoculars:
Muchas gracias. ¡Esto es increíble!

screen shot 2017-04-13 at 14 43 25

💯

hmoco:
@brianjgeiger, if we do want this feature, the sooner it is merged the better, as tests will begin failing on develop when we merge any PR pointed to develop which includes new terms. This also means that this will require continual upkeep if it doesn't get merged.

brianjgeiger:
@hmoco Only one thing missing. Let's get that fixed and then get this merged in.

jamescdavis:
I went through these kind of quickly and most look pretty good. There are couple that might not have the meaning exactly right and a couple of places where verbs were used when nouns should have been, but I'd say good enough for now.

PR #326, opened 2017-04-13
Subject tooltip for top-level disciplines by preprint providers (2) [EOSF-491] 

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-491 

## Purpose
When hovering over top-level discipline in the index page, a tooltip shows an example of subdisciplines. Currently, the subdisciplines are hard-coded, causing the display of one or more subjects that are not used (allowed) by the provider. 

This PR provides solution#2. Solution#1 is presented in PR #312.
This solution use tooltips that show children subjects chosen from the subset that the provider allows, and the children information come from the server API along with the parent in the same request.

## Changes

- Update the OSF API in separate pull request
- Update ember osf in separate pull request
- Add new helper for post-processing subject children
- Add test for the new helper
- Update the taxonomy-top-list component template
- Update the taxonomy-top-list test


laurenbarker:
Closing since this improvement was resolved by removing tooltips on subjects.

PR #332, opened 2017-04-21
Feature/Font-Awesome in Safari [EOSF-548]

pattisdr:
## Ticket

https://openscience.atlassian.net/browse/EOSF-548

## Purpose

When you clear your cache, font awesome fonts aren't loading on first-load in _Safari_. Subsequent loads are fine.

## Changes

Add minified Font-Awesome CDN to head tag of index.html.  There might be better ways to do this, but this seems to ensure that the icons are available on first load.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -21,7 +21,7 @@
             skipStartupTypeset: true
         });
     </script>
-    <link href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.6.3/css/font-awesome.css" rel="stylesheet" type="text/css">
+    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" type="text/css">
   </head>
   <body>
     {{content-for "body"}}
--- a/app/index.html
+++ b/app/index.html
@@ -21,6 +21,7 @@
             skipStartupTypeset: true
         });
     </script>
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.6.3/css/font-awesome.css" rel="stylesheet" type="text/css">
   </head>
   <body>
     {{content-for "body"}}


abought:
It appears that font-awesome is included in other ways. Does this mean the user is downloading it 2+ times?

Also: do the icons *never* load, or is it just slow? We could, eg, make the icons load first using `prepend` [when we import](https://ember-cli.com/api/classes/EmberApp.html#method_import) it. In theory an [addon](https://github.com/martndemus/ember-font-awesome) should be making sure the assets are part of the build already, so this is admittedly quite odd. If we go the CDN route I might suggest telling the addon to exclude some assets as appropriate.

https://github.com/abought/ember-preprints/blob/88a2893c2c1a5b995bf35ba961a94bfbafa69996/bower.json#L8-L8
https://github.com/abought/ember-preprints/blob/723206e519f0d56ea37ded238e9bb2a7b0d833ba/package.json#L60-L60

pattisdr:
@abought, it looks like the icons never load on first load in Safari.

pattisdr:
To address your comments @abought:

> It appears that font-awesome is included in other ways. Does this mean the user is downloading it 2+ times?

It looks like we are using the ember-font-awesome addon and then font-awesome is in bower.json.  However, it doesn't seem like we're importing font-awesome in ember-cli-build.  So I think font-awesome is being downloaded 2X with 1) ember-font-awesome and 2) the CDN



> Also: do the icons never load, or is it just slow? 

The icons are never loading on first load in Safari.



>  I might suggest telling the addon to exclude some assets as appropriate.

I attempted by putting these lines in ember-cli-build.  This also kept the icons from loading on first load.
```
'ember-font-awesome': {
        includeFontAwesomeAssets: false
},
```


> consider using the minified version

Good idea.


From what it looked like to me, having both the CDN and downloading font-awesome as a part of ember-font-awesome ensured the icon was there on first load.

PR #333, opened 2017-04-24
[IN-76][Preprints] Choose Preprint Provider on Create Preprint

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/IN-76

## Purpose

Add a way to select the preprint provider (aka server) when submitting a preprint from unbranded OSF preprints.

## Changes

* Extended the provider carousel component to add a "selectable" mode that uses provider logos from the API and whereby clicking on a logo will select it and trigger a controller action that is passed in as a parameter to the component.
* Modified the submit controller to use a currentProvider attribute to track the current provider. This is initialized from the provider attached to the theme.
* Created an action in the submit controller that sets the current provider and clears the list of selected subjects.
* Added the extended provider carousel to the submit interface only for the unbranded theme and only when not in edit mode. Set the action to be the aforementioned controller action.
* Modified the subject picker component to:
  * expect the current provider to be passed in as a parameter rather than querying the theme.
  * reset the list of available subjects when the parameters change.
* Passed in the current provider to the subject picker in the submit interface.
* Modified/added verbiage to the english locale translations and added template logic to display the appropriate verbiage depending on whether we are branded or unbranded.

## Side effects

The provider carousel is also used on index and discover. Care was taken to not have any side effects on these interfaces, but this should be well tested to confirm.


app/templates/submit.hbs
```diff
@@ -14,9 +14,22 @@
     <div class="container">
         <div class="row m-t-lg">
             <div class="col-xs-12 col-md-10 col-md-offset-1 ">
-                <p>{{if editMode (t "submit.body.p_edit") (t "submit.body.p_add" name=(if theme.isProvider theme.provider.name (t "global.brand_name")))}}</p>
+                <p>{{if editMode (t "submit.body.p_edit") (t (if theme.isProvider "submit.body.p_add_provider" "submit.body.p_add") name=(if theme.isProvider theme.provider.name (t "global.brand_name")))}}</p>
                 {{#cp-panels accordion=true}}
-                    {{#with _names.[0] as |name|}}
+                    {{#unless editMode}}
```
abought: Drive-by comment: you may be able to consolidate some of these nested conditionals using the syntax enabled by [ember-truth-helpers](https://github.com/jmurphyau/ember-truth-helpers).

jamescdavis: Thanks @abought! I started using truth helpers elsewhere and forgot to go back and refactor this.

jamescdavis:
Closing in favor of #533

PR #335, opened 2017-04-24
Preprints Branded Not Found Page Broken Link [EOSF-621]

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-621 

## Purpose
When navigating to a branded detail page that 404s, such as https://osf.io/preprints/socarxiv/dfjghdjshgjkf ) and by inspecting the "Go to **[provider]** Preprints" link, it shows https://osf.io/preprints/undefined

## Changes
Update the error-page component and use a link with href that points to the url value [pathPrefix] extracted from the theme service.

### screen shots after the changes
image

image


hmoco:
Could you explain why `theme.pathPrefix` is defined when `theme.id` is not? Or is this problem an inherit problem of the slug link-to thing we have going on?

sheriefvt:
Thanks @hmoco for pointing this out. I tried several trials using link-to helper and it was broken. I don't know the cause of the problem. I decided to folow this approach as a way to avoid breaking other pieces of code. I used pathPrefix as it contains the id with "/" proprely inserted.

hmoco:
Approved (since I can't find the code review controls :( )

PR #336, opened 2017-04-24
Update PsyArXiv branded footer color [EOSF-620] 

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-620 

## Purpose
Recently, PsyArXiv changed their colors from #ff383f to #CF1D35.
The buttons and logo colors were changed, but the footer did not get updated to their new color.

## Changes
Update the footer color to #CF1D35 in psyarxiv.scss

### before 
image

### after 
image

### DIFF
--- a/app/styles/brands/psyarxiv.scss
+++ b/app/styles/brands/psyarxiv.scss
@@ -25,5 +25,5 @@
 
 .branded-footer a,
 #view-page .osf-box-lt .fa {
-    color: #ff383f;
+    color: #CF1D35;
 }


app/styles/brands/psyarxiv.scss
```diff
@@ -25,5 +25,5 @@
 
 .branded-footer a,
 #view-page .osf-box-lt .fa {
-    color: #ff383f;
+    color: #CF1D35;
```
abought: The JIRA ticket describes the situation as follows:
> The buttons and logo colors were changed, but the footer did not get updated to their new color.

This sounds like a great use case for [SASS variables](http://sass-lang.com/guide#topic-2), which would let us define the color in just one place that could be more easily maintained in the future.

brianjgeiger: Generally I would say yes, but we're probably going to end up pulling all provider-specific CSS into a separate CSS file that's loaded onto a CDN and addressed using a scheme rather than relying on it being pre-processable, so that change would probably end up being discarded before long.

abought: Good to know! Going with a CDN will really simplify some of our code in the future.

Not a blocker for this PR, but speaking in general terms: we probably could still use SASS in provider styles if we wanted to. It should be possible to build SASS separate from ember, as part of the "deploy new styles to production" script.

PR #338, opened 2017-04-25
Handle preprint providers' pages not ready for production [EOSF-619]

sheriefvt:
## Ticket

https://openscience.atlassian.net/browse/EOSF-619 

## Purpose

Preprint providers pages that are not currently active in Production (or incomplete). Attempting to visit their page should lead to a "Page not found." Instead, it leads to error and a completely blank page.  We have in this case sceilo and lawarxiv.

## Changes
Update the preprints config and add flag "ready" that indicate whether the preprint provider is ready on production or not. Handle the unready providers and display "page not found".

### DIFF
--- a/app/routes/provider.js
+++ b/app/routes/provider.js
@@ -22,10 +22,7 @@ export default Ember.Route.extend({
 
         if (providerIds.includes(slugLower)) {
             const {domain} = providers.find(provider => provider.id === slugLower) || {};
-            const isReady = providers.find(function (p) {
-                return p.id === slug;
-            }).ready;
-
+            const isReady = providers.find(provider => provider.id === slug).ready;
             // This should be caught by the proxy, but we'll redirect just in case it is not.
             if (domain) {
                 window.location.replace(
--- a/app/routes/provider.js
+++ b/app/routes/provider.js
@@ -3,7 +3,6 @@ import config from 'ember-get-config';
 import getRedirectUrl from '../utils/get-redirect-url';
 
 const providers = config.PREPRINTS.providers.slice(1);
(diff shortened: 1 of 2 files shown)


config/environment.js
```diff
@@ -46,7 +46,8 @@ module.exports = function(environment) {
                         width: 363,
                         height: 242
                     },
-                    permissionLanguage: 'no_trademark'
+                    permissionLanguage: 'no_trademark',
```
brianjgeiger: I would rather fix this by building this list from the preprint_provider endpoint rather than having it stored in this js. @hmoco thought this should be doable without causing a lot of trouble. We should explore that option.

app/routes/provider.js
```diff
@@ -19,39 +18,49 @@ export default Ember.Route.extend({
     beforeModel(transition) {
         const {slug} = transition.params.provider;
         const slugLower = (slug || '').toLowerCase();
+        this.store
+            .findAll('preprint-provider')
```
brianjgeiger: Instead of this, just try to load the preprint provider detail page. The api equivalent is, rather than looking through `/v2/preprint_providers/` for `socarxiv`, try to load `/v2/preprint_providers/socarxiv/` and see if it give a 200 or a 400 series error.

The only reason to do this is if you're pre-fetching all of the preprint_provider models for the life of the app, and it's just a few calls once rather than, say, each time you switch providers. But I will let the next reviewer (who should be better versed in ember) decide if this way makes sense or not.

sheriefvt: I made another PR #340 with your suggestion. Thanks

brianjgeiger:
Closing in favor of #340

PR #344, opened 2017-05-04
Fix alignment for subject taxonomies [EOSF-285]

baylee-d:
## Purpose

To fix the alignment for the subject taxonomies.

## Changes

The ` inline-block ` style was throwing the subject class off.  Changes include removing that styling from the ` .subject ` class and tweaking styling for other containers that used that property of the ` .subject ` class.

## Before
screen shot 2017-05-04 at 11 35 37 am

## After
After changes, the subject taxonomies look like they did before previous changes to the class.
screen shot 2017-05-04 at 11 33 55 am

The subject-picker, which used the ` inline-block ` style of the ` .subject ` class, also looks like it was designed to before changes.
screen shot 2017-05-04 at 11 33 08 am

https://openscience.atlassian.net/browse/EOSF-285

### DIFF
--- a/app/components/subject-picker/template.hbs
+++ b/app/components/subject-picker/template.hbs
@@ -1,6 +1,6 @@
 <div class="row">
     {{#each selected as |subject|}}
-        <span class="subject">
+        <span class="subject subject-align">
             {{#each subject as |segment|}}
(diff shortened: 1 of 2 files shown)


brianjgeiger:
@baylee-d Can you do before and after screenshots?

PR #354, opened 2017-05-12
[EOSF-488] Disable spanish translations by renaming the locale directory

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-488

## Purpose

Disable the Spanish translations

## Changes

Renamed the es locale directory to an invalid/unused locale

## Side effects

Spanish translations will not be used in preprints. Will use default locale (en).


brianjgeiger:
```
not ok 376 PhantomJS 2.1 - ESLint - utils/get-redirect-url.js: global failure
    ---
        actual: >
            null
            null
        stack: >
            http://localhost:4201/assets/test-support.js:7424
        message: >
            Error: 
            Error: Could not find module `preprint-service/locales/es/translations` imported from `preprint-service/tests/unit/utils/locales-test`
        Log: |
    ...
```

jamescdavis:
Ah yes. So, should I fix the test to continue to check that we have Spanish translations for every key (even though we're not using them now) or just skip it? I'd vote for continue to check so we can be sure they'll be ready to go when we enable them again.

PR #356, opened 2017-05-12
[EOSF-657] Fix color of branded subjects on submit subject picker

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-657

## Purpose
Fix color of branded subjects on submit subject picker

## Changes
Move CSS around so it can be overwritten by brand, add right-arrow element to branded css

#### Before:
screen shot 2017-05-12 at 2 42 23 pm

#### After:

screen shot 2017-05-12 at 2 42 01 pm


brianjgeiger:
Before and after screenshots, please.

PR #361, opened 2017-05-18
[EOSF-601] Only send metrics when license is actually being saved

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-601

## Purpose
Only send metrics when license is saved, not when it changes.


brianjgeiger:
And this will have to be set, once we've updated ember osf, to use 0.5.1

PR #364, opened 2017-05-22
[EOSF-648] Use static css files and osf_assets file structure for preprint provider assets

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-648

## Purpose
Use the new file structure which will be available through https://github.com/brianjgeiger/osf-assets

## Changes
Addition of assets in the same file structure as the osf-assets repo, including static css, and pointing to those new assets


public/assets/img/provider_logos/bitss/style.css
```diff
@@ -0,0 +1,135 @@
...
+.preprint-form-section header {
+  color: #1C4063; }
+
+.preprint-header {
+  background-image: none;
+  background-color: #e4f3ff;
+  color: black;
+  padding-top: 100px; }
+  .preprint-header .provider-brand {
+    height: 140px; }
```
brianjgeiger: Image sizes (and their related containers) should be defined outside of these CSS files as they are all standardized now. In a pinch we can do more simplification later; it doesn't have to be perfect to start, but the better it is, the easier it will be to maintain.

public/assets/img/provider_logos/bitss/style.css
```diff
@@ -0,0 +1,135 @@
...
+  .preprint-header a:not(.btn) {
+    color: black; }
+
+.navbar-image {
+  background-image: url("./square_color_transparent.png"); }
```
brianjgeiger: As discussed, all image names and locations should be defined in the template rather than the CSS.

hmoco:
@brianjgeiger This is good for a code review

app/services/theme.js
```diff
@@ -70,29 +70,36 @@ export default Ember.Service.extend({
...
+    assetsPath: Ember.computed('id', function() {
         const id = this.get('id');
 
         if (!id)
             return;
-
-        const prefix = this.get('isDomain') ? '' : '/preprints';
-        const suffix = config.ASSET_SUFFIX ? `-${config.ASSET_SUFFIX}` : '';
-        return `${prefix}/assets/css/${id}${suffix}.css`;
+        return `${config.providerAssetsURL}${id}/`;
```
brianjgeiger: Aaaactually… there should be a javascript library for building URL for parts. We are probably using it somewhere, maybe in Ember OSF or possibly somewhere in preprints. We need to use that for these types of things.

PR #342, opened 2017-04-28
[EOSF-623] domain redirects

binoculars:
### ❗️ Needs:
- [x] CenterForOpenScience/osf.io#7167
- [x] CenterForOpenScience/ember-osf#203

## Ticket

https://openscience.atlassian.net/browse/EOSF-623

## Purpose

Redirect only if domain_redirect_enabled field is true on the preprint provider

## Changes

- Updates the ember-cli-build and detects if the OSF URL in the current origin
- add-domains script now uses the API to get the domains for /etc/hosts
- remove domains from config/environment (no longer needed)
- router updated to set theme id on init from the API
- theme service updated to redirect to the domain if domainRedirectEnabled = true for the provider (only needed for local dev)
- Simplify redirects and remove get-redirect-url util

PR #366, opened 2017-05-24
Merge master

binoculars:
Master back into develop

Ref: #342


brianjgeiger:
No, I reverted out the old version. I just wanted a develop-targeted version of the previous hotfix. It shouldn't have targeted master, since none of the OSF code is merged to master yet.

PR #367, opened 2017-05-24
Feature/db domain redirects

binoculars:
~Hotfix~ Feature Ref: #342


brianjgeiger:
I don't want to be too picky, but could this just cherry pick your two relevant commits? There's a lot more "Merge branch 'master'" commits than I usually like to see in feature branches.

binoculars:
@brianjgeiger rebased and RTM

PR #374, opened 2017-05-30
Make providerAssetsURL settable via environment variable

jamescdavis:
## Ticket

N/A

## Purpose

Make providerAssetsURL settable via environment variable.

## Changes

Changed providerAssetsURL in config/environment.js to be set via PROVIDER_ASSETS_URL environment variable and fall back to staging-cdn if the environment variable is not set.

## Side effects

None.


jamescdavis:
I tested this both with environment variable set and unset.

PR #383, opened 2017-06-08
Feature/ASU Repository [EOSF-680]

pattisdr:
## Ticket
https://openscience.atlassian.net/browse/EOSF-680
❗️ depends on https://github.com/CenterForOpenScience/ember-osf/pull/223
❗️ Might need some follow-up changes to the new navbar to conditionally remove the "Add preprint" link on a branded provider.
⭐️ NOTE: additionalProviders array on preprint-provider should have the names of the sources that SHARE is expecting, otherwise we will need some sort of transformation in preprints
⭐️ OSF Provider also needs allowSubmissions populated

## Purpose

Make preprints-side changes to implement ASU repository.  Preprints templates/controllers/routes have been modified themselves to make a collection of multiple types of resources with multiple providers work with the same code as branded preprint-providers.

## Changes
If additionalProviders exist on the preprint-provider:
- Header on Discover Page is "Repository Search" instead of "Preprint Archive Search"
- Use the following SHARE facets: "sources", "date", "type", and "tags" facets instead of  "provider" and "taxonomy" facets
-  Return search results of any type, not just preprints
- Don't show activeFilters (SHARE filters are not designed to work with this at this time)
- Show additionalProviders list instead of top-level taxonomies on index page
- Search placeholder is "Search repository... instead of search preprints..."

If allowSubmissions is False:
- Remove Add preprint link on navbar
- Remove Add preprint button on index page
- Remove Add preprint button on discover page
- Transition to not found if attempting to navigate to submit route directly

Additionally:
- Added c3, d3, bootstrap-daterange picker dependencies so new share-facets could be used
- Increase the list of query params that are permitted for preprints - now "sources", "tags", "type", "start", and "end" are enabled.

## Landing Page
index

## Discover Page
asu-discover

## Submit Page
submit-asu

## For QA
Most of these changes were made to code for branded preprints.  OSF Preprints and another branded preprint, like psyarxiv's Landing Pages, Discover Pages, and Submit Pages should be checked to make sure all the pieces are there. For submit page - just make sure it loads.

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -27,13 +27,15 @@
                     {{total-share-results}}
                 </div>
             </div> {{!END ROW}}
-            <div class="row">
-                <div class="text-center col-xs-12">
-                    <p class="lead">{{t 'index.header.or'}}</p>
-                    <a type="button" class="btn btn-success btn-lg preprint-submit-button" onclick={{action "click" "link" "Index - Add preprint"}} href={{concat theme.pathPrefix 'submit'}}>{{t "global.add_preprint"}}</a>
-                    <div class="m-t-md">{{#link-to (route-prefix 'content') (if theme.provider.example theme.provider.example 'khbvy') invokeAction=(action "click" "link" "Index - See example")}}{{t "index.header.example"}}{{/link-to}}</div>
+            {{#if theme.provider.allowSubmissions}}
+                <div class="row">
+                    <div class="text-center col-xs-12">
+                        <p class="lead">{{t 'index.header.or'}}</p>
+                        <a type="button" class="btn btn-success btn-lg preprint-submit-button" onclick={{action "click" "link" "Index - Add preprint"}} href={{concat theme.pathPrefix 'submit'}}>{{t "global.add_preprint"}}</a>
+                        <div class="m-t-md">{{#link-to (route-prefix 'content') (if theme.provider.example theme.provider.example 'khbvy') invokeAction=(action "click" "link" "Index - See example")}}{{t "index.header.example"}}{{/link-to}}</div>
+                    </div>
                 </div>
-            </div>
+            {{/if}}
         </div> {{!END CONTAINER}}
     </div> {{!END HEADER}}
 
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -44,7 +44,7 @@
         <div class="container"> {{!CONTAINER DIV}}
             <div class="row p-v-md">{{!SUBJECTS ROW}}
                 <div class="col-xs-12">
-                    <h2>{{t "index.subjects.heading"}}</h2>
+                    <h2>{{if theme.provider.additionalProviders (t "index.subjects.heading" browse-by="provider") (t "index.subjects.heading" browse-by="subject")}}</h2>
                     <div class="subjectsList p-lg"> {{!SUBJECTS LIST}}
                         {{#if theme.provider.additionalProviders}}
                             {{additional-provider-list additionalProviders=theme.provider.additionalProviders}}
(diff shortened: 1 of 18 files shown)


app/templates/index.hbs
```diff
@@ -44,7 +46,11 @@
                 <div class="col-xs-12">
                     <h2>{{t "index.subjects.heading"}}</h2>
                     <div class="subjectsList p-lg"> {{!SUBJECTS LIST}}
-                        {{taxonomy-top-list list=model.taxonomies}}
+                        {{#if theme.provider.additionalProviders}}
+                            {{additional-provider-list additionalProviders=theme.provider.additionalProviders}}
```
hmoco: Is there not an usecase for providers having additional providers show up on the search page, while still allowing the submissions (and thus using taxonomy)?
I guess what I'm asking is shouldn't this be using theme.provider.allowSubmissions to show taxonomy? And if so, does showing both look ok (probably not somthing that needs to be looked into until it becomes an actual scenarion)?

pattisdr: I thought the presence of additionalProviders made more sense here, because if they exist, then that is what we show.  Otherwise, we show taxonomies.

I have expressed that these should be temporary changes to get ASU out the door. (You can see some of my comments in the JIRA ticket for this).   There will be a lot of usecases that this PR doesn't adequately handle, and many more options that other collections might want. ASU is being treated like a branded preprint provider here, but it is actually a collection of lots of resource types, not just preprints, and is not a single brand, but a bunch of providers.  

I am uncomfortable putting a bunch of special case handling in these templates for one type of collection and think these pages need to become more API-driven than they already are.

brianjgeiger: The intention was to have additional providers be separate from allowing submissions but allowing submissions to be tied to not having taxonomies. That being said, yes, getting this out the door is key, but we should make follow-up tickets to clean this up immediately so that we don't have to maintain the hacks long-term and we can actually do it properly.

hmoco:
With the context provided by @brianjgeiger and @pattisdr, this looks good! Incredibly awesome and detailed PR description too!

hmoco:
Will do a quick functional review shortly (being gone for 2 weeks has broken some things locally)

pattisdr:
Thanks @hmoco!  There's a `populate_fake_preprint_providers script` now, and I added the following to test ASU locally to the list of preprint_providers:

```
{
        '_id': 'asu',
        'name': '[TEST] ASU',
        'domain': '',
        'domain_redirect_enabled': False,
        'default_license': '',
        'licenses_acceptable': [],
        'description': 'Explore the latest research being performed in Arizona’s best learning institutions',
        'allow_submissions': False,
        'additional_providers': ['Arizona State University Digital Repository', 'The Digital Archaeological Record', 'Symbiota']
    },
```

app/templates/index.hbs
```diff
@@ -44,7 +46,11 @@
                 <div class="col-xs-12">
                     <h2>{{t "index.subjects.heading"}}</h2>
```
hmoco: Is this supposed to still have a subject heading? 
screen shot 2017-06-13 at 2 19 59 pm

pattisdr: good catch! I will add another conditional for this.

hmoco:
One quick thing about the heading, and the search widgets are a little finecky:
screen shot 2017-06-13 at 2 25 44 pm
screen shot 2017-06-13 at 2 26 21 pm

Playing with them leads to a few small weird issues every now and then. I'm assuming this is an ember-osf share thing rather than a problem with this PR.

pattisdr:
@hmoco   "0 sources" is expected if you further constrain your search where there are no results. Do you recall what you did to see the first image?

hmoco:
@pattisdr can't reproduce it :(. This PR looks good to go though, will put it on passed code review!

PR #388, opened 2017-06-15
[EOSF-696] easy update of assets

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-696

## Purpose

Add a way to pull in provider assets from the assets repo and update them easily. Make it possible to use the local assets with fingerprinting instead of assets from CDN (while still allowing use of CDN).

To use local assets instead of a CDN, set the PROVIDER_ASSETS_URL environment variable to 'local'.

To create a hotfix with updated local assets run `npm run update-assets-hotfix` or `yarn update-assets-hotfix`.

To update local assets only run `npm run update-assets` or `yarn update-assets`.

## Changes

* added assets repo as git submodule
* added util and helper for resolving provider asset paths
* used the util or helper everywhere provider assets are referenced
* added npm script to update assets "npm update-assets"

Also:
* updated to latest ember-osf
* upgraded ember-cli to 2.11.1 to be compatible with latest ember-osf

## Side effects

Need to make sure all provider assets are resolving in all cases.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "public/assets/osf-assets"]
+	path = public/assets/osf-assets
+	url = https://github.com/brianjgeiger/osf-assets.git
```
brianjgeiger: Perhaps I'll get this transferred over to COS this morning, though I presume it will redirect in any case.

app/templates/components/provider-carousel.hbs
```diff
@@ -10,14 +10,14 @@
...
-                                <div class="provider-logo provider-{{provider.id}}" style="background-image: url('{{providerAssetsURL}}{{provider.id}}/wide_white.png')"></div>
+                                <div class="provider-logo provider-{{provider.id}}" style="background-image: url('{{providerAsset provider.id 'wide_white.png'}}')"></div>
                             </a>
                         {{else}}
                             <a href="/preprints/{{provider.id}}/discover"
                                 title={{provider.name}}
                                 onclick={{action 'click' 'link' (concat 'Discover - Provider Carousel ' provider.name)}}
                             >
-                                <div class="discover-provider-logo provider-{{provider.id}}-dark" style="background-image: url('{{providerAssetsURL}}{{provider.id}}/wide_black.png')"></div>
+                                <div class="discover-provider-logo provider-{{provider.id}}-dark" style="background-image: url('{{providerAsset provider.id 'wide_black.png'}}')"></div>
```
brianjgeiger: So much better

app/utils/build-provider-asset-path.js
```diff
@@ -0,0 +1,16 @@
...
+import pathJoin from '../utils/path-join';
+
+export default function buildProviderAssetPath(providerId, assetName, isDomain) {
+    if (config.providerAssetsURL != 'local') {
+        return pathJoin(config.providerAssetsURL, providerId, assetName);
+    }
+    if (config.ASSET_SUFFIX) {
+        assetName = assetName.replace(/(\.[^\.]+)$/, function(match, $1) { return `-${config.ASSET_SUFFIX}${$1}`; });
+    }
+    const providerAssetPath = 'assets/osf-assets/files/preprints-assets';
```
brianjgeiger: This should probably be in a configuration setting rather than in the middle of a file. Not that I expect it to change a lot, but just so it's centralized.

tests/unit/utils/build-provider-asset-path-test.js
```diff
@@ -0,0 +1,15 @@
...
+
+test('build correct path for domain', function(assert) {
+    let result = buildProviderAssetPath('foo', 'bar.baz', true);
+    assert.equal(result, `${config.providerAssetsURL}foo/bar.baz`);
+});
+
+test('build correct path for non-domain', function(assert) {
+    let result = buildProviderAssetPath('foo', 'bar.baz', false);
+    assert.equal(result, `${config.providerAssetsURL}foo/bar.baz`);
+});
```
brianjgeiger: I think there should probably be some test cases for local/non-local as well, yes? Or is that not easily testable?

brianjgeiger:
Okay, the osf-assets repo is now firmly at centerforopenscience/osf-assets.

app/utils/build-provider-asset-path.js
```diff
@@ -2,7 +2,7 @@ import config from 'ember-get-config';
 import pathJoin from '../utils/path-join';
 
 export default function buildProviderAssetPath(providerId, assetName, isDomain) {
-    if (config.providerAssetsURL) {
+    if (config.providerAssetsURL != 'local') {
```
brianjgeiger: Ohhhhh, I see. providerAssetsURL can either be a URL or can be 'local' as a special case, so it's just the one variable. Yeah, that works, just mention in the ticket description.

brianjgeiger: And also should be documented by the setting if it isn't.

app/utils/build-provider-asset-path.js
jamescdavis: Yeah, it felt wrong hardcoding this here. If it's in config, it'd be easier to update if the assets repo structure changes, we bring assets in using a different method, etc.

tests/unit/utils/build-provider-asset-path-test.js
jamescdavis: I agree and certainly wanted to test that, but I haven't quite figured out how to run automated tests with different config environments. I'll see what I can do...

jamescdavis:
Regarding hotfixing, do you want it to create a hotfix branch, update the assets, and make a commit?

brianjgeiger:
And optionally push would be ideal. But again, we can cover that in an additional ticket.

app/utils/build-provider-asset-path.js
```diff
@@ -0,0 +1,15 @@
+import config from 'ember-get-config';
+import pathJoin from '../utils/path-join';
+
+export default function buildProviderAssetPath(providerId, assetName, isDomain) {
```
brianjgeiger: If you passed in the config.providerAssetsURL rather than pulling it here, then it would be super-simple to test.

jamescdavis:
Now does everything except push.

PR #392, opened 2017-06-19
[EOSF-655] Use t-plus helper to add provider based logic to naming things.

hmoco:
## Needs: https://github.com/CenterForOpenScience/ember-osf/pull/254
Tests will fail until that is merged and pinned here.

## Ticket

https://openscience.atlassian.net/browse/EOSF-655

## Purpose
Give the ability to change the "preprint" wording to something providers might prefer, such as "papers".

## Changes
Any mention of preprints in user-facing strings (the ones seen on `translation.js`) uses `{{preprintWords.preprint}}` over `preprint`, `{{preprintWords.Preprints}}` over `Preprints` and so on.

## Side effects
All new strings need to be defined using preprintWords


jamescdavis:
Try regenerating yarn.lock

jamescdavis:
Should be good once updated to use ember-osf with extended t helper.

PR #393, opened 2017-06-19
[EOSF-716] Update deps, fixes for latest ember-osf

binoculars:
Had to update some dependencies and skip tests to make it work with the latest ember-osf updates

## Ticket

https://openscience.atlassian.net/browse/EOSF-716 

## Purpose

Make preprints work with the latest ember-osf

## Changes

Dependency Updates, all developer-facing changes.

## Notes for QA
Normal regression only, just updating dependencies to get preprints tests working again


binoculars:
Resolved

package.json
```diff
@@ -83,11 +85,9 @@
     "ember-welcome-page": "1.0.1",
     "eslint": "3.18.0",
     "faker": "^4.1.0",
-    "jscs": "^3.0.7",
-    "liquid-fire": "0.24.1",
+    "liquid-fire": "^0.27.3",
```
jamescdavis: Should we pin this?

binoculars: We can. If not, it will only get updated when someone adds/removes/updates package.json, otherwise, everything is set to use `--pure-lockfile` (travis, docker, docker-compose, etc)

PR #401, opened 2017-06-23
[feature] Edit Advisory Board Allowed Tags [No Ticket]

caseyrollins:
Relates to OSF PR [#7381](https://github.com/CenterForOpenScience/osf.io/pull/7381)

#### Ticket
- No ticket.

#### Purpose
- Agrixiv uses `h3` tags in their advisory board on production, so we should probably allow that. 

#### Changes
- Edit allowed tags in advisory-board.js


caseyrollins:
@laurenrevere, is this the correct place to add another safe tag?

laurenrevere:
yup!

PR #414, opened 2017-07-13
[EOSF-764] Add preprint mint DOI's in metatags

baylee-d:
## Ticket

https://openscience.atlassian.net/browse/EOSF-764

## Purpose

To add mint DOI's to the meta tags so that Google Scholar can get access to them.

## Changes

- Added a conditional to check for either the Peer Review DOI's or the mint DOI's for Highwire Press
- Changed Dublin Core DOI to point at mint DOI's
- Added a conditional Dublin Core DOI in the case that there is a Peer Review DOI

In the case that there is a Peer Review DOI, `citation_doi` and `dc.relation` will point to that DOI.
screen shot 2017-07-13 at 2 14 45 pm

However, if there isn't a Peer Review DOI, then `citation_doi` will point to the mint DOI and `dc.relation` will not exist (since there isn't a peer review DOI)
screen shot 2017-07-13 at 2 19 24 pm

### DIFF
--- a/app/routes/content/index.js
+++ b/app/routes/content/index.js
@@ -143,11 +143,11 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
                     ['dc.title', title],
                     ['dc.abstract', description],
                     ['dc.identifier', canonicalUrl],
-                    ['dc.identifier', mintDoi],
+                    ['dc.identifier', 'doi:' + mintDoi],
                 ];
                 if(peerDoi !== '') {
                     dublinCore.push(
-                        ['dc.relation', peerDoi]
+                        ['dc.relation', 'doi:' + peerDoi]
                     );
                 }
 
--- a/app/routes/content/index.js
+++ b/app/routes/content/index.js
@@ -88,7 +88,7 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
                 const description = node.get('description');
                 const mintDoi = extractDoiFromString(preprint.get('links.preprint_doi'));
                 const peerDoi = preprint.get('doi');
-                const doi = peerDoi !== ''? peerDoi: mintDoi;
+                const doi = peerDoi !== '' ? peerDoi : mintDoi;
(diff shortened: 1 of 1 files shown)


app/routes/content/index.js
```diff
@@ -85,7 +86,9 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
             .then(([provider, node, license]) => {
                 const title = node.get('title');
                 const description = node.get('description');
-                const doi = preprint.get('doi');
+                const mintDoi = extractDoiFromString(preprint.get('links.preprint_doi'));
+                const peerDoi = preprint.get('doi');
+                const doi = peerDoi !== ''? peerDoi: mintDoi;
```
jamescdavis: Please add whitespace on both sides of ? and :

See: http://eslint.org/docs/rules/space-infix-ops

app/routes/content/index.js
```diff
@@ -136,12 +139,17 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
                 const prism = [];
 
                 // Dublin Core
-                const dublinCore = [
+                var dublinCore = [
                     ['dc.title', title],
                     ['dc.abstract', description],
                     ['dc.identifier', canonicalUrl],
-                    ['dc.identifier', doi]
+                    ['dc.identifier', mintDoi], // Should be the minted doi
```
jamescdavis: This comment is probably not necessary.

app/routes/content/index.js
```diff
@@ -136,12 +139,17 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
...
                 ];
+                if(peerDoi !== '') {
+                    dublinCore.push(
+                        ['dc.relation', peerDoi] // Should be peer doi
```
jamescdavis: Comment probably unneeded here as well.

jamescdavis:
Code looks good, but waiting to confirm whether dc.identifier and dc.relation should be prefixed with "doi:".

PR #425, opened 2017-08-02
[EOSF-691] show highlighted taxonomies

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-691

## Purpose

Show highlighted instead of top level taxonomies on index page.

## Changes

* use latest ember-osf to be able to access highlighted taxonomies from preprint provider
* query highlighted taxonomies instead of taxonomies with no parent

## Testing notes

* make sure highlighted child subjects appear and non-highlighted top-level subjects don't
* make sure all top level subjects appear when a provider has no highlighted taxonomies

### DIFF
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -19,10 +19,7 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, {
         return Ember.RSVP.hash({
             taxonomies: this.get('theme.provider')
                 .then(provider => provider
-                    .query('taxonomies', {
-                        filter: {
-                            parents: 'null'
-                        },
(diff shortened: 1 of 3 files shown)


hmoco:
@jamescdavis, although this has been merged, I just thought of something: will all providers have highlightedTaxonomies for sure? if that query returns 0 results, should it default to making the original query?

jamescdavis:
The way the back end works is that if a provider does not have any taxonomies explicitly highlighted, it will return the top level taxonomies as the highlighted taxonomies. We will always get the correct thing from the highlighted endpoint.

PR #434, opened 2017-08-14
[EOSF-773] Provider carousel link to domain with redirect enabled

baylee-d:
## Ticket

https://openscience.atlassian.net/browse/EOSF-773

## Purpose

When a provider has a domain and has domain_redirect_enabled set to true, it should link to it's domain instead of use an osf.io url.

## Changes

Added a conditional to check the domain and redirect settings for providers in the carousel.

## Testing

To test this, be sure to have a provider with both a domain and domain_redirect_enabled set to true.  The final result should look like this on preprints:
screen shot 2017-08-14 at 10 58 23 am
screen shot 2017-08-14 at 10 55 13 am

### DIFF
--- a/app/helpers/branded-preprint-url.js
+++ b/app/helpers/branded-preprint-url.js
@@ -1,0 +1,25 @@
+import Ember from 'ember';
+import pathJoin from '../utils/path-join';
+
+/**
+ * @module ember-preprints
+ * @submodule helpers
+ */
+
+/**
+ * To determine the link url for branded preprints on the provider carousel
+ *
+ * @class route-prefix
+ */
+
+export function brandedPreprintUrl(provider, lightLogo) {
+    let url = '';
+    if (provider[0].get('domain') && provider[0].get('domainRedirectEnabled')) {
+        url = lightLogo.type ? provider[0].get('domain') : pathJoin(provider[0].get('domain'), 'discover');
+    } else {
+        url = lightLogo.type ? '/preprints/' + provider[0].get('id') : '/preprints/' + provider[0].get('id') + '/discover';
+    }
+    return url;
+}
+
+export default Ember.Helper.helper(brandedPreprintUrl);
--- a/app/helpers/branded-preprint-url.js
+++ b/app/helpers/branded-preprint-url.js
@@ -6,12 +6,17 @@ import pathJoin from '../utils/path-join';
  * @submodule helpers
  */
 
-/**
- * To determine the link url for branded preprints on the provider carousel
- *
- * @class route-prefix
- */
-
+ /**
+  * To determine the link url for branded preprints on the provider carousel
+  * Returns a string of the url with the necessary path
+  *
+  * @class brandedPreprintUrl
+  * @param {Object[]} params all positional parameters.
+  * @param {Object} params[0] The provider the url is being generated for.
+  * @param {Object} hash all named parameters.
+  * @param {String} hash.path The path to append.
+  * @return {String} The url.
+  */
 export function brandedPreprintUrl(params, hash) {
     let url = '';
     let [provider] = params;
--- a/app/helpers/branded-preprint-url.js
+++ b/app/helpers/branded-preprint-url.js
@@ -12,12 +12,16 @@ import pathJoin from '../utils/path-join';
  * @class route-prefix
  */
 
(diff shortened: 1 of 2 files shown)


app/templates/components/provider-carousel.hbs
```diff
@@ -5,19 +5,27 @@
             <div class={{if (not index) "item active" "item"}}>
                 {{#each slide as |provider index|}}
                     <div class={{if (not index) (concat "col-sm-2 " columnOffset) "col-sm-2"}}>
-                        {{#if lightLogo}}
-                            <a href="/preprints/{{provider.id}}"
+                        {{#if (and provider.domain provider.domainRedirectEnabled)}}
```
jamescdavis: I'm not sure this logic should be in the template. Should probably create a computed property for this.

app/templates/components/provider-carousel.hbs
```diff
@@ -5,19 +5,27 @@
...
+                            <a href="{{provider.domain}}{{if lightLogo '' '/discover'}}"
```
jamescdavis: use computed property and pathJoin()

app/templates/components/provider-carousel.hbs
```diff
@@ -5,19 +5,27 @@
...
                                 title={{provider.name}}
                                 onclick={{action 'click' 'link' (concat 'Discover - Provider Carousel ' provider.name)}}
-                            >
-                                <div class="discover-provider-logo provider-{{provider.id}}-dark" style="background-image: url('{{providerAsset provider.id 'wide_black.png'}}')"></div>
+                                >
+                                {{#if lightLogo}}
+                                    <div class="provider-logo provider-{{provider.id}}" style="background-image: url('{{providerAsset provider.id 'wide_white.png'}}')"></div>
+                                {{else}}
+                                    <div class="discover-provider-logo provider-{{provider.id}}-dark" style="background-image: url('{{providerAsset provider.id 'wide_black.png'}}')"></div>
+                                {{/if}}
```
jamescdavis: Too much code duplication here.

app/helpers/branded-preprint-url.js
```diff
@@ -0,0 +1,25 @@
...
+ * @submodule helpers
+ */
+
+/**
+ * To determine the link url for branded preprints on the provider carousel
+ *
+ * @class route-prefix
+ */
+
+export function brandedPreprintUrl(provider, lightLogo) {
```
jamescdavis: The canonical way to use parameters in a helper is:
```
export function myHelper(params, hash) {
    let [param1, param2] = params;
    let myNamedParam = hash.foo;
}
```
So `params` is a list of positional parameters and `hash` is an object containing the named parameters.

app/helpers/branded-preprint-url.js
```diff
@@ -0,0 +1,25 @@
...
+    let url = '';
+    if (provider[0].get('domain') && provider[0].get('domainRedirectEnabled')) {
+        url = lightLogo.type ? provider[0].get('domain') : pathJoin(provider[0].get('domain'), 'discover');
+    } else {
+        url = lightLogo.type ? '/preprints/' + provider[0].get('id') : '/preprints/' + provider[0].get('id') + '/discover';
+    }
```
jamescdavis: I think the logic would be cleaner is you first build the base url and then append `/discover` if necessary.

app/templates/components/provider-carousel.hbs
```diff
@@ -5,21 +5,16 @@
...
-                                title={{provider.name}}
-                                onclick={{action 'click' 'link' (concat 'Index - Provider Carousel - ' provider.name)}}
-                                >
-                                <div class="provider-logo provider-{{provider.id}}" style="background-image: url('{{providerAsset provider.id 'wide_white.png'}}')"></div>
-                            </a>
-                        {{else}}
-                            <a href="/preprints/{{provider.id}}/discover"
-                                title={{provider.name}}
-                                onclick={{action 'click' 'link' (concat 'Discover - Provider Carousel ' provider.name)}}
+                        <a href="{{brandedPreprintUrl provider type=lightLogo}}"
```
jamescdavis: Instead of `type` we should use a better name for this parameter. What might even be better is to just pass in the path to append and leave the logic of deciding what to append in the template.

PR #435, opened 2017-08-14
Hotfix/0.111.15 [EOSF-655]

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-655 

## Purpose

Hide "Theses" from page title and nav bar for Thesis Commons

## Changes

* add property to theme service for determining if the preprint word should be included in the title
* use this property to determine when to include the preprint word in the:
  * page title
  * nav bar

## Side effects

N/A

### DIFF
--- a/app/services/theme.js
+++ b/app/services/theme.js
@@ -53,6 +53,11 @@ export default Ember.Service.extend({
         return this.get('id') !== 'osf';
     }),
 
+    // If we should include the preprint word in the title
+    preprintWordInTitle: Ember.computed('id', function() {
+        return this.get('id') !== 'thesiscommons';
+    }),
+
     // If we're using a branded provider and not under a branded domain (e.g. /preprints/<provider>)
     isSubRoute: Ember.computed('isProvider', 'isDomain', function() {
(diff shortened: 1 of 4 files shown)


hmoco:
A small test would be good for this

hmoco:
General comment: We reaally don't want to keep on adding exceptions for specific providers. I understand the context of this specific fix, but if this is something we might do again, a generalization of the wanted behavior is "This is the name of the brand", like OSF Preprints, and as such I can see it as an attribute of preprint provider (under something like 'brandName'). The way the titles are constructed right now don't allow for that, due to allowing translation of the type of document being uploaded, but, as a brand, I don't think it is a stretch to say that it's ok if something like EngrXiv Papers is shown as such in other languages, provided other mentions of "paper", when referring to the document a user is looking at or uploading, are translated. 
The short of it is that this type of behavior should be reserved to the preprint provider model, and we don't really want to be doing this in the first place for most things (for example, we don't currently support having extra description in the steps to submit a preprint; if a new provider suddenly wanted that, I would argue that our responsibility is to communicate our limitations better, rather than add pieces of functionality that are going to be used for one provider).

PR #436, opened 2017-08-15
Hotfix/0.111.16 [EOSF-814]

jamescdavis:
## Ticket

https://openscience.atlassian.net/browse/EOSF-814

## Purpose

"Other Preprint Repositories" should always link to the OSF domain, even when on a branded domain.

## Changes

Use OSF.url in the link.

## Side effects

Should do the right thing in all environments.

### DIFF
--- a/app/components/search-facet-provider.js
+++ b/app/components/search-facet-provider.js
@@ -30,6 +30,7 @@ export default Ember.Component.extend(Analytics, {
     otherProviders: [],
     searchUrl: config.OSF.shareSearchUrl,
     whiteListedProviders: config.whiteListedProviders,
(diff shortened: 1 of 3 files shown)


brianjgeiger:
Passing note: This makes sense for preprint providers, but does it for Thesis Commons? I imagine this question will have to be answered with an improvement ticket in a normal release.

jamescdavis:
@brianjgeiger I asked Product about this last night and the answer was "Sort of..." 😄 
The link does change to "Other Thesis Providers" on Thesis Commons, which doesn't really make since right now, but will make more sense when we aggregate more thesis providers.

PR #451, opened 2017-09-01
Feature/cleanup

chrisseto:
## Ticket

https://openscience.atlassian.net/browse/EOSF-


laurenbarker:
Closing this since the PR won't ever be updated. The first three commits should be in preprints already. @alexschiller the last commit has some upgrades that might be nice to reference for upgrading the dependencies (in particular ember-bootstrap introduced some breaking changes to modals).

PR #456, opened 2017-09-20
[EOSF-615] Remove maintenance banner from preprints, in ember-osf now

hmoco:
## Ticket

https://openscience.atlassian.net/browse/EOSF-615
## Purpose

[EOSF-615] Allow use of different types of maintenances and custom messages


hmoco:
@jamescdavis, could you take a look at this? (and do a functional test please?)

app/components/maintenance-banner/component.js
```diff
@@ -1,12 +1,18 @@
 import Ember from 'ember';
 import config from 'ember-get-config';
 
+var levelMap = {
```
jamescdavis: const?

app/components/maintenance-banner/component.js
```diff
@@ -15,6 +21,11 @@ export default Ember.Component.extend({
 
     hidden: 'true',
 
+    alert: true,
+    _info: true,
```
jamescdavis: Why not use a single bound property that's a string? The you can just do something like:
```
classNameBindings: ['alertClass'],

alertClass: 'alert-info',
.
.
.
this.set('alertClass', `alert-${levelMap[maintenance.level]}`);
```

jamescdavis: Also, you can use classNames and classNameBindings simultaneously so you can put `alert` in classNames without needing a bound property.

app/components/maintenance-banner/component.js
```diff
@@ -1,12 +1,18 @@
...
 
+var levelMap = {
+    1: 'info',
+    2: 'warning',
+    3: 'danger'
+};
+
 export default Ember.Component.extend({
-    classNames: [
+    classNamesBindings: [
```
jamescdavis: Should be `classNameBindings` (you have an extra s)

app/components/maintenance-banner/component.js
```diff
@@ -34,6 +45,8 @@ export default Ember.Component.extend({
                             hidden: false,
                             maintenance
                         });
+                        this.set('_info', false);
+                        this.set(`_${levelMap[maintenance.level]}`, true);
```
jamescdavis: Should gracefully handle the case when maintenance.level is `null`, `undefined`, or not mapped in levelMap.

jamescdavis:
Changes look good. Tested and it works! Will merge after tests pass.

PR #463, opened 2017-10-10
[MOD-138][Improvement] Show status banner to contributors

laurenbarker:
## Ticket

https://openscience.atlassian.net/browse/MOD-138 

## Purpose

Show status banner to all contributors.

## Side effects

There is a bit of a delay before the status bar shows up for non-admin contributors.

### DIFF
--- a/app/controllers/content/index.js
+++ b/app/controllers/content/index.js
@@ -62,9 +62,18 @@ export default Ember.Controller.extend(Analytics, {
     currentUser: Ember.inject.service(),
     expandedAuthors: true,
     showLicenseText: false,
+    currentUserId: '',
     expandedAbstract: navigator.userAgent.includes('Prerender'),
     queryParams: {
         chosenFile: 'file'
+    },
+
+    init() {
+        this._super(...arguments);
+        this.get('currentUser').load()
+            .then(user => {
+                this.set('currentUserId', user.id);
+            });
     },
 
     dateLabel: Ember.computed('model.provider.reviewsWorkflow', function() {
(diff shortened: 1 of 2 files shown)


aaxelb:
Out of curiosity, could we instead check whether the user has `write` permission on the preprint's node (basically how `isAdmin` is implemented)? Would that be equivalent?

laurenbarker:
I wish. There are read only contributors and turns out everyone has `read` permissions on public preprints.

app/controllers/content/index.js
```diff
@@ -82,6 +91,19 @@ export default Ember.Controller.extend(Analytics, {
         // True if the current user has admin permissions for the node that contains the preprint
         return (this.get('node.currentUserPermissions') || []).includes(permissions.ADMIN);
     }),
+
+    userIsContrib: Ember.computed('authors.[]', 'isAdmin', function() {
```
aaxelb: Could this use/depend on `currentUser.currentUserId` instead of loading the user in `init`?

app/controllers/content/index.js
```diff
@@ -261,23 +281,11 @@ export default Ember.Controller.extend(Analytics, {
...
-                        Ember.get(this, 'metrics').invoke('trackSpecificCollection', 'Keen', keenPayload); // Sends event to Keen if logged-in user is not a contributor
-                    }
-                })
-                .catch(() => {
-                    Ember.get(this, 'metrics').invoke('trackSpecificCollection', 'Keen', keenPayload); // Sends event to Keen for non-authenticated user
-                });
+            if (!this.get('userIsContrib')) {
+                Ember.get(this, 'metrics').invoke('trackSpecificCollection', 'Keen', keenPayload); // Sends event to Keen if logged-in user is not a contributor
+            } else {
+                Ember.get(this, 'metrics').invoke('trackSpecificCollection', 'Keen', keenPayload); // Sends event to Keen for non-authenticated user
```
aaxelb: This if statement does the same thing in both branches. Based on the original code, seems like it should just be the first one -- the new `userIsContrib` will be false for non-authenticated users, too.

laurenbarker: Oh yup

PR #467, opened 2017-10-16
[EOSF-859] Update node when license changes

baylee-d:
## Ticket

https://openscience.atlassian.net/browse/EOSF-859

## Purpose

Currently, the node is not being updated whenever the license data on a preprint is being changed.  The changes should fix this.

## Changes

Add `node.setProperties()` section when updating after the license data is changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Get the sentryDSN from the ember config
 - Point Bower to new Bower registry (https://registry.bower.io)
 - Update to @centerforopenscience/ember-osf@0.11.0
+- Update node when license data is changed
 
 ### Removed
 - arXiv license permission language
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -17,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Get the sentryDSN from the ember config
 - Point Bower to new Bower registry (https://registry.bower.io)
 - Update to @centerforopenscience/ember-osf@0.11.0
+- Update node when license data is changed
```
jamescdavis: This should be added to the Unreleased section. Actually, for hotfixes, we should call it [Hotfix]

app/controllers/submit.js
```diff
@@ -720,6 +720,12 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
                     },
                     license: this.get('basicsLicense.licenseType')
                 });
+                node.setProperties({
+                    description: currentAbstract,
+                    tags: currentTags,
+                    license: this.get('basicsLicense.licenseType'),
+                    nodeLicense: this.get('basicsLicense.licenseType'),
+                });
```
jamescdavis: This should only happen if the check box to apply the license to the node is checked.

baylee-d:
After testing, all of this works locally but not on staging/production.  Closing this PR in favor of the proper fix.

PR #469, opened 2017-10-18
Use old-file-browser

hmoco:
## Purpose
We (read I) changed file-browser significantly to support quick files, which in turn broke usability for its one other use case, choosing a file from an existing node when creating a preprint.


jamescdavis:
Looks like this change has already been made: https://github.com/CenterForOpenScience/ember-osf-preprints/blob/develop/app/templates/components/preprint-form-project-select.hbs#L75

PR #471, opened 2017-10-19
[EOSF-851] Use provider favicon on provider pages

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-851

## Purpose
Allow provider pages to have their own icons as the page's favicon, instead of blank or OSF, regardless of the domain.
screen shot 2017-10-19 at 12 08 44 pm

## Testing considerations
Make sure it works on branded domains (such as engrxiv.org) as well as our version of it (such as osf.io/preprints/engrxiv)


app/routes/application.js
```diff
@@ -15,7 +16,19 @@ export default Ember.Route.extend(Analytics, OSFAgnosticAuthRouteMixin, {
...
+            type: 'link',
+            attrs: {
+                rel: 'shortcut icon',
+                href: buildProviderAssetPath(config, this.get('theme.id'), 'favicon.ico', window.isProviderDomain)
+            }
+        }]
+    },
+    themeObserver: Ember.observer('theme.id', function() {
+        this.get('headTagsService').collectHeadTags();
+    }),
```
jamescdavis: It feels like all of this should go in the theme and here we should just do something like:
`headTags: this.get('theme.headTags')`

Thoughts?

hmoco: I like that, although i would be a bigger fan of it if we could also use the 'head-tags' handlebars helper within theme, but I don't know how we would go about doing that

app/services/theme.js
```diff
@@ -121,4 +122,17 @@ export default Ember.Service.extend({
...
+    headTags: Ember.computed('id', function() {
+        return [{
+            type: 'link',
+            attrs: {
+                rel: 'shortcut icon',
+                href: buildProviderAssetPath(config, this.get('id'), 'favicon.ico', window.isProviderDomain)
+            }
+        }]
+    }),
+    themeObserver: Ember.observer('id', function() {
```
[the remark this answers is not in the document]
jamescdavis: [The Ember Guide](https://guides.emberjs.com/v2.16.0/object-model/observers/) uses `${propName}Changed` as a convention. I like it. So `idChanged`?

hmoco:
@jamescdavis, This is a change that I don't think can be tested easily, as the test-able parts are mostly the headTag internals and changes to <head>. Opinion on coveralls failing?

PR #472, opened 2017-10-20
[MOD-158][Fix] Refresh model after resubmission

laurenbarker:
## Ticket

https://openscience.atlassian.net/browse/MOD-158

## Purpose

Show "pending" banner after resubmission.

## Changes

Refresh the model on resubmit.

## Side effects

None

### DIFF
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -1085,6 +1085,7 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
 
             return submitAction.save()
                 .then(() => {
+                    this.get('model').reload();
                     this.transitionToRoute(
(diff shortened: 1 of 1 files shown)


app/controllers/submit.js
```diff
@@ -1085,6 +1085,7 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
 
             return submitAction.save()
                 .then(() => {
+                    this.get('model').reload()
```
aaxelb: `;`?

laurenbarker: Dang it.

laurenbarker: Fixed

PR #479, opened 2017-11-02
[IN-78][Preprints] Add original publication date to preprints.

adlius:
## Purpose

This is the second of two PRs that add original publication date to basics form and preprint content page. (The first PR is [here](https://github.com/CenterForOpenScience/ember-osf/pull/297).)

## Summary of Changes

1. Remove `validated-input` component and its related tests. (The component is moved to `ember-osf` repo in the first PR.)

2. Respective changes are made to the handlebar file and the controller to make sure it works.

3. Added a style for datepicker in `app.scss`.

None

## Ticket

https://openscience.atlassian.net/browse/IN-78

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/validated-date-input/component.js
```diff
@@ -0,0 +1,7 @@
+import ValidatedInput from '../validated-input'
+
+export default ValidatedInput.extend({
+    init() {
+        this._super(...arguments);
+    },
```
jamescdavis: This is not necessary.

app/controllers/submit.js
```diff
@@ -357,8 +369,16 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
         return changed;
     }),
 
+    basicsOriginalPublicationDate: Ember.computed.or('model.originalPublicationDate'),
```
jamescdavis: I'm not sure why you're using the computed or macro here. Maybe I'm just not following the logic?

app/components/validated-date-input/template.hbs
```diff
@@ -0,0 +1,20 @@
+<div class="form-group">
+    {{yield}}
+    {{#if large}}
+        {{bootstrap-datepicker type=type value=value placeholder=placeholder class="form-control" name=valuePath}}
+    {{else}}
+        {{bootstrap-datepicker type=type value=value placeholder=placeholder class="form-control" name=valuePath}}
+    {{/if}}
```
jamescdavis: No difference?

jamescdavis:
There's also a conflict in the translations file, but it's easy to resolve.

adlius:
@jamescdavis PR is updated. Moved the `validated-input` component to the `ember-osf` repo.

app/controllers/submit.js
adlius: After our discussion last week, I've tried to initialize the `basicsOriginalPublicationDate` and `basicsDOI` within the `clearFields()`, which would be invoked upon controller setup. However, turns out that `clearFields()` would only be called if `model.isLoaded` is set, therefore would not successfully initialize. I decided to keep it this way as it is kind of a pattern and also it works. Tell me what you think.

jamescdavis: I'm ok with this as long as you add a comment above it explaining what's going on. I think a computed alias would also be clearer than a computed or.

app/styles/app.scss
```diff
@@ -1458,3 +1458,7 @@ nav.branded-navbar {
         padding-left: 40px;
     }
 }
+
+.datepicker.dropdown-menu {
+    background-color:#FFFFFF;
+}
```
jamescdavis: We should set this in the scss for the validated-input component in ember-osf.

app/templates/submit.hbs
```diff
@@ -176,7 +182,7 @@
                                             <span class="required">{{t "global.abstract"}}:</span>
                                         </label>
                                         <form onchange={{if editMode (action 'track' 'input' 'onchange' 'Edit - Abstract Text Change') (action 'track' 'input' 'onchange' 'Submit - Abstract Text Change')}}>
-                                            {{validated-input model=this valuePath='basicsAbstract' placeholder=(t "submit.body.basics.abstract.placeholder") value=basicsAbstract large=true}}
+                                            {{validated-input model=this valuePath='basicsAbstract' placeholder=(t "submit.body.basics.abstract.placeholder") value=basicsAbstract inputType='textarea'}}
```
jamescdavis: Can you make inputType the first parameter so it's really obvious that this will render as a textarea?

tests/unit/controllers/submit-test.js
```diff
@@ -174,6 +176,15 @@ test('doiValid computed property', function(assert) {
...
     assert.equal(ctrl.get('doiValid'), true);
 });
+test('originalPublicationDateValid computed property', function(assert) {
+    const ctrl = this.subject();
+    let tomorrow = moment().add(1, 'days');
+    let yesterday = moment().subtract(1, 'days');
+    ctrl.set('basicsOriginalPublicationDate', tomorrow);
+    assert.equal(ctrl.get('originalPublicationDateValid'), false);
+    ctrl.set('basicsOriginalPublicationDate', yesterday);
+    assert.equal(ctrl.get('originalPublicationDateValid'), true);
```
jamescdavis: Maybe also `today`?

app/styles/app.scss
adlius: Problem is that the bootstrap is not rendered inside the "validated-input" element. It is actually rendered in the body of the page so the css for the validated-input component does not work on it.

adlius:
Comment is added.

jamescdavis:
ok, https://github.com/CenterForOpenScience/ember-osf/pull/297 is merged. Please update to latest ember-osf.

CHANGELOG.md
```diff
@@ -9,11 +9,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Headless Firefox
+<<<<<<< HEAD
```
baylee-d: You missed a spot ^

CHANGELOG.md
```diff
@@ -9,11 +9,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - Auto-expansion on selected subjects on the Discover page
 - `noscript` message if JavaScript is disabled
 - Favicons are updated to reflect their provider, on a branded domain or on a /preprints/provider page
 - DOI message, used on the preprint detail page to show users when they will have a DOI for their preprint.
 - Show a message on detail page when a DOI is created but not yet minted that it takes up to 24 hours to be minted.
+- Original Publication Date is added to preprint.
+  - `validated-input` component is moved to ember-osf repo
```
baylee-d: Please remove the space before the `-` to make it line up with the ones on top

tests/unit/controllers/submit-test.js
```diff
@@ -182,10 +196,12 @@ test('basicsValid computed property', function(assert) {
     ctrl.set('abstractValid', false);
     ctrl.set('doiValid', false);
     ctrl.set('licenseValid', false);
+    ctrl.set('originalPublcationDateValid', false);
```
baylee-d: Forgot the i in `originalPublicationDateValid`

CHANGELOG.md
adlius: The second line was indented because it was part of the change described in the first one. I don't want to put it as two separate items because the second line is actually a removal instead of an addition. But moving it to "Removed" makes it look like two unrelated things.

jamescdavis: I think putting this in the "Removed" section makes sense since moving `validated-input` to ember-osf wasn't really a dependency, just something else you did in this ticket. They are related, but not in a way that is probably that important to the reader of the changelog. More important is the fact that something was removed from this app.

app/styles/app.scss
jamescdavis: You can still move it to the SCSS file for the validated-input component, just leave it generic, like you have it, instead of putting it under the component div's class.

laurenbarker:
@jamescdavis it looks like the conflicts were resolved. Is this ready?

PR #480, opened 2017-11-02
[EOSF-783] Warning when leaving submission/edit page

baylee-d:
## Purpose

When leaving the submit/edit page, there should be a warning to stop users from leaving if there are unsaved changes.  The purpose of this PR is to add a dialogue box that pops up when the user tries to leave or refresh the page in the case that there are unsaved changes.

## Summary of Changes

- Added `ember-onbeforeunload` to handle transition checks on the submit and edit routes
- Re-implemented the `disciplineChanged` computed function, which checks if there have been user changes on the discipline list.
- Added a `computed` function to check the changed fields (`uploadChanged`, `basicsChanged` and `disciplineChanged`).  The `computed` is used to determine whether or not the page is dirty and if the dialogue box will show on redirect.

## Ticket

https://openscience.atlassian.net/browse/EOSF-783

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -178,6 +178,8 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
 
     hasFile: Ember.computed.or('file', 'selectedFile'),
 
+    hasDirtyFields: Ember.computed.or('uploadChanged', 'basicsChanged', 'disciplineChanged'), // True when fields have been changed
+
     clearFields() {
         // Restores submit form defaults.  Called when user submits preprint, then hits back button, for example.
         this.get('panelActions').open('Upload');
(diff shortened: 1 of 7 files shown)


app/routes/content/edit.js
```diff
@@ -14,10 +15,11 @@ import permissions from 'ember-osf/const/permissions';
  * Fetches current preprint. Redirects to preprint provider route if necessary.
  * @class Edit Route Handler
  */
-export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitControllerMixin, {
+export default Ember.Route.extend(ConfirmationMixin, Analytics, ResetScrollMixin, SetupSubmitControllerMixin, {
     theme: Ember.inject.service(),
     headTagsService: Ember.inject.service('head-tags'),
     currentUser: Ember.inject.service('currentUser'),
+    confirmationMessage: 'Are you sure you want to abandon changes to this preprint?', // Confirmation message when navigating away from preprint
```
sheriefvt: can you use the i18n service here

app/controllers/submit.js
```diff
@@ -374,6 +404,12 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
         return Ember.$.extend(true, [], this.get('model.subjects')).reduce((acc, val) => acc.concat(val), []).uniqBy('id');
     }),
 
+    // Compares the model's and current subjectLists's array of arrays of discipline ids
+    // to determine if there has been a change.
+    disciplineChanged: Ember.computed('model.subjects.@each.subject', 'subjectsList.@each.subject', 'disciplineModifiedToggle', function () {
+        return !this.disciplineArraysEqual(subjectIdMap(this.get('model.subjects')), subjectIdMap(this.get('subjectsList')));
```
jamescdavis: I think there is a simpler way to do this. What about?:
```
return JSON.stringify(this.get('model.subjects')) !== JSON.stringify(this.get('subjectsList'));
```

app/routes/submit.js
```diff
@@ -35,14 +41,21 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, CasAuthenticatedR
...
-    actions: {
-        willTransition: function(transition) {
-            // Displays confirmation message if user attempts to navigate away from Add Preprint process with unsaved changes
-            var hasFile = this.controller.get('file') !== null || this.controller.get('selectedFile') !== null;
+    isPageDirty() {
+        // If true, shows a confirmation message when leaving the page
+        // True if the user already created/chosen a project node
+        return this.controller.get('hasDirtyFields');
+    },
+    shouldCheckIsPageDirty(transition) {
```
jamescdavis: This method is confusingly named. One would think is should return false to skip the isPageDirty check, but you actually return true (as you have done and the docs indicate). They should have named this `shouldSkipIsPageDirty`.

PR #487, opened 2017-11-10
[EOSF-881] Fix contrast issue for error pages.

adlius:
## Purpose

Currently, the color contrast for support email link on ECSarxiv and EarthArxiv error pages are not enough. This PR fix the issue.

## Summary of Changes

Change the component's CSS class to 'prerpint-header' and 'preprint-hearder-error'.
Added a style file to the component to add padding and bottom margin.

The error page for OSF preprint would also be changed to using the background image on the landing page, which should be more consistent.

## Ticket

https://openscience.atlassian.net/browse/EOSF-881

# Reviewer Checklist

- [x] meets requirements
- [x] easy to understand
- [x] DRY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - component integration tests to work in Firefox
   - provider-carousel
   - search-facet-taxonomy
+- CSS contrast issue for support email link on some branded provider error pages.
 
 ## [0.115.1] - 2017-11-07
 ## Changed
--- a/app/components/error-page.js
+++ b/app/components/error-page.js
@@ -3,7 +3,7 @@ import Analytics from 'ember-osf/mixins/analytics';
 
 export default Ember.Component.extend(Analytics, {
     theme: Ember.inject.service(),
-    classNames: ['preprint-error-page'],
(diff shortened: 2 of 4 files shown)


adlius:
CHANGELOG.md is modified.

hmoco:
This looks good. One small concern, and this seems to be intended, but if so we should be more explicit about it. Replacing `preprint-error-page` does not allow brands to change their color here. The PR description implies that that is intended. However, `preprint-error-page` is still a class in the https://github.com/CenterForOpenScience/ember-osf-preprints/blob/develop/app/styles/brands/_brand.scss file. My opinion is that if we intend to remove that option, the class should also be removed from the customized .css brand files. Opinions? @jamescdavis @adlius

adlius:
You're right. The side effect is that the brands cannot customize how they want the error page to look like without also changing the header elements. This should be more favorable as it results in more consistent looks. If that is the case, I can go and remove that from the CSS files for brands.

hmoco:
I honestly prefer the button solution, to keep brands' themes consistent. However, that is inconsistent UI with our other support links, inconsistent UX to what buttons normally do versus anchor tags, and it might look ugly. Also, its not like all our brands have great themes (__cough__ __cough__ agrixiv). So, I'm happy with just removing the CSS pertinent to error pages from those files and the brand template.

PR #492, opened 2017-11-30
[IN-70][Preprints] Add whitelist functionality for preprint discover page

adlius:
## Purpose

This is the first of two PRs for the preprint discover page to get whitelisted SHARE providers from `/preprint_providers` endpoint of API V2. (The corresponding backend ticket is [OSF-8596](https://openscience.atlassian.net/browse/OSF-8595)).

## Summary of Changes

In `routes/discover.js`, the model hook is modified to use `query()`, as `findAll()` does not return API data with top level metadata. The route then set the controller's `meta` property to allow use of metadata in the controller.

In `controllers/discover.js`, a new computed property `whiteListedProviders` is set to get whitelist provider info from metadata. The computed property would be passed as parameters down to `search-facet-provider` component. (The actual passing of parameters in templates are in the [second PR](https://github.com/CenterForOpenScience/ember-osf/pull/321) to ember-osf repo where the components are located.)

None.

## Ticket

https://openscience.atlassian.net/browse/IN-70

# Reviewer Checklist

- [x] meets requirements
- [x] easy to understand
- [x] DRY
- [x] testable and includes test(s)


app/routes/discover.js
```diff
@@ -20,7 +20,16 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, {
...
+            .query('preprint-provider', { reload: true }).then((results) => {
+                return {
+                    preprintProviders: results,
+                    meta: results.get('meta')
+                };
+            });
+    },
+    setupController(controller, { preprintProviders, meta }) {
+        this._super(controller, preprintProviders);
+        controller.set('meta', meta);
```
hmoco: Is this necessary? Wouldn't you have access to this in the controller through `model.meta`?

app/controllers/discover.js
```diff
@@ -108,7 +107,9 @@ export default Ember.Controller.extend(Analytics, {
         return themeProvider;
     }),
     type: '', //Type query param. Must be passed to component, so can be reflected in URL
-    whiteListedProviders: config.whiteListedProviders,
+    whiteListedProviders: Ember.computed(function() {
+       return this.get('meta.whitelisted_providers');
```
hmoco: :+1:

app/routes/discover.js
adlius: According to Ember official guide, to access top-level metadata outside of model hook we have to set it for the controller. (https://guides.emberjs.com/v2.17.0/models/handling-metadata/)

hmoco: Had never seen this, and assumed that it worked like RSVP.hash, where the model became an object and meta would just be another key in it.

hmoco:
Blocked by https://github.com/CenterForOpenScience/osf.io/pull/7926/. Ready to merge once that is released.

PR #475, opened 2017-10-25
[Improvement][MOD-159] Add DOI message - Preprints

sheriefvt:
## Problem
Users don't know when they will have a DOI for their preprint.

## Solution
Add "DOI created after moderator approval." on pending pre-moderation preprints

image
image

## Ticket
https://openscience.atlassian.net/browse/MOD-159

PR #502, opened 2017-12-15
[Feature][EOSF-959] Add message about when DOI will be minted

sheriefvt:
## Purpose
Add message about when DOI will be minted. Users will want to know when they will have a DOI for their preprint.

follow-up PR for previous PRs #475 and #501 

## Summary of Changes
if DOI: display DOI
else if node.private: "DOI created after preprint is made public"
else if workflow and !isPublished: "DOI created after moderator approval"
else: display nothing

* Add unit tests for preprint-status-banner

## Ticket

https://openscience.atlassian.net/browse/EOSF-959

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `noscript` message if JavaScript is disabled
 - Favicons are updated to reflect their provider, on a branded domain or on a /preprints/provider page
 - DOI message, used on the preprint detail page to show users when they will have a DOI for their preprint.
+- Show a message on detail page when a DOI is created but not yet minted that it takes up to 24 hours to be minted.
 
 ### Changed
 - Use yarn --frozen-lockfile instead of --pure-lockfile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- Favicons are updated to reflect their provider, on a branded domain or on a /preprints/provider page
 
 ### Added
 - Headless Firefox
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Wording on `Edit preprint` button to `Edit and resubmit` on preprint detail page if the preprint is rejected
  and the workflow is pre-moderation.
 - Removed footer styling
+- Updated `meta.total` to `meta.total_pages` in preprint-form-authors component
+- Modified `preprint-status-banner` component for review action rename.
 
 ### Fixed
 - component integration tests to work in Firefox
   - provider-carousel
   - search-facet-taxonomy
+- CSS contrast issue for support email link on some branded provider error pages.
+
+## [0.115.9] - 2017-12-19
+### Added
+- Settings for prerender
 
 ## [0.115.8] - 2017-12-07
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Favicons are updated to reflect their provider, on a branded domain or on a /preprints/provider page
 - DOI message, used on the preprint detail page to show users when they will have a DOI for their preprint.
 - Show a message on detail page when a DOI is created but not yet minted that it takes up to 24 hours to be minted.
+- Confirmation message when user tries to reload or navigate away from the preprint submission page after adding/selecting a node
+- `Computed` to check if there has been user changes on the discipline field on the preprint submission/edit form
 
 ### Changed
 - Use yarn --frozen-lockfile instead of --pure-lockfile
--- a/app/components/preprint-status-banner/component.js
+++ b/app/components/preprint-status-banner/component.js
@@ -61,11 +61,11 @@ export default Ember.Component.extend({
         if (this.get('submission.provider.reviewsCommentsPrivate')) {
             this.set('latestAction', null);
         } else {
-            const submissionActions = this.get('submission.actions');
+            const submissionActions = this.get('submission.reviewActions');
             if (submissionActions) {
-                submissionActions.then(actions => {
-                    if (actions.length) {
-                        this.set('latestAction', actions.get('firstObject'));
+                submissionActions.then(reviewActions => {
+                    if (reviewActions.length) {
+                        this.set('latestAction', reviewActions.get('firstObject'));
                     } else {
                         this.set('latestAction', null);
                     }
--- a/app/components/preprint-status-banner/component.js
+++ b/app/components/preprint-status-banner/component.js
@@ -57,17 +57,22 @@ export default Ember.Component.extend({
             CLASS_NAMES[this.get('submission.reviewsState')];
     }),
 
-    didInsertElement() {
+    didReceiveAttrs() {
         if (this.get('submission.provider.reviewsCommentsPrivate')) {
             this.set('latestAction', null);
         } else {
-            this.get('submission.actions').then(actions => {
-                if (actions.length) {
-                    this.set('latestAction', actions.get('firstObject'));
-                } else {
-                    this.set('latestAction', null);
-                }
-            });
+            const submissionActions = this.get('submission.actions');
+            if (submissionActions) {
+                submissionActions.then(actions => {
+                    if (actions.length) {
+                        this.set('latestAction', actions.get('firstObject'));
+                    } else {
+                        this.set('latestAction', null);
+                    }
+                });
+            } else {
+                this.set('latestAction', null);
+            }
         }
     },
 
--- a/app/components/preprint-status-banner/component.js
+++ b/app/components/preprint-status-banner/component.js
@@ -57,7 +57,7 @@ export default Ember.Component.extend({
             CLASS_NAMES[this.get('submission.reviewsState')];
     }),
 
-    didReceiveAttrs() {
+    didInsertElement() {
         if (this.get('submission.provider.reviewsCommentsPrivate')) {
             this.set('latestAction', null);
         } else {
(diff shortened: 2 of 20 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- Favicons are updated to reflect their provider, on a branded domain or on a /preprints/provider page
```
aaxelb: I don't think we should delete this

app/locales/en/translations.js
```diff
@@ -58,6 +58,8 @@ export default {
         version: 'Version',
         preprint_doi: `{{preprintWords.Preprint}} DOI`,
         article_doi: `Peer-reviewed Publication DOI`,
+        preprint_pending_doi: `DOI created after preprint is made public`,
```
aaxelb: `preprintWords.preprint`

app/templates/content/index.hbs
```diff
@@ -90,17 +90,23 @@
                         </button>
                     </div>
 
-                    {{#if model.links.preprint_doi}}
-                        <div class="p-t-xs">
-                            <h4 class="p-v-md f-w-md"><strong>{{t "content.preprint_doi"}}</strong></h4>
+                    <div class="p-t-xs">
+                        <h4 class="p-v-md f-w-md"><strong>{{t "content.preprint_doi"}}</strong></h4>
+                        {{#if model.links.preprint_doi}}
```
aaxelb: Would you mind updating this section to use `model.preprintDoiUrl` instead of `model.links.preprint_doi`?

...and if you're not opposed to gentle scope creep, updating the next section to use `model.articleDoiUrl` instead of `model.doi` and `doiUrl`?

app/components/preprint-status-banner/component.js
```diff
@@ -57,7 +57,7 @@ export default Ember.Component.extend({
             CLASS_NAMES[this.get('submission.reviewsState')];
     }),
 
-    didReceiveAttrs() {
+    didInsertElement() {
```
aaxelb: `didInsertElement` is generally used for things that mess with the DOM, which doesn't seem to be the case here.

tests/unit/components/preprint-status-banner-test.js
```diff
@@ -0,0 +1,203 @@
...
+        component.set('submission.reviewsState', 'pending');
+        assert.strictEqual(component.get('getClassName'), 'preprint-status-pending-pre');
+
+        component.set('submission.reviewsState', 'accepted');
+        assert.strictEqual(component.get('getClassName'), 'preprint-status-accepted');
+    });
+});
+
+// Stub i18n service
+const i18nStub = Ember.Service.extend({
```
aaxelb: Are translations unavailable during tests? Why do you need to stub `i18n`?

CHANGELOG.md
jamescdavis: This line was the result of a dirty merge conflict resolution and shouldn't be here. It's already included under the Added section below. Can either delete it here or I will clean it up when I release.

app/templates/content/index.hbs
jamescdavis: 👍

tests/unit/components/preprint-status-banner-test.js
jamescdavis: Since this is a unit test for the component and we have extended i18n, I think it's actually a good idea to stub i18n here so that these tests don't rely on our extension of i18n working correctly. If this were an integration test, I would agree that it should just use our extended i18n.

app/locales/en/translations.js
jamescdavis: 👍

app/components/preprint-status-banner/component.js
```diff
@@ -61,13 +61,18 @@ export default Ember.Component.extend({
...
         } else {
-            this.get('submission.actions').then(actions => {
-                if (actions.length) {
-                    this.set('latestAction', actions.get('firstObject'));
-                } else {
-                    this.set('latestAction', null);
-                }
-            });
+            const submissionActions = this.get('submission.actions');
+            if (submissionActions) {
```
aaxelb: Is `submission.actions` null sometimes? I'm curious when that happens.

app/templates/content/index.hbs
```diff
@@ -90,19 +90,25 @@
...
                     </div>
 
-                    {{#if model.links.preprint_doi}}
-                        <div class="p-t-xs">
-                            <h4 class="p-v-md f-w-md"><strong>{{t "content.preprint_doi"}}</strong></h4>
+                    <div class="p-t-xs">
+                        <h4 class="p-v-md f-w-md"><strong>{{t "content.preprint_doi"}}</strong></h4>
+                        {{#if model.preprintDoiUrl}}
                             {{#if model.preprintDoiCreated}}
                                 <a href={{model.links.preprint_doi}} onclick={{action "click" "link" "Content - Preprint DOI" model.links.preprint_doi}}>{{extract-doi model.links.preprint_doi}}</a>
```
aaxelb: When I mentioned updating the whole section to use `preprintDoiUrl`, I meant the code you didn't write, too, so it's consistent. If you don't mind?

app/templates/content/index.hbs
```diff
@@ -90,19 +90,25 @@
...
                             {{else}}
                                 <p> {{extract-doi model.links.preprint_doi}} </p>
```
aaxelb: `preprintDoiUrl`

app/templates/content/index.hbs
```diff
@@ -90,19 +90,25 @@
...
+                                {{t 'content.preprint_pending_doi_moderation'}}
+                            {{/if}}
+                        {{/if}}
+                    </div>
 
-                    {{#if model.doi}}
+                    {{#if model.articleDoiUrl}}
                         <div class="p-t-xs">
                             <h4 class="p-v-md f-w-md"><strong>{{t "content.article_doi"}}</strong></h4>
                             <a href={{doiUrl}} onlick={{action "click" "link" "Content - DOI" doiUrl}}>{{model.doi}}</a>
```
aaxelb: `articleDoiUrl` instead of `doiUrl` and `model.doi`

app/components/preprint-status-banner/component.js
sheriefvt: the main reason is to make the code easier to test, previously unit tests were failing with submission.actions is not defined.  Also assuming that actions are correctly retrieved might possibly introduce error in the future.

aaxelb: `submission.actions` should not be undefined... `hasMany` relations on models should return a promise-like object. Did you include `model:action` in the `needs` for that test?

If there's an error retrieving, the returned promise-like object will reject/fail. The way to handle that would be a `.catch()`.

jamescdavis: What about when `submission` is undefined? That's what's happening in the unit test and theoretically could happen in real-world use of this component depending on where the `submission` parameter comes from. The component shouldn't make an assumption that `submission` is always defined (or that it's necessarily a preprint if we want to make it really robust).

aaxelb: Hmm, fair point. In practice it doesn't make sense to even render this component without an actual `submission`, but you're right it should be able to handle that case.

app/templates/content/index.hbs
```diff
@@ -90,22 +90,28 @@
...
+                            {{/if}}
+                        {{/if}}
+                    </div>
 
-                    {{#if model.doi}}
+                    {{#if model.articleDoiUrl}}
                         <div class="p-t-xs">
                             <h4 class="p-v-md f-w-md"><strong>{{t "content.article_doi"}}</strong></h4>
-                            <a href={{doiUrl}} onlick={{action "click" "link" "Content - DOI" doiUrl}}>{{model.doi}}</a>
+                            <a href={{model.articleDoiUrl}} onlick={{action "click" "link" "Content - DOI" model.articleDoiUrl}}>{{model.articleDoiUrl}}</a>
```
aaxelb: The text here should maybe be `{{extract-doi model.articleDoiUrl}}`, to be consistent with the previous section and what was displayed before.

jamescdavis: This is not a bad idea, but also note that the `extractDoiFromString` utility (which the extract-doi helper wraps) is [used to sanitize user input](https://github.com/CenterForOpenScience/ember-osf-preprints/blob/develop/app/controllers/submit.js#L792). It's always good to make sure, though, especially since the API doesn't enforce this and we may have bulk uploads in the near future. We may want to rename `model.articleDoiUrl` because it is actually just a bare DOI and not a full URL.

jamescdavis: Also, I just noticed that you are switching the link from `doiUrl` to `model.articleDoiUrl`. This will result in a bad link. `doiUrl` is a computed that [makes the bare DOI into a URL](https://github.com/CenterForOpenScience/ember-osf-preprints/blob/develop/app/controllers/content/index.js#L186).

aaxelb: Looking at [the API](https://api.osf.io/v2/preprints/), `model.articleDoiUrl` (aka `model.links.doi`) is the full URL, while `model.doi` is just the bare DOI.

jamescdavis: Wait, now I'm confused, we a doi attribute AND a doi link?

aaxelb: Yep! Good times. That confusion is why I'd rather not use either, and stick to `articleDoiUrl`

jamescdavis: If the API is giving us the full url, we should definitely use that, but we should also then remove the `doiUrl` computed.

aaxelb: [Totally agree](https://github.com/CenterForOpenScience/ember-osf-preprints/pull/502/files#diff-1940ae5c2ed759c5fcdbeea83483b130L185)

jamescdavis: Nevermind, ignore everything I've said here. :)

jamescdavis: Only question is whether to use `{{extract-doi model.articleDoiUrl}}` or `{{model.doi}}` for the text since they should be equivalent.

aaxelb: Yeah, either's valid. But because it's so confusingly named, I lean away from using `model.doi` at all ever.

I wouldn't be opposed to renaming the ember model field to `articleDoi`, though...

sheriefvt:
@jamescdavis is this was caused by running yarn instead of yarn --pure-lockfile

PR #508, opened 2018-01-02
[IN-79][Preprints] Add auto expansion for top level taxos when there are less than 3.

adlius:
## Purpose

Auto-expand top levels (on preprint discover page) when there are <= 3 top level taxonomies. Additionally, if there are <= 3 top level taxonomies, auto-expand secondary taxonomies when there are <= 3 of secondary taxos for each of the top level taxos.

## Summary of Changes

Add a `_expandDefault()` function to be invoked in `init()`.
Additionally, removed `themeStub` service in `search-facet-taxonomy-tests.js` as it creates some unintended side effects with auto-expansion of taxonomies.

For testing purposes, four json files are included in the JIRA ticket attachment so that you may import them from the admin panel to setup preprint providers with different taxonomies. Also, please check that auto-expansion doe not happen when user goes to the discover page by clicking one of the highlighted taxonomies from the landing page.

## Ticket

https://openscience.atlassian.net/browse/IN-79

# Reviewer Checklist

- [x] meets requirements
- [x] easy to understand
- [x] DRY

### DIFF
--- a/app/components/search-facet-taxonomy.js
+++ b/app/components/search-facet-taxonomy.js
@@ -86,6 +86,7 @@ export default Ember.Component.extend(Analytics, {
                     })
                 });
                 this._expandMany(items);
+                this._expandDefault();
             });
     },
     actions: {
@@ -97,6 +98,20 @@ export default Ember.Component.extend(Analytics, {
                     label: `Discover - ${item.text}`
                 });
             this._expand(item);
+        }
+    },
+    _expandDefault(){
+        let topLevelItem = this.get('topLevelItem');
+        if (topLevelItem.length <= 3){
+            topLevelItem.forEach(item => {
+                this._expand(item).then(() =>{
+                    if (item.childCount <= 3){
+                        item.children.forEach(item => {
+                           this._expand(item);
+                        });
+                    }
+                });
+            });
         }
     },
     _expand(item) {
--- a/app/components/search-facet-taxonomy.js
+++ b/app/components/search-facet-taxonomy.js
@@ -86,7 +86,9 @@ export default Ember.Component.extend(Analytics, {
                     })
                 });
                 this._expandMany(items);
-                this._expandDefault();
+                //Only auto-expand if no subjects are selected.
+                if (!items.length > 0)
+                    this._expandDefault();
             });
     },
     actions: {
--- a/app/components/search-facet-taxonomy.js
+++ b/app/components/search-facet-taxonomy.js
@@ -102,11 +102,11 @@ export default Ember.Component.extend(Analytics, {
             this._expand(item);
         }
     },
-    _expandDefault(){
+    _expandDefault() {
         let topLevelItem = this.get('topLevelItem');
         if (topLevelItem.length <= 3){
             topLevelItem.forEach(item => {
-                this._expand(item).then(() =>{
+                this._expand(item).then(() => {
                     if (item.children && item.childCount <= 3){
                         item.children.forEach(item => {
                            this._expand(item);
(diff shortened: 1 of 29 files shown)


app/components/search-facet-taxonomy.js
```diff
@@ -86,6 +86,7 @@ export default Ember.Component.extend(Analytics, {
                     })
                 });
                 this._expandMany(items);
```
alexschiller: There is definitely the opportunity for conflict between the expanding done by this line and your expandDefault. 
For example, I tested clicking a 3rd level subject and then refreshed the page. Instead of showing the selected subject it only expanded out two levels. Swapped back to develop and it expanded out for my clicked subject. 

This may require a product for to answer if they possibly want one or the other to occur (e.g., expand the clicked items or expand the default items) or if they always want both, in which case a non-toggle function will need to be written that always expands. 

If they want only one or the other, you can probably check if anything is in items and if not, expandDefault.

app/components/search-facet-taxonomy.js
```diff
@@ -99,6 +100,20 @@ export default Ember.Component.extend(Analytics, {
             this._expand(item);
         }
     },
+    _expandDefault(){
```
alexschiller: Minor:
Not sure what the es6 linting will suggest, but there is some inconsistent paren spacing. I think it is supposed to be `fun() {` instead of `fun(){` but I actually don't know which we want to do, either way consistency is better.

app/components/search-facet-taxonomy.js
```diff
@@ -99,6 +100,20 @@ export default Ember.Component.extend(Analytics, {
...
+        let topLevelItem = this.get('topLevelItem');
+        if (topLevelItem.length <= 3){
+            topLevelItem.forEach(item => {
```
alexschiller: I think I like the `=> {` better than `=>{` below, nit picky things, but might as well fix while in there.

app/components/search-facet-taxonomy.js
```diff
@@ -86,6 +86,9 @@ export default Ember.Component.extend(Analytics, {
                     })
                 });
                 this._expandMany(items);
+                //Only auto-expand if no subjects are selected.
+                if (!items.length > 0)
```
baylee-d: A couple of things here.  

1. Would it be better to write this out as `if (items.length === 0)`.  It might make it a bit clearer to other people reading it out.

2. I think the preferred way to write this statement out would be with the curly braces.

app/components/search-facet-taxonomy.js
```diff
@@ -99,6 +102,20 @@ export default Ember.Component.extend(Analytics, {
             this._expand(item);
         }
     },
+    _expandDefault() {
+        let topLevelItem = this.get('topLevelItem');
+        if (topLevelItem.length <= 3){
```
baylee-d: This will fail eslint.  Be sure to put a space before the `{`

app/components/search-facet-taxonomy.js
```diff
@@ -99,6 +102,20 @@ export default Ember.Component.extend(Analytics, {
...
+            topLevelItem.forEach(item => {
+                this._expand(item).then(() => {
+                    if (item.children && item.childCount <= 3){
```
baylee-d: Space here too please

tests/integration/components/search-facet-taxonomy-test.js
```diff
@@ -55,4 +64,6 @@ test('One-level hierarchy taxonomies', function(assert) {
     render(this);
     assert.equal(this.$('label')[0].innerText.trim(), 'Arts and Humanities');
     assert.equal(this.$('label')[1].innerText.trim(), 'Education');
+    assert.equal(this.$('label')[2].innerText.trim(), 'Filmography');
```
baylee-d: Do these tests do anything different than the previously existing ones?

alexschiller:
Also @adlius, please add testing considerations and make sure you note that auto expand only occurs when no taxonomy items are pre-selected. I'd also add that Paleorxiv is a good test case for this.

PR #517, opened 2018-01-22
[IN-23][Preprints] Change to use fullName for preprint detail page.

adlius:
## Purpose

Currently, if a user does not specify a Family name and a Given name in their profile settings, their name would not appear on the preprint detail page. This PR addresses the problem by using their full name, which should not be empty.

## Summary of Changes

Change to use `fullName`.

None

## Ticket

https://openscience.atlassian.net/browse/IN-23

# Reviewer Checklist

- [x] meets requirements
- [x] easy to understand
- [x] DRY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Show `fullname` for contributors on preprint detail page
 
 ## [0.116.3] - 2018-01-12
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,29 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Auto-expansion on selected subjects on the Discover page
+- `noscript` message if JavaScript is disabled
+- Favicons are updated to reflect their provider, on a branded domain or on a /preprints/provider page
+- DOI message, used on the preprint detail page to show users when they will have a DOI for their preprint.
+- Show a message on detail page when a DOI is created but not yet minted that it takes up to 24 hours to be minted.
+- Confirmation message when user tries to reload or navigate away from the preprint submission page after adding/selecting a node
+- `Computed` to check if there has been user changes on the discipline field on the preprint submission/edit form
+- Translations for `Browse by featured subjects`, `Browse by subjects`, and `Browse by providers`
+
+### Changed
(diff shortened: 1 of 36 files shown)


adlius:
CHANGELOG is added.

app/templates/content/index.hbs
```diff
@@ -14,7 +14,7 @@
                                         {{#if author.unregisteredContributor}}
                                             <li>{{author.unregisteredContributor}}</li>
                                         {{else}}
-                                            <li><a href={{author.users.profileURL}} onclick={{action 'click' 'link' 'Content - Author' author.users.profileURL}}>{{author.users.givenName}} {{author.users.familyName}}</a></li>
+                                            <li><a href={{author.users.profileURL}} onclick={{action 'click' 'link' 'Content - Author' author.users.profileURL}}>{{author.users.fullName}}</a></li>
```
alexschiller: After some pondering, I had a quick product consult and I think there is a slightly more ideal solution:

1. Add a helper to return Given + Family Name if both exist, otherwise return Full name
2. Add tests for the helper

The reasoning is that Full name can contain a multitude of other information. Some people add other things to their fullname like "Alexander J. Schiller, PhD" which adds a fair amount of unnecessary clutter compared to "Alexander Schiller." It is safer to keep everyone the same and only use the subpar Full name when no given/family name exist. 

Let me know if you have any questions about that.

adlius:
A helper is added to concatenate givenName and familyName if they are not undefined. Otherwise, we fall back to using fullName. Tests for the helper are also added.

app/helpers/contributor-concat-name.js
```diff
@@ -0,0 +1,11 @@
+import Ember from 'ember';
+
+export function contributorConcatName([givenName, familyName, fullName]) {
```
jamescdavis: This probably makes more sense as a computed property of user. Perhaps just call it `name`? Then you can just say `{{author.users.name}}` in the template and the logic is in the user model.

app/helpers/contributor-concat-name.js
```diff
@@ -0,0 +1,11 @@
...
+    if (givenName && familyName){
+        return [givenName, familyName].join(' ');
```
jamescdavis: I think
```
return `${givenName} ${familyName}`;
```
is probably more readable.

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Get pagination info directly from meta in preprint-form-authors component
 - Check for `has_highlighted_subjects` flag to determine what wording should be used on index page
 
+### Fixed
+- Show `fullname` for contributors on preprint detail page
```
alexschiller: Also changelog may need to be slightly modfied.

adlius:
This PR is closed. Two new PRs are created for `ember-osf` and `ember-osf-preprints`.
https://github.com/CenterForOpenScience/ember-osf/pull/346
https://github.com/CenterForOpenScience/ember-osf-preprints/pull/521

PR #519, opened 2018-01-25
[IN-67][Preprints]  Add conditional to prevent redirect at any point in preprint submission

baylee-d:
## Purpose

Users can still redirect away from the preprint submission page when clicking `Save and Continue` after each form block.

## Summary of Changes

Added to the already existing `computed.or` to determine if the user is adding or editing a preprint.  Will default to `True` if adding.

This will prevent the user from leaving the page at all when uploading and creating a preprint.  The user shouldn't be able to redirect when:
- In the middle of editing a section
- After clicking `Save and Continue` between sections (even when nothing else has been started/changed)

When the user is editing a preprint, the user can leave the page unless they have any dirty fields.  So the user should not be able to redirect when:
- They've changed anything without hitting `Save and Continue`
They should be able to redirect when:
- They've changed things and have clicked `Save and Continue`

The idea being that a user can always come back after saving parts of their already existing preprint.  They cannot (easily) come back and continue creating a preprint in the middle of the upload process.

## Ticket

https://openscience.atlassian.net/browse/IN-67

# Reviewer Checklist

- [x] meets requirements
- [x] easy to understand
- [x] DRY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Confirmation message when user tries to reload or navigate away from the preprint submission page after adding/selecting a node
 - `Computed` to check if there has been user changes on the discipline field on the preprint submission/edit form
 - Translations for `Browse by featured subjects`, `Browse by subjects`, and `Browse by providers`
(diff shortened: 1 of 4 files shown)


fabmiz:
LGTM 🎉 🎉

alexschiller:
Please update to latest ember-osf develop and then this can be merged.

PR #525, opened 2018-02-06
[IN-45][Preprints] Truncate unicode strings correctly for Linkedin sharing.

adlius:
## Purpose

When a user clicks on the Linkedin sharing button on some preprints with Unicode encoded strings as titles or descriptions, the pop-up sharing page says "page not found". This is because Linkedin sharing works quite differently in that it gets sharing information from the query parameters of the sharing URL, which we construct on the front end side. Because javascript is not Unicode-aware, when we truncate Unicode strings according to specified length limit, it breaks the encoding and therefore causing errors for Linkedin when parsing the URL. 
This PR fix the problem by using an NPM package to do Unicode-aware string truncating.

## Summary of Changes

An NPM package `unicode-byte-truncate` is added as a dependency.
In `controllers/content/index.js`, `trunc` from the newly added dependency is used to truncate the title and description.

None.

## Ticket

https://openscience.atlassian.net/browse/IN-45

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "loader.js": "4.0.1",
     "pagination-pager": "2.4.2",
     "postcss": "5.1.0",
-    "unicode-byte-truncate": "^1.0.0",
+    "unicode-byte-truncate": "^1.0.0"
   },
   "bugs": {
     "url": "https://github.com/CenterForOpenScience/ember-osf-preprints/issues"
--- a/package.json
+++ b/package.json
@@ -96,5 +96,8 @@
   "bugs": {
     "url": "https://github.com/CenterForOpenScience/ember-osf-preprints/issues"
   },
-  "homepage": "https://github.com/CenterForOpenScience/ember-osf-preprints#readme"
+  "homepage": "https://github.com/CenterForOpenScience/ember-osf-preprints#readme",
+  "dependencies": {
+    "unicode-byte-truncate": "^1.0.0"
+  }
 }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -96,5 +96,8 @@
   "bugs": {
     "url": "https://github.com/CenterForOpenScience/ember-osf-preprints/issues"
   },
-  "homepage": "https://github.com/CenterForOpenScience/ember-osf-preprints#readme"
+  "homepage": "https://github.com/CenterForOpenScience/ember-osf-preprints#readme",
+  "dependencies": {
+    "unicode-byte-truncate": "^1.0.0"
```
jamescdavis: should go in devDependencies (see https://stackoverflow.com/questions/42159153/why-does-ember-installs-everything-as-devdependencies-instead-of-normal-dependen/42162627 for explanation of why this makes sense)

adlius:
Some simple tests are added.

package.json
```diff
@@ -91,10 +91,13 @@
...
-    "postcss": "5.1.0"
+    "postcss": "5.1.0",
+    "unicode-byte-truncate": "^1.0.0"
   },
   "bugs": {
     "url": "https://github.com/CenterForOpenScience/ember-osf-preprints/issues"
   },
-  "homepage": "https://github.com/CenterForOpenScience/ember-osf-preprints#readme"
+  "homepage": "https://github.com/CenterForOpenScience/ember-osf-preprints#readme",
+  "dependencies": {
```
baylee-d: If there isn't anything in `dependencies`, do we still need to include it in here?

alexschiller:
Merge conflicts

PR #553, opened 2018-03-29
[IN-91][Preprints] Update EmberCLI version [WIP]

baylee-d:
## Purpose

To update Preprints to use Ember 2.18

## Summary of Changes

- Update Ember to 2.18 and update other packages
- Remove almost all bower dependencies
- Update syntax to match 2.18 standards

## TODO

This is a work in progress.  Still to complete before it can be merged:

### Edit page:
- [ ] Fix `Upload` section to be editable (can open primary collapsible-panel but none of the secondary panels
- [ ] Fix `License` section to only include `Copyright holder` box if the license requires it (currently displays it in all cases and won't let user save until they put something inside it)
- [ ] Editing disciplines currently overrides previously selected disciplines.  Should only add to the array of current disciplines
- [ ] Moderation message at the bottom of the edit page is currently broken

### Detail page:
- [ ] Citation widget's placeholder has a broken translation

### Submit page:
- [ ] Stop the user from going between panels when they shouldn't be allowed to (ex. going from disciplines to provider carousel shouldn't be allowed, but currently is)
- [ ] Selected disciplines do not show when the panel is collapsed
- [ ] Copyright holder field shows up even when it isn't needed (and is required to save the section)
- [ ] Abstract field is too small
- [ ] License text doesn't show
- [ ] Selected basics fields do not show when the panel is collapsed
- [ ] Can create a preprint to branded providers, but fails to transition to the new preprint

### Routing:
- [ ] Can not transition to a branded provider.  Will put the user through an infinite transition loop.
- [ ] Can not go to a branded preprint detail page.  The route on transition adds the guid a second time at the end of the url and makes a broken link

This will need a full regression.

## Ticket

https://openscience.atlassian.net/browse/IN-91

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


package.json
```diff
@@ -19,84 +24,88 @@
...
+    "eslint": "^4.4.1",
+    "eslint-plugin-ember": "^5.0.0",
+    "liquid-fire": "^0.27.3",
+    "loader.js": "^4.2.3",
+    "postcss": "^5.1.0",
     "unicode-byte-truncate": "^1.0.0"
   },
+  "engineStrict": true,
+  "engines": {
+    "node": "^4.5 || 6.* || >= 7.* || >= 0.12.0"
```
laurenbarker: Drive by: we should set this to `>= 8`

app/components/search-preprints.js
```diff
@@ -1,4 +1,8 @@
-import Ember from 'ember';
+import { inject } from '@ember/service';
+import $ from 'jquery';
```
laurenbarker: Is this necessary? I think `$` is just globally available

app/components/search-facet-provider.js
```diff
@@ -51,7 +53,7 @@ export default Ember.Component.extend(Analytics, {
                     return providerNames;
                 }),
             // The providers list from SHARE
-            Ember.$
+            $
```
laurenbarker: `$.ajax({` can be on the same line

app/components/preprint-form-project-select.js
```diff
@@ -129,6 +135,23 @@ export default Ember.Component.extend(Analytics, {
     }).enqueue(),
 
     actions: {
+/*
```
laurenbarker: delete?

app/components/file-uploader.js
```diff
@@ -131,7 +154,7 @@ export default Ember.Component.extend(Analytics, {
         createProjectAndUploadFile() {
             // Upload case where user starting from scratch - new project/new file.  Creates project and then uploads file to newly
             // created project
-            Ember.get(this, 'metrics')
+            get(this, 'metrics')
```
laurenbarker: Is there a reason we are doing `get(this, 'metrics')` instead of `this.get('metrics')` everywhere? In v3.1 of ember we can start doing `this.metrics` so it might be best to go ahead an switch to `this.get('metrics')`.

app/components/preprint-form-section.js
```diff
@@ -39,7 +41,7 @@ export default CpPanelComponent.extend(Analytics, {
      */
     hasOpened: false,
 
-    trackOpenState: Ember.observer('isOpen', function() {
+    trackOpenState: observer('isOpen', function() {
```
adlius: Is there any particular reason why we are still using observers here? This might not be within the scope of this ticket, but it will be nice if we can avoid using observers as much as possible.

app/components/preprint-title-editor.js
```diff
@@ -33,10 +33,10 @@ const TitleValidation = buildValidations({
  * @class preprint-title-editor
  */
 
-export default Ember.Component.extend(TitleValidation, {
+export default Component.extend(TitleValidation, {
     title: null,
     titlePlaceholder: 'Enter preprint title',
-    isValid: Ember.observer('title', function() {
+    isValid: observer('title', function() {
```
adlius: The observer here could be removed as well. Again this may belong to another ticket.

app/index.html
```diff
@@ -21,7 +21,12 @@
             skipStartupTypeset: true
         });
     </script>
+    <link href="jquery.tagsinput.css" rel="stylesheet" type="text/css" />
     <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" type="text/css">
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/dropzone/5.1.1/min/basic.min.css" rel="stylesheet" type="text/css">
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/dropzone/5.1.1/min/dropzone.min.css" rel="stylesheet" type="text/css">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/dropzone/5.1.1/min/dropzone.min.js"></script>
```
adlius: These script-tag style import of dependencies doesn't look right. Talked to @baylee-d and it seems that Dropzone.js does not work without these imports. Same thing with tagsinput. Somehow `ember-osf-web` is also importing Dropzone.js this way, in addition to including them in package.json.

adlius: Ember-cli-dropzonejs could be a future solution.

app/components/add-preprint-box.js
```diff
@@ -1,4 +1,5 @@
-import Ember from 'ember';
+import { inject } from '@ember/service';
```
jamescdavis: The [recommended way](https://guides.emberjs.com/v2.18.0/applications/services/#toc_accessing-services) to do this is:
```
import { inject as service } from '@ember/service';
```

app/components/preprint-form-authors.js
```diff
@@ -36,15 +39,15 @@ import Analytics from 'ember-osf/mixins/analytics';
  * @class preprint-form-authors
  */
 export default CpPanelBodyComponent.extend(Analytics, {
-    i18n: Ember.inject.service(),
-    valid: Ember.computed.alias('newContributorId'),
+    i18n: inject(),
+    valid: computed.alias('newContributorId'),
```
jamescdavis: can also just `import { alias } from '@ember/object/computed';`  and just do:
```
valid: alias('newContributorId'),
```

app/components/search-preprints.js
jamescdavis: This ensures you're using the bundled jQuery

app/components/supplementary-file-browser.js
```diff
@@ -1,4 +1,7 @@
-import Ember from 'ember';
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { observer } from '@ember/object';
+import { get } from '@ember/object';
```
jamescdavis: You can roll these into one.

app/components/unregistered-contributor-form.js
```diff
@@ -1,4 +1,6 @@
-import Ember from 'ember';
+import { get } from '@ember/object';
+import { computed } from '@ember/object';
```
jamescdavis: combine with line above

app/controllers/submit.js
```diff
@@ -504,48 +518,83 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
...
+            this.get('panelActions').open(this.get(`_names.${this.get('_names').indexOf(nextPanel)}`));
+            get(this, 'metrics')
+                .trackEvent({
+                    category: 'div',
+                    action: 'click',
+                    label: `${this.get('editMode') ? 'Edit' : 'Submit'} - Click to edit, ${this.nextPanel} section`
+                });
+            this.set('currentPanelName', nextPanel);
+        },
+        */
```
jamescdavis: delete?

app/controllers/submit.js
```diff
@@ -566,19 +615,22 @@ export default Ember.Controller.extend(Analytics, BasicsValidations, NodeActions
...
                     .trackEvent({
                         category: 'button',
                         action: 'click',
                         label: `${this.get('editMode') ? 'Edit' : 'Submit'} - Authors Next Button`
                     });
             }
+            this.get('panelActions').close(this.get(`_names.${this.get('_names').indexOf(currentPanelName)}`));
             this.get('panelActions').open(this.get(`_names.${this.get('_names').indexOf(currentPanelName) + 1}`));
+            this.set('currentPanelName', this.get(`_names.${this.get('_names').indexOf(currentPanelName) + 1}`));
+//            this.set('currentPanelName', currentPanelName);
```
jamescdavis: delete?

app/routes/content/index.js
```diff
@@ -241,8 +246,8 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
             fileDownloadURL: this.get('fileDownloadURL'),
         });
 
-        Ember.run.scheduleOnce('afterRender', this, function() {
-            MathJax.Hub.Queue(['Typeset', MathJax.Hub, [Ember.$('.abstract')[0], Ember.$('#preprintTitle')[0]]]);  // jshint ignore:line
+        run.scheduleOnce('afterRender', this, function() {
+            MathJax.Hub.Queue(['Typeset', MathJax.Hub, [$('.abstract')[0], $('#preprintTitle')[0]]]);  // jshint ignore:line
```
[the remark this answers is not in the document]
jamescdavis: Nope!

app/routes/content/index.js
```diff
@@ -251,7 +256,7 @@ export default Ember.Route.extend(Analytics, ResetScrollMixin, SetupSubmitContro
     actions: {
         error(error) {
             // Handle API Errors
-            if (error && error.errors && Ember.isArray(error.errors)) {
+            if (error && error.errors && isArray(error.errors)) {
```
jamescdavis: Array.isArray() would work here too and save you an import

app/routes/index.js
```diff
@@ -12,11 +13,11 @@ import Analytics from 'ember-osf/mixins/analytics';
...
  */
-export default Ember.Route.extend(Analytics, ResetScrollMixin, {
-    // store: Ember.inject.service(),
-    theme: Ember.inject.service(),
+export default Route.extend(Analytics, ResetScrollMixin, {
+    store: inject(),
+    theme: inject(),
     model() {
-        return Ember.RSVP.hash({
+        return hash({
```
jamescdavis: This is a case where I'd `import RSVP from 'rsvp';` and do `RSVP.hash({`

app/routes/provider.js
```diff
@@ -42,8 +44,9 @@ export default Ember.Route.extend({
     },
 
     actions: {
-        error(error) {
-            Logger.error(error);
+        error() {
+            // TODO: this should be uncommented and import fixed.
+            // Ember.Logger.error(error);
```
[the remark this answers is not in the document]
jamescdavis: https://emberjs.com/deprecations/v3.x#toc_ember-console-deprecate-logger

app/routes/submit.js
```diff
@@ -15,18 +15,18 @@ import ConfirmationMixin from 'ember-onbeforeunload/mixins/confirmation';
...
+    panelActions: inject('panelActions'),
+    confirmationMessage: computed('i18n', function() {
         return this.get('i18n').t('submit.abandon_preprint_confirmation');
     }),
-
     model() {
         // Store the empty preprint to be created on the model hook for page. Node will be fetched
         //  internally during submission process.
-        return this.store.createRecord('preprint', {
+        return this.get('store').createRecord('preprint', {
```
jamescdavis: The previous code will be fine when Preprints goes to Ember 3.1 :wink:

app/utils/build-provider-asset-path.js
```diff
@@ -5,7 +5,7 @@ export default function buildProviderAssetPath(config, providerId, assetName, is
         return pathJoin(config.providerAssetsURL, providerId, assetName);
     }
     if (config.ASSET_SUFFIX) {
-        assetName = assetName.replace(/\.[^\.]+$/, `-${config.ASSET_SUFFIX}$&`);
+        assetName = assetName.replace(/\.[^.]+$/, `-${config.ASSET_SUFFIX}$&`);
```
jamescdavis: 👍

bower.json
```diff
@@ -1,26 +1,7 @@
...
-    "d3": "3.5.17",
-    "bootstrap-daterangepicker": "^2.1.23",
-    "sanitize.js": "^1.0.0"
-  },
-  "resolutions": {
-    "jquery": "2.2.4"
+    "sanitize.js": "^1.0.0",
+    "jquery.tagsinput": "1.3.6"
   }
 }
```
jamescdavis: You should be able to remove this file completely now.

ember-cli-build.js
```diff
@@ -33,7 +33,7 @@ module.exports = function(defaults) {
     }
 
     // Reference: https://github.com/travis-ci/travis-web/blob/master/ember-cli-build.js
-    const app = new EmberApp(defaults, {
+    let app = new EmberApp(defaults, {
```
jamescdavis: `const app` is appropriate here because `app` is never reassigned.

tests/helpers/module-for-acceptance.js
```diff
@@ -42,7 +41,7 @@ export default function(name, options = {}) {
     afterEach() {
         FakeServer.stop();
         let afterEach = options.afterEach && options.afterEach.apply(this, arguments);
-        return Promise.resolve(afterEach).then(() => destroyApp(this.application));
+        return resolve(afterEach).then(() => destroyApp(this.application));
```
jamescdavis: `RSVP.resolve`

tests/helpers/start-app.js
```diff
@@ -1,18 +1,17 @@
...
+  attributes.autoboot = true;
+  attributes = merge(attributes, attrs); // use defaults, but you can override;
 
-  let attributes = Ember.merge({}, config.APP);
-  attributes = Ember.merge(attributes, attrs); // use defaults, but you can override;
-
-  Ember.run(() => {
-    application = Application.create(attributes);
+  return run(() => {
+    let application = Application.create(attributes);
```
jamescdavis: const

adlius:
Also did a first pass. The only problem that stands out is the script-tag style imports. Maybe installing dropzone with bower could solve the problem, as that was what we did previously. If not, then we could probably address this later by using addons like ember-cli-dropzonejs.

app/components/unregistered-contributor-form.js
alexschiller: Those are all caught by ESLint and will go into the same release. There is another PR branched off this one that is partially done.

baylee-d:
Okay, I updated a few things: 
~ Changed all metrics calls to use `this.get('metrics')` instead of `get(this, 'metrics')`
~ Changed injections to be referenced as `service`
~ Merged imports together to remove redundant calls
~ Removed all observers

I looked back into moving dropzone imports from `index.html` back to `bower.json` to remove the static import strings, but I was getting the same error about not knowing what Dropzone was.  For now, I think it'll be okay as it is, but it can definitely be something that we look into in the future since we'll also have to revisit removing bower imports entirely.

public/crossdomain.xml
```diff
@@ -1,15 +0,0 @@
-<?xml version="1.0"?>
```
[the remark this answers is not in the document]
laurenbarker: It was removed from ember-cli: https://github.com/ember-cli/ember-new-output/compare/v2.11.0...v2.18.0#diff-ee1e9c1082dd87318aa6d8b1111ef7d4

app/services/i18n.js
```diff
@@ -0,0 +1,25 @@
+import EmberObject from '@ember/object';
+import { get } from '@ember/object';
+import { merge } from '@ember/polyfills';
+import I18nService from 'ember-i18n/services/i18n';
```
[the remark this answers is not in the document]
baylee-d: This was added to fix the problems I was having with variables in translation strings.  I think there was a fix for it at the time but it wasn't released yet.  Updating to the latest next-interfaces should fix this.

PR #561, opened 2018-05-03
[WIP][IN-271][Preprints] Resave preprint file relationship on submit

alexschiller:
## Purpose
This is a stupid fix for an annoying problem that was surprisingly difficult to figure out: how do you resave something that the front end already believes is saved? h/t's to @jamescdavis, @laurenbarker, and @baylee-d for help along the way.

We believe that somewhere in the backend celery tasks there is a race condition that is causing preprint files to get lost sometimes. That race condition likely will go away with the Node-Preprint divorce because it likely is due to the interaction of saving nodes and preprints and getting stale data. The "simple" solution was to resave the data on the front end upon completion of the preprint. That's what this does... albeit less elegantly than I hoped. 

## Summary of Changes (And a journey)
1. After saving the preprint model and node, remove the preprint model from the ember store.
2. Refetch the preprint data from the api
3. Reassign the preprint file to the preprint
4. Save the preprint

## The Journey
1. Reassigning and saving a field on the model that it doesn't believe is dirty does change anything. I feel like this is generally expected, but is an important part of a fun issue later.
2. Reloading data on a model first fetches it from the store if possible, which means if it has it, it won't go back to the API to get fresh data.
3. You can add {reload: true} to a findRecord request to tell it to repull the data from the API. At this point I thought the problem would be nailed, but...
4. If a relationship is null, our API doesn't return anything for that relationship. JSON API spec says empty to one relationships should be null (see ticket James created: https://openscience.atlassian.net/browse/PLAT-840). 
5. On reloading the data, ember puts everywhere where it belongs, but does not prune or remove any relationships from the store that aren't returned (this is likely the correct behavior given JSON API spec).
6. This means that despite force reloading, the now null preprint file relationship still exists in the ember data store, so any attempts to reset the value of the relationship to the primary file are treated as a clean field and are not saved (point 1 coming back to bite). 
7. Until the backend is fixed to return empty to none as null, the solution is to unload the model in ember store, refetch the data from the API, assign the relationship value, and then save the now dirty field. 

Extras:
My first "working" solution broke post moderation because it tried to publish the preprint before reattaching the file. The logic had to be refactored mildly to get things to occur in the right order.

Test:
1. Test creating preprints on each variation of moderated services (osf no moderation, premod, and post mod)
2. Test editing a preprint after each of those variations to ensure it saves properly when done editing.

## Ticket

https://openscience.atlassian.net/browse/IN-271

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
 ## [0.118.3] - 2018-03-12
 ### Added
 - Event Tracking to `Select a Preprint Provider` on preprint upload
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Event Tracking to `Select a Preprint Provider` on preprint upload
+
+### Changed
 - Restricted width and centered preprint provider logos on index page
 
 ## [0.118.2] - 2018-03-08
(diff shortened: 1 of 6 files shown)


alexschiller:
Closing in favor of #571

PR #579, opened 2018-06-13
[IN-198][Preprints] Node parent title

chrisseto:
## Purpose
Node parents sometimes show as private when searching for a project.

## Summary of Changes

Embed node parent to allow parent title's to appear in the select project dropdown

Alex notes: Make sure that parents of projects don't say private when not private (see issue description). There are two functions that fetch the data, one for the first fetch and one for every fetch after. So make sure both the first fetch of data works and additional filtering works. Ask me (alex) if that isn't clear.

## Ticket

https://openscience.atlassian.net/browse/IN-198

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/preprint-form-project-select.js
+++ b/app/components/preprint-form-project-select.js
@@ -184,6 +184,7 @@ export default Component.extend(Analytics, {
                 preprint: false,
                 title: searchTerm,
             },
+            embed: 'parent',
         });
         // When the promise finishes, set the searchTerm
         this.set('searchTerm', searchTerm);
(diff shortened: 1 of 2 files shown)


app/components/preprint-form-project-select.js
```diff
@@ -184,6 +184,7 @@ export default Component.extend(Analytics, {
                 preprint: false,
                 title: searchTerm,
             },
+            embed: 'parent',
```
alexschiller: should `_getMoreUserNodes` also embed parent?

chrisseto: Yep! Good catch.

chrisseto: Yep! Good catch.

PR #580, opened 2018-06-14
[IN-255][Preprints] Use "OSF Preprints"

fabmiz:
## Purpose

Use "OSF Preprints" instead of "Open Science Framework"

## Summary of Changes
- Update search provider facet to use "OSF Preprints" (on discover page)
- Update `filterReplace` to sub with "OSF Preprints"  in lieu of "OSF"
- Fixed `whiteListedProviders` to handle case where is still `undefined`

Please make sure that 
- the search provider facet on discover page left pane uses "OSF Preprints"
- on submit page, after selecting OSF Preprints and moving on to the next section, it shows "OSF Preprints"
- on preprint detail page, the publisher head meta tags use "OSF Preprints" for osf preprints.

## Ticket

[IN-255](https://openscience.atlassian.net/browse/IN-255)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/controllers/discover.js
+++ b/app/controllers/discover.js
@@ -34,7 +34,7 @@ export default Controller.extend(Analytics, {
     },
     // TODO: Add a conversion from shareSource to provider names here if desired
     filterReplace: { // Map filter names for front-end display
-        'Open Science Framework': 'OSF',
+        'Open Science Framework': 'OSF Preprints',
         'Cognitive Sciences ePrint Archive': 'Cogprints',
         OSF: 'OSF Preprints',
         'Research Papers in Economics': 'RePEc',
--- a/app/templates/components/preprint-form-header.hbs
+++ b/app/templates/components/preprint-form-header.hbs
@@ -26,7 +26,7 @@
             <em class="m-r-md"> {{t "components.preprint-form-header.server" documentType=provider.documentType}}: </em>
             {{#if currentProvider}}
                 <img alt="provider logo" class="current-provider" title="{{currentProvider.name}}" src={{provider-asset currentProvider.id 'square_color_no_transparent.png'}}>
-                {{currentProvider.name}}
+                {{if (not theme.isProvider) 'OSF Preprints' currentProvider.name}}
             {{/if}}
         </p>
     </div>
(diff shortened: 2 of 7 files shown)


app/controllers/discover.js
```diff
@@ -34,7 +34,7 @@ export default Controller.extend(Analytics, {
     },
     // TODO: Add a conversion from shareSource to provider names here if desired
     filterReplace: { // Map filter names for front-end display
-        'Open Science Framework': 'OSF',
+        'Open Science Framework': 'OSF Preprints',
         'Cognitive Sciences ePrint Archive': 'Cogprints',
         OSF: 'OSF Preprints',
```
alexschiller: is this one still relevant?

app/templates/components/preprint-form-header.hbs
```diff
@@ -26,7 +26,7 @@
             <em class="m-r-md"> {{t "components.preprint-form-header.server" documentType=provider.documentType}}: </em>
             {{#if currentProvider}}
                 <img alt="provider logo" class="current-provider" title="{{currentProvider.name}}" src={{provider-asset currentProvider.id 'square_color_no_transparent.png'}}>
-                {{currentProvider.name}}
+                {{if isOSFPreprints 'OSF Preprints' currentProvider.name}}
```
alexschiller: Where is `isOSFPreprints` coming from? I don't see this var anywhere?

app/controllers/discover.js
fabmiz: OSF preprints from share results has the provider/source as `OSF`.

app/templates/components/preprint-form-header.hbs
fabmiz: It's defined on submit controller below and is passed down as argument to any of the child components that could use the special casing. Alternatively, it could be also computed on the template or be added to the preprint provider model, but I thought it's cleaner to compute once and pass it down as arguments to other components.

https://github.com/CenterForOpenScience/ember-osf-preprints/blob/874030da78b0ddc9bcdcb31d034061b901ba53e2/app/controllers/submit.js#L289-L291

PR #581, opened 2018-06-18
[IN-180][Preprints] Track custom dimensions in Google Analytics

baylee-d:
## Purpose

To add custom dimensions to our `trackPage` function.

## Summary of Changes

- Add `authorized`, `resource`, and `isPublic` dimensions to the `trackPage` function
- Remove old version of `ember-metrics` since we're getting a newer version from `ember-osf`

There should be three new dimensions that you see in a trackPage request:

If the page is discover, index, or submit/edit:
authorized => [whether or not the user is logged in will be true or false]
isPublic => [n/a]
resource => [n/a]

If the page is content/detail page:
authorized => [whether or not the user is logged in will be true or false]
isPublic => [public]
resource => [preprint]

## Ticket

https://openscience.atlassian.net/browse/IN-180

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -9,6 +9,7 @@ const Router = EmberRouter.extend({
     rootURL: config.rootURL,
     metrics: service(),
     theme: service(),
+    session: service(),
 
     didTransition() {
         this._super(...arguments);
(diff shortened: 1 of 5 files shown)


app/router.js
```diff
@@ -17,10 +18,32 @@ const Router = EmberRouter.extend({
...
+            const {
+                authenticated,
+                isPublic,
+                resource,
+            } = config.metricsAdapters[0].dimensions;
+
             const page = document.location.pathname;
             const title = this.getWithDefault('currentRouteName', 'unknown');
+            const isAuthenticated = this.get('session.isAuthenticated');
+            const publicType = title.endsWith('content.index') ? 'public' : 'n/a';
```
alexschiller: In theory you can have a private node or an unpublished preprint. This doesn't seem to handle that at all. Should it?

package.json
```diff
@@ -27,7 +27,7 @@
   "devDependencies": {
     "@centerforopenscience/ember-osf": "https://github.com/CenterForOpenScience/ember-osf.git#release/next-interfaces#2018-05-30T18:58:56Z",
     "@centerforopenscience/eslint-config": "^2.0.0",
-    "@centerforopenscience/osf-style": "https://github.com/CenterForOpenScience/osf-style#b09451d759c8544d2fab4bd7dbd5ca8b6dc242b8",
+    "@centerforopenscience/osf-style": "https://github.com/CenterForOpenScience/osf-style#370b544affcdafc5734cdd388587575895e4b33a",
```
alexschiller: I have a feeling this is just caused by being out of date, can you rebase with next-next-interfaces

app/router.js
baylee-d: After talking it out this seems like it isn't something we should worry about, since it'll be fixed after the node/preprint divorce.

PR #585, opened 2018-06-22
[WIP][IN-305][Preprints] Fix LiveData (by refactoring the discover page) 

laurenbarker:
## Purpose
Get LiveData working with Ember 2.18.

## Summary of Changes
* moved most of the preprints specific logic out of ember-osf and into preprints
* used ember-parachute for query param management
* upgraded some dependencies
* removed "ActiveFilters" in favor of managing all of the query params the same way

Should look at the styling and functionality of the discover page for preprint providers and LiveData. This is a pretty big change so anything could be broken.

## Ticket

https://openscience.atlassian.net/browse/IN-305

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/search-facet-typeahead/component.js
```diff
@@ -0,0 +1,85 @@
...
+        return { size: RESULTS, query: match };
+    },
+
+    searchElastic: task(function* (term) {
+        if (isBlank(term)) { yield []; }
+        yield timeout(DEBOUNCE_MS);
+
+        const data = JSON.stringify(this.buildTypeaheadQuery(term));
+
+        const response = yield $.ajax({
```
laurenbarker: Should add `import $ from 'jquery';`

.stylelintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
laurenbarker: I'm not sure why this got added -- remove

app/components/search-facet-source/component.js
```diff
@@ -0,0 +1,69 @@
...
+                    label: {
+                        show: false,
+                    },
+                },
+                size: { height: 200 },
+            });
+            this.set('donut', donut);
+        } catch (e) {
+            // eslint-disable-next-line no-console
+            console.debug(e);
```
laurenbarker: Not sure if this should be debug or error

app/controllers/discover.js
```diff
@@ -126,17 +249,320 @@ export default Controller.extend(Analytics, {
         });
         return themeProvider;
     }),
-    _clearFilters() {
-        this.set('activeFilters', {
-            providers: [],
-            subjects: [],
+
+    searchUrl: computed(function() {
```
laurenbarker: This either doesn't need to be a computed property or should be dependent on `currentUser.sessionKey` (copied from ember-osf)

app/controllers/discover.js
```diff
@@ -126,17 +249,320 @@ export default Controller.extend(Analytics, {
...
+         */
+        let filters = this.buildLockedQueryBody(this.get('lockedParams')); // Empty list if no locked query parameters
+        // From Ember-SHARE. Looks at facetFilters
+        // (partial SHARE queries already built) and adds them to query body
+        if (this.get('additionalProviders')) {
+            const facetFilters = this.constructFacetFilters();
+            for (const k of Object.keys(facetFilters)) {
+                const filter = facetFilters[k];
+                if (filter) {
+                    if ($.isArray(filter)) {
```
laurenbarker: Should add `import $ from 'jquery';`

app/controllers/discover.js
```diff
@@ -126,17 +249,320 @@ export default Controller.extend(Analytics, {
...
+        if (page === 1 || this.get('firstLoad')) {
+            queryBody.aggregations = elasticAggregations;
+        }
+
+        return this.set('queryBody', queryBody);
     },
-    _clearQueryString() {
-        this.set('q', '');
+
+    trackDebouncedSearch() {
```
laurenbarker: This isn't currently called anywhere -- need to figure out where makes sense

app/controllers/discover.js
```diff
@@ -126,17 +249,320 @@ export default Controller.extend(Analytics, {
...
+        });
     },
-    // Type query param. Must be passed to component, so can be reflected in URL
+
+    fetchData: task(function* (queryParams) {
+        yield timeout(DEBOUNCE_MS);
+        const queryBody = JSON.stringify(this.getQueryBody(queryParams));
+
+        try {
+            const response = yield $.ajax({
```
laurenbarker: import $ from 'jquery';

app/styles/app.scss
```diff
@@ -2,12 +2,17 @@
 @import 'variables';
 @import 'bootstrap';
 @import 'ember-power-select';
+@import 'daterangepicker.scss';
+@import 'c3';
```
laurenbarker: Some import is causing a deprecation warning:
```
DEPRECATION WARNING on line 7, column 8 of /code/tmp/sass_compiler-input_base_path-0ZTDxWsh.tmp/0/app/styles/app.scss:
Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass.
Use a custom importer to maintain this behaviour. Check your implementations documentation on how to create a custom importer.
```

I think it might be [hint.css](https://github.com/CenterForOpenScience/ember-osf-preprints/pull/585/files#diff-2fe2a7faee7de1e0d5d82e032498d84cR66)

app/styles/app.scss
```diff
@@ -219,7 +224,7 @@ $logo-dir: 'img/provider_logos/';
     }
 }
 .search-main {
-    background-color: #fbfbfb;
+    background-color: #fbfbfb !important;
```
laurenbarker: There is probably a better solution than `!important`

app/styles/search-facets.scss
```diff
@@ -0,0 +1,78 @@
+.facet-indent {
+    padding-left: 30px;
+}
+
+// faceted-search
+.ember-power-select-trigger {
+    background-color: #fff !important;
```
laurenbarker: Also here -- probably something better than `!important`

app/utils/elastic-query.js
```diff
@@ -0,0 +1,113 @@
+import moment from 'moment';
```
laurenbarker: Want to make sure that all of these are actually used

laurenbarker: They are in fact all used

app/styles/app.scss
laurenbarker: This can be taken care of as a separate ticket as it wasn't caused by my refactor.

app/components/search-facet-source/component.js
alexschiller: Wasn't sure the exact difference between all of em, so found the below. Doesn't really matter. If we are concerned about it, a proper error to sentry would be better,  but I don't really care.

```
Technically console.log console.debug and console.info are identical
However the way they display the data is little different

console.log Black color text with no icon

console.info Blue color text with icon

console.debug Pure black color text

console.warn Yellow color text with icon

console.error Red Color text with icon
```
https://stackoverflow.com/questions/21876461/difference-between-console-log-and-console-debug

app/components/additional-provider-list.js
```diff
@@ -28,9 +28,11 @@ export default Component.extend(Analytics, {
         const pairedList = [];
         for (let i = 0; i < sortedList.get('length'); i += 2) {
             const pair = [];
-            pair.pushObject(sortedList.objectAt(i));
+            // provider in pair needs to be a list because that's
+            // what the discover controller is expecting
+            pair.pushObject([sortedList.objectAt(i)]);
```
alexschiller: Was this broken before? Just curious why it changed. Different input data now?

app/controllers/discover.js
```diff
@@ -126,17 +250,324 @@ export default Controller.extend(Analytics, {
...
+                    if ($.isArray(filter)) {
+                        filters = filters.concat(filter);
+                    } else {
+                        filters.push(filter);
+                    }
+                }
+            }
+        }
+
+        // For PREPRINTS and REGISTRIES.
```
alexschiller: Probably deletable, since definitely not for registries now

app/controllers/discover.js
```diff
@@ -126,17 +250,324 @@ export default Controller.extend(Analytics, {
...
+            } else {
+                filters.push({
+                    terms: {
+                        [val]: filterList,
+                    },
+                });
+            }
+        });
+
+        // For PREPRINTS and REGISTRIES. If theme.isProvider, add provider(s) to query body
```
alexschiller: also reg comment here

app/controllers/discover.js
```diff
@@ -126,17 +250,324 @@ export default Controller.extend(Analytics, {
...
+            const results = response.hits.hits.map((hit) => {
+                // HACK: Make share data look like apiv2 preprints data
+                const result = merge(hit._source, {
+                    id: hit._id,
+                    type: 'elastic-search-result',
+                    workType: hit._source['@type'],
+                    abstract: hit._source.description,
+                    subjects: hit._source.subjects.map(each => ({ text: each })),
+                    subject_synonyms: hit._source.subject_synonyms.map(each => ({ text: each })),
+                    providers: hit._source.sources.map(item => ({ // PREPRINTS, REGISTRIES
```
alexschiller: more reg comments

app/index.html
```diff
@@ -25,7 +25,6 @@
     <link href="https://cdnjs.cloudflare.com/ajax/libs/dropzone/5.1.1/min/basic.min.css" rel="stylesheet" type="text/css">
     <link href="https://cdnjs.cloudflare.com/ajax/libs/dropzone/5.1.1/min/dropzone.min.css" rel="stylesheet" type="text/css">
     <script src="https://cdnjs.cloudflare.com/ajax/libs/dropzone/5.1.1/min/dropzone.min.js"></script>
-    <script src="jquery.tagsinput.js"></script>
```
alexschiller: 👍

app/styles/app.scss
alexschiller: I feel like I've seen that one for a while. I have vague recollections of trying to solve but having issues. Either way is fine by me.

app/styles/app.scss
alexschiller: Probably, but also probably not worth the effort atm 😆

app/components/additional-provider-list.js
laurenbarker: It was not, I just made the serialization/deserialization consistent. I could add "convert to list if not a list" as part of that if you prefer. I figured it would be more clear if it breaks for improper type.

alexschiller: nah, was more just curious

app/components/search-facet-source/component.js
laurenbarker: Are sentry errors only sent for uncaught exceptions or is there a way to send them directly?

alexschiller: @fabmiz, want to comment here? You've messed with sending things to sentry more than I have.

laurenbarker: Looks like it's definitely possible https://docs.sentry.io/clients/javascript/usage/#reporting-errors-correctly

app/styles/app.scss
laurenbarker: Fair, will hold off on this then

PR #614, opened 2018-08-09
[IN-386][Preprints] Fix carousel jumping

alexschiller:
## Purpose

If on the second page of a smaller than 1200 carousel, it bounces back to the first page on select.

Move the starting count to the component instead of being passed in. The carousel is ONLY used here now, so this is fine, since previously it was used in more than one place.

BEFORE:
jumpservice
AFTER:
service_fix

## Testing Notes

Make sure everythign with the carousel works as expected. Test various screen widths. Make sure clicking on icons does not bounce to other pages, etc.

## Ticket

https://openscience.atlassian.net/browse/IN-386

## Notes for Reviewer

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


adlius:
Tested locally. LGTM.

PR #616, opened 2018-08-14
[IN-128][Preprints] Remove branded footer line-height

mfraezz:
## Purpose
From ticket
>Reduce line height on footer text when wrapping. It doesn't look like the text is all one thing if the space between the wrapped text is the same as the space between the other lines.

* Remove `line-height`

### Before:
lh-before

### After:
lh-after

## Testing Notes
Should only affect footer, make sure it looks right on small screens

## Ticket
[[IN-128]](https://openscience.atlassian.net/browse/IN-128)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - word break in license text
 
+### Removed
+- branded footer line-height
+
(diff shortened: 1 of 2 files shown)


alexschiller:
That was weird, I fixed conflict but it didn't go right away. TCI was passing before that, minor changes so merging.

PR #617, opened 2018-08-15
[IN-363][Preprints] Allow downloading previous versions of preprints and/or suppl. files

fabmiz:
## Purpose

Based on @binoculars PR https://github.com/CenterForOpenScience/ember-osf-preprints/pull/285
Allow versions of preprint or suppl. files to be downloaded from the preprint detail page.
ht @binoculars 

- Updated `supplementary-file-browser` to support downloading file versions
- Updated/fixed and added tests 

## Testing Notes
....coming soon after demo.

## Ticket

>>> [IN-363 ](https://openscience.atlassian.net/browse/IN-363)

## After

http://recordit.co/uDykCHmWSo
**latest: ** http://recordit.co/UZ0S0MYly0

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/supplementary-file-browser.js
+++ b/app/components/supplementary-file-browser.js
@@ -1,5 +1,8 @@
+import Ember from 'ember';
 import Component from '@ember/component';
-import { computed, observer } from '@ember/object';
+import { task } from 'ember-concurrency';
+import DS from 'ember-data';
+import { computed } from '@ember/object';
 import { inject as service } from '@ember/service';
 import loadAll from 'ember-osf/utils/load-relationship';
 import Analytics from 'ember-osf/mixins/analytics';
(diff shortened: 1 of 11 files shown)


app/templates/components/supplementary-file-browser.hbs
```diff
@@ -5,45 +5,69 @@
...
+    <div class="clearfix visible-xs-block"></div>
+    <div class="col-lg-6 col-xs-12 text-nowrap supplemental-downloads">
+        {{#if (eq selectedFile.id preprint.primaryFile.id)}}
+            <div class="dropdown lock-span">
+                <button class="btn btn-default btn-sm dropdown-toggle" type="button" id="previousVersionsDropdown"
+                  data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
+                    {{t 'components.supplementary-file-browser.download_previous_versions'}}
+                    <span class="caret"></span>
+                </button>
+                <ul id="previousVersionsList" class="dropdown-menu" aria-labelledby="previousVersionsDropdown">
```
alexschiller: This needs to fetch all or paginate or something. Possibly just page size huge (maybe 100), doubtful anyone would have more versions than that.
screen shot 2018-08-28 at 10 03 44 am

app/components/supplementary-file-browser.js
```diff
@@ -148,29 +153,36 @@ export default Component.extend(Analytics, {
...
-            this.set('selectedFile', file);
-            if (this.chooseFile) {
-                this.chooseFile(file);
+            this.setProperties({
+                selectedFile,
+                selectedIndex,
+            });
+
+            if (this.attrs.chooseFile) {
+                this.sendAction('chooseFile', selectedFile); // eslint-disable-line ember/closure-actions
```
adlius: Haven't looked at this deep enough. But is there any particular reason why we need to use `sendAction`, which is on its way to being deprecated, instead of closure actions?

app/components/supplementary-file-browser.js
```diff
@@ -25,100 +27,95 @@ import fileDownloadPath from '../utils/file-download-path';
...
-    }),
+    hasAdditionalFiles: computed.gt('files.length', 1),
+    hasPrev: computed.gt('startIndex', 0),
+    selectedFileIsPrimaryFile: computed.equal('selectedIndex', 0),
 
-    hasPrev: computed('files', 'endIndex', 'startIndex', function() {
-        return this.get('startIndex') > 0;
+    primaryFileHasVersions: computed('selectedFileIsPrimaryFile', 'versions', function() {
+        const versions = this.get('versions');
+        if (!versions) return;
```
adlius: Should this return `false`?

app/components/supplementary-file-browser.js
```diff
@@ -25,100 +27,95 @@ import fileDownloadPath from '../utils/file-download-path';
...
+                providers.findBy('name', 'osfstorage'),
+                'files',
+                this.get('files'),
+                {
+                    'page[size]': 50,
+                },
+            ))
+            .then(() => this.get('preprint.primaryFile'))
+            .then(this.__initProperties.bind(this));
+    }.observes('preprint'),
```
adlius: I don't know how I should feel about using `observes` here. This is probably due to the fact that promises in computed functions don't work so well. But I have found some literature on [this](https://stackoverflow.com/questions/36945858/promise-result-in-ember-data-computed-property) which may worth a look.

app/components/supplementary-file-browser.js
```diff
@@ -25,100 +27,95 @@ import fileDownloadPath from '../utils/file-download-path';
...
+
     init() {
         this._super(...arguments);
+        this.setProperties({
+            files: A(),
+            selectedFile: null,
+            selectedIndex: null,
+        });
         this.__files();
     },
```
adlius: Do we want to avoid running these upon rerender? If not, then maybe worth looking at moving these to `didReceiveAttrs()`, which is a lifecycle hook that runs immediately after `init()` upon initial render. Not super important though. As a personal preference, I always avoid overriding `init()` unless I absolutely have to.

app/components/supplementary-file-browser.js
adlius: I feel like the `__files()` method could be refactored to a computed function while making use of `ember-concurrency`. I don't like the chaining promises and the `observes` in particular. I can see that this is somewhat a port of what was done in #285 . Therefore if we can refactor it with not much effort, it would be nice. Maybe don't worry about it if it takes too much time.

app/components/supplementary-file-browser.js
```diff
@@ -179,4 +182,50 @@ export default Component.extend(Analytics, {
...
+
+        this.setProperties({
+            primaryFile,
+            selectedFile,
+            files,
+            selectedIndex: files.indexOf(selectedFile),
+        });
+    },
+    getFiles: task(function* () {
+        yield timeout(500);
```
adlius: Also, why a 500ms timeout here?

PR #629, opened 2018-10-05
[EMB-519] Allow claiming of unregistered users.

adlius:
## Purpose

This PR allows users to claim unregistered users on Preprints app.
The `ember-osf` side PR is [here](https://github.com/CenterForOpenScience/ember-osf/pull/419).

Added a new component `claim-user` and related 

## Testing Notes

Note: Logged in users cannot claim unregistered users.

## Ticket

https://openscience.atlassian.net/browse/EMB-519

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/claim-user.js
```diff
@@ -0,0 +1,90 @@
...
+import { validator, buildValidations }
+    from 'ember-cp-validations';
+
+const Validations = buildValidations({
+    email: {
+        validators: [
+            validator('presence', true),
+            validator('format', {
+                type: 'email',
+                regex: /^[-a-z0-9~!$%^&*_=+}{'?]+(\.[-a-z0-9~!$%^&*_=+}{'?]+)*@([a-z0-9_][-a-z0-9_]*(\.[-a-z0-9_]+)*\.(aero|arpa|biz|com|coop|edu|gov|info|int|mil|museum|name|net|org|pro|travel|mobi|[a-z][a-z])|([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}))(:[0-9]{1,5})?$/i,
```
alexschiller: iirc this comes from somewhere else. Should this be referenced from a shared file?

app/components/claim-user.js
```diff
@@ -0,0 +1,90 @@
...
+    },
+});
+
+/**
+ * @module ember-preprints
+ * @submodule components
+ */
+
+/**
+ * Confirm share preprint modal
```
alexschiller: er this was all copy pasta  I didn't rewrite, can you make it accurate for this new component

app/styles/preprint-detail-page.scss
```diff
@@ -0,0 +1,18 @@
...
+    margin-left: 7px;
+    height: 34px;
+}
+
+.popover-content {
+    background: white;
+    overflow: auto;
+
+    .form-group {
+        margin-bottom: 0px;
```
alexschiller: px shouldn't be needed for 0

adlius: I have to force the `form-group`'s bottom margin to be 0px because the default it inherited from somewhere else is not, and the default breaks the widget layout.

app/templates/components/claim-user.hbs
```diff
@@ -0,0 +1,12 @@
+<span {{action 'togglePopup'}}>{{author.unregisteredContributor}}
+    {{#bs-tooltip}}{{t 'components.claim-user.tooltip_message'}}{{/bs-tooltip}}
+    {{#bs-popover placement='bottom' title=(t 'components.claim-user.title') visible=showPopup}}
+        {{#click-outside action=(action 'hidePopup') }}
+            <div style="display: flex">
```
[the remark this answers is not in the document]
adlius: Normally it could either be an inline style or a css class. Be since we don't use css files that's local to only the component in `ember-osf-preprints` (we do use local css files for `ember-osf`), it means I have to put it either in `app.scss` or other similar files in the `app/styles` directory, which is something I want to avoid as much as possible.

app/locales/en/translations.js
```diff
@@ -466,7 +471,7 @@ export default {
         },
         'unregistered-contributor-form': {
             full_name: 'Full Name',
-            email: 'Email',
+            email: 'Enter email',
```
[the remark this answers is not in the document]
adlius: It does.

app/components/claim-user.js
```diff
@@ -0,0 +1,90 @@
...
+    isValid: alias('validations.isValid'),
+    actions: {
+        hidePopup() {
+            this.set('showPopup', false);
+        },
+        togglePopup() {
+            this.toggleProperty('showPopup');
+        },
+    },
+    claimUser: task(function* () {
```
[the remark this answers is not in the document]
adlius: I thought about doing this on the `users` model, but I don't think it is quite right. Here what I think:
If we are to put it in the `users` model, that means we have to override the `createRecord()` method, as well as the `urlForCreateRecord()` method of the adapter. That is too much footprint for something that is going to be used once here. 
Also, upon claiming the user, the API returns 204 with empty content, instead of details of an `OSFUser` (because technically the OSFUser is not created until the email is confirmed). That means ember doesn't really get a `user` instance in return when we do `user.save()`, which COULD cause some more problems we need to fix by adding more logic to the `user` serializer as this is probably not JSONAPI-compliant anyway. To go down that rabbit hole does not worth the hassle in my opinion.

app/components/claim-user.js
adlius: um... If possible I don't want to reference a regex string but the entire validation object from a shared file. I'll see what I can do.

app/styles/preprint-detail-page.scss
alexschiller: ah gotcha

app/components/claim-user.js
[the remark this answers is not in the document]
alexschiller: Was bugging jams about this, slightly better on the model, I suggest asking him for clarification on best way to do that, doesn't need to be adapter changes, etc.
```
But, either way, I agree the code for this should go in the user model. This abstracts it away from the component so you could go with the POST to /claim now and make it simpler later.
```

app/components/claim-user.js
alexschiller: Don't spend long on this, not worth a lot of effort

adlius:
Updated to latest ember-osf. Conflicts resolved.

PR #642, opened 2018-12-02
Fix/NPD - Advancing Discipline Section [PLAT-1248]

pattisdr:
## Purpose
Bugfix for NPD: When editing a preprint, hitting "Save and Continue" on disciplines doesn't advance you through the form if there have been no changes

`this` was undefined when just calling `this._moveFromDisciplines`

## Ticket

https://openscience.atlassian.net/browse/PLAT-1248
## Notes for Reviewer

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/controllers/submit.js
+++ b/app/controllers/submit.js
@@ -884,7 +884,7 @@ export default Controller.extend(Analytics, BasicsValidations, NodeActionsMixin,
                     label: `${this.get('editMode') ? 'Edit' : 'Submit'} - Discipline Save and Continue`,
                 });
 
-            const sendNext = this._moveFromDisciplines;
+            const sendNext = () => this.send('next', this.get('_names.3'));
 
             if (!hasChanged) {
                 return sendNext();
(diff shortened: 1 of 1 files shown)


app/controllers/submit.js
```diff
@@ -884,7 +884,7 @@ export default Controller.extend(Analytics, BasicsValidations, NodeActionsMixin,
                     label: `${this.get('editMode') ? 'Edit' : 'Submit'} - Discipline Save and Continue`,
                 });
 
-            const sendNext = this._moveFromDisciplines;
+            const sendNext = () => this.send('next', this.get('_names.3'));
```
jamescdavis:
```js
const sendNext = this._moveFromDisciplines.bind(this);
```
would also have worked here, but this is fine too.

pattisdr: Thanks @jamescdavis, that's better!

PR #703, opened 2020-03-20
[ENG-1435]Add component to handle multiple text inputs

futa-ikeda:
## Purpose
- Add a component to handle a variable amount of text inputs. Needed for Sloan Signals of Trust feature for users to add links.

- Added component `multiple-textbox-input`
 - takes `placeholder` and `textFields` argument. `textFields` is a string[] that is used to populate this field if there are values already present.

## Testing Notes
- There are some cases where it can remove the wrong textbox, but I'm having trouble narrowing down when/why this happens. And occasionally the `Add` button can get put between the `Remove` buttons, but this is also very sporadic. Not sure how to reproduce.
- Please test when textfields are already populated (i.e.: a user edits a preprint with data links or prereg links already filled out)

## Ticket

https://openscience.atlassian.net/browse/ENG-1435

## Notes for Reviewer
- Please let me know if there are preprints conventions that are not being followed!

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/multiple-textbox-input.js
+++ b/app/components/multiple-textbox-input.js
@@ -1,0 +1,37 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { task, timeout } from 'ember-concurrency';
+import Analytics from 'ember-osf/mixins/analytics';
+
+export default Component.extend(Analytics, {
+    textFields: null, // passed in array of text fields on invocation, default to null
+    alphabet: 'abcdefghijklmnop',
+    textFieldsLastIndex: computed('textFields.[]', function() {
+        return this.get('textFields').length - 1;
+    }),
+
+    actions: {
+        addTextField() {
+            let fields = this.get('textFields');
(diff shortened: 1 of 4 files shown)


adlius:
Talked to @futa-ikeda in person and suggested trying to use the `value` binding of the input helper to bind each text fields to individual array elements.

app/components/multiple-textbox-input.js
```diff
@@ -0,0 +1,37 @@
...
+            this.set('textFields', fields);
+        },
+    },
+    onFieldChange: task(function* (index) {
+        yield timeout(200); // debounce
+        let fields = this.get('textFields');
+        if (!fields) {
+            fields = [$('.ember-text-field')[0].value];
+        }
+        fields[index] = $('.ember-text-field')[index].value;
```
futa-ikeda: This is not the best way to get these values, so will be looking into using a different approach

PR #705, opened 2020-04-02
[ENG-1435] Add public data and prereg fields to Author Assertions panel.

adlius:
## Purpose
Add public data field to Author Assertions panel. Some other minor cleanup. Update the `multi-text-input` component.


app/controllers/submit.js
```diff
@@ -1315,6 +1388,25 @@ export default Controller.extend(Analytics, BasicsValidations, COIValidations, N
...
+            this.get('metrics')
+                .trackEvent({
+                    category: 'button',
+                    action: 'click',
+                    label: `${this.get('editMode') ? 'Edit' : 'Submit'} - Save and Continue Author Assertions Section`,
+                });
+            const model = this.get('model');
+            model.set('hasDataLinks', this.get('hasDataLinks'));
+            model.set('dataLinks', this.get('dataLinks'));
+            model.set('whyNoData', this.get('whyNoData'));
```
futa-ikeda, line 1401: Should `dataLinks` and `whyNoData` be set on the model all the time? Or should these be set for `available` and `no` respectively?

app/controllers/submit.js
```diff
@@ -1315,6 +1388,25 @@ export default Controller.extend(Analytics, BasicsValidations, COIValidations, N
...
+            model.save().then(this._moveFromAuthorAssertions.bind(this))
+                .catch(this._failMoveFromAuthorAssertions.bind(this));
+        },
+        discardAuthorAssertions() {
+            this.set('hasDataLinks', this.get('model.hasDataLinks'));
+            this.set('dataLinks', this.get('model.dataLinks'));
+            this.set('whyNoData', this.get('model.whyNoData'));
```
futa-ikeda, line 1408: should the discard option be tracked as well? I don't remember if it was asked to be.

adlius: I don't remember any GA requirement for that.

app/controllers/submit.js
adlius: We have other mechanisms in place to make sure `dataLinks` is only set on the compute when `hasDataLinks === 'available'` and `whyNoData` is defined when `hasDataLinks==='no'`, so this should be fine.

app/components/multiple-textbox-input.js
```diff
@@ -3,16 +3,22 @@ import { computed } from '@ember/object';
...
     textFieldsLastIndex: computed('textFields.[]', function() {
         return this.get('textFields').length - 1;
     }),
 
     didReceiveAttrs() {
-        if (!this.textFields) {
-            this.set('textFields', [{ value: '' }]);
+        const valuesFromModel = this.model.get(this.valuePath);
+        if (valuesFromModel) {
+            this.set('textFields', valuesFromModel.map((x) => {
```
fabmiz, line 16: shouldn't `valuesFromModel` be an array of links, why is the `valuesFromModel.map(...)` necessary?

app/templates/components/author-assertions.hbs
```diff
@@ -37,7 +37,11 @@
                                 analyticsName=(t 'components.author-assertions.public_data.title')
                             }}
                         {{else if (eq preprint.hasDataLinks 'no')}}
-                            {{~t 'components.author-assertions.public_data.no'~}}
+                            {{#if preprint.whyNoData}}
+                                {{preprint.whyNoData}}
+                            {{else}}
+                                {{~t 'components.author-assertions.public_data.no'~}}
+                            {{/if}}
```
fabmiz, line 44: Same needs to be done [below](https://github.com/CenterForOpenScience/ember-osf-preprints/pull/705/files#diff-02aab108141199396345ae5399e10111R66) for `whyNoPrereg`?

app/templates/components/author-assertions.hbs
```diff
@@ -1,6 +1,6 @@
 {{#if this.shouldShowSloanIcons}}
     <div class='author-assertions-header-label'>
-        {{~t 'components.author-assertions.header_label'~}}
+        {{~t 'components.preprint-form-header.name.Assertions'~}}
```
fabmiz, line 3: Translations are grouped by component. Is there any specific reason why this changed despite the translation being used in `author-assertions`?

app/components/multiple-textbox-input.js
adlius: If you look at the function we pass to `map()`, you'll see that it returns `{ value: x }`. So what is does is transforming an array `['link1', 'link2', 'link3']` into another array `[{ value: 'link1'}, { value: 'link2'}, { value: 'link3'}]. The reason behind such transformation is because if we want the `value` binding of the input helper to work, we have to bind to objects instead of primitives.

app/templates/components/preprint-form-header.hbs
```diff
@@ -96,9 +98,36 @@
...
+                        {{#if (eq hasDataLinks "available")}}
+                            {{t "components.author-assertions.available.available"}}
+                        {{/if}}
+                        {{#if (eq hasDataLinks "no")}}
+                            {{t "components.author-assertions.available.no"}}
+                        {{/if}}
+                        {{#if (eq hasDataLinks "not_applicable")}}
+                            {{t "components.author-assertions.available.not_applicable"}}
+                        {{/if}}
+                    </p>
```
fabmiz, line 116: According to the mockups (saved states), you should add another `<p>` here that holds the actual value (based) on the model? So if it's `available`, you should have the links listed here; if it's `no`, you should have the `whyNo*` or the placeholder, and so forth.

app/components/multiple-textbox-input.js
fabmiz: ah, right! Got things mixed up. 🤕

app/controllers/submit.js
```diff
@@ -288,6 +297,16 @@ export default Controller.extend(Analytics, BasicsValidations, COIValidations, N
     // Does preprint have saved coi?
     savedCoi: computed.notEmpty('model.hasCoi'),
 
+    // Does preprint have saved hasDataLinks?
```
fabmiz, line 300: Delete or update?

PR #709, opened 2020-04-16
[LukewarmFix] Make another v2 request on content route

adlius:
In the `content` route, we fire an extra `v2` request for the waffle flags because somehow the first request sent upon app initialization doesn't get the correct flags.


fabmiz:
@adlius  PR description please! The only issue with this fix is that the extra request is fired all the times whenever a transition into the content route occurs (no matter logging in or not) but that can be addressed later.

PR #713, opened 2020-05-27
[ENG-1864][ENG-1838] Add helpful error messages when url is invalid. Fix remove save error.

adlius:
## Purpose
This PR do three things:
- Show helpful error messages returned from the backend when the url input for data links/prereg links are invalid.
- Fix various save error due to how we PATCH data.
- Fix "Discard Changes"  not working when removing links.

- Take error from API response and toast that error in `_failMoveFromAuthorAssertions`.
- Modify `multiple-textbox-input` component.
  - Instead of relying on `valuePath`, we directly pass the computed property into the component for updates so that "Discard changes" may work.
- In `updateHasDataLinks` and `updateHasPreregLinks`, we no longer reset `dataLinks`, `preregLinks` and `preregLinkInfo` because the backend doesn't like `null` for those fields.
  - It was suggested that we set `dataLinks` and `preregLinks` to empty arrays, which the backend does accept. However this doesn't work due to an edge case where user clicks `Available` and then `No`, at which point `dataLinks` will be set to `[]` and the backend doesn't accept `"dataLinks": []` when `"hasDataLinks": 'no'`. (I know, this is really messy...)

## Testing Notes
No notes other than trying the previous test cases. May ping me to test together if it makes it easier

## Ticket

https://openscience.atlassian.net/browse/ENG-1864
https://openscience.atlassian.net/browse/ENG-1838

## Notes for Reviewer

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/multiple-textbox-input.js
+++ b/app/components/multiple-textbox-input.js
@@ -11,7 +11,7 @@ export default Component.extend(Analytics, {
 
     didReceiveAttrs() {
         const valuesFromModel = this.model;
-        if (valuesFromModel.length > 0) {
+        if (valuesFromModel && valuesFromModel.length > 0) {
             this.set('textFields', valuesFromModel.map((value) => {
                 return { value };
             }));
--- a/app/components/multiple-textbox-input.js
+++ b/app/components/multiple-textbox-input.js
@@ -4,21 +4,20 @@ import Analytics from 'ember-osf/mixins/analytics';
 
 export default Component.extend(Analytics, {
     model: null,
-    valuePath: null,
     textFields: null,
     textFieldsLastIndex: computed('textFields.[]', function() {
         return this.get('textFields').length - 1;
     }),
 
     didReceiveAttrs() {
-        const valuesFromModel = this.model.get(this.valuePath);
-        if (valuesFromModel) {
+        const valuesFromModel = this.model;
+        if (valuesFromModel.length > 0) {
             this.set('textFields', valuesFromModel.map((value) => {
                 return { value };
             }));
-            return;
+        } else {
+            this.set('textFields', [{ value: '' }]);
         }
-        this.set('textFields', [{ value: '' }]);
     },
 
     actions: {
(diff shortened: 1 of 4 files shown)


app/components/multiple-textbox-input.js
```diff
@@ -4,21 +4,20 @@ import Analytics from 'ember-osf/mixins/analytics';
 
 export default Component.extend(Analytics, {
     model: null,
```
fabmiz, line 6: Consider renaming to something other than model.

app/controllers/submit.js
```diff
@@ -1189,12 +1189,9 @@ export default Controller.extend(Analytics, BasicsValidations, COIValidations, N
         },
         updateHasDataLinks(value) {
             this.set('hasDataLinks', value);
-            this.set('dataLinks', null);
         },
         updateHasPreregLinks(value) {
             this.set('hasPreregLinks', value);
-            this.set('preregLinks', null);
```
fabmiz, line 1196: Address case where old data is submitted in the patch.

app/templates/submit.hbs
```diff
@@ -194,8 +194,7 @@
                                         <br>
                                         {{#if (eq hasDataLinks 'available')}}
                                             {{#multiple-textbox-input
-                                                model=this
-                                                valuePath='dataLinks'
+                                                model=dataLinks
```
fabmiz, line 197: Pass in an action that modifies `dataLinks`

app/templates/submit.hbs
```diff
@@ -257,8 +256,7 @@
                                         {{#if (eq hasPreregLinks 'available')}}
                                             <div class='form-group'>
                                                 {{#multiple-textbox-input
-                                                    model=this
-                                                    valuePath='preregLinks'
+                                                    model=preregLinks
```
fabmiz, line 259: Pass in an action that modifies preregLinks

adlius:
@fabmiz Back to you. I've updated how we reset `dataLinks`, `preregLinks` and `prereLinkInfo`. However I did not change how we update the field because we did not use action to change the computed before this change.

PR #725, opened 2021-04-06
[ENG-2556] Allow more email domains in validation

futa-ikeda:
## Purpose
- Allow more email domains to be accepted for preprint users

- Update email validation regex to remove hard-coded allowed domains and use a more generalized email validation regex

## Testing Notes
- Please make sure that the email domains listed in the ticket are not rejected in the `Add author by email` section of the preprint submission process.

## Ticket

https://openscience.atlassian.net/browse/ENG-2556

## Notes for Reviewer
- the regex used in this PR was taken from [this stackoverflow thread](https://stackoverflow.com/questions/46155/how-to-validate-an-email-address-in-javascript). It's not the easiest to read I feel, but I think it does what is needed without being too loose on the validation

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/unregistered-contributor-form.js
```diff
@@ -23,7 +23,7 @@ const Validations = buildValidations({
             validator('presence', true),
             validator('format', {
                 type: 'email',
-                regex: /^[-a-z0-9~!$%^&*_=+}{'?]+(\.[-a-z0-9~!$%^&*_=+}{'?]+)*@([a-z0-9_][-a-z0-9_]*(\.[-a-z0-9_]+)*\.(aero|arpa|biz|com|coop|edu|gov|info|int|mil|museum|name|net|org|pro|travel|mobi|swiss|[a-z][a-z])|([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}))(:[0-9]{1,5})?$/i,
+                regex: /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
```
brianjgeiger, line 26: It would be good to put a comment in the code documenting where you found the regex. Why did you go with this one over the one at https://emailregex.com ?

brianjgeiger (commented):
One comment and a question

app/components/unregistered-contributor-form.js
futa-ikeda: It looks like the JS regex in this site and the regex used in the stackoverflow link/this PR are the same. Although it looks like the link you posted escapes the `[`, so that might be better? Either way, I will leave a comment here, as the diagram in your link is very handy!

brianjgeiger (approved):

adlius (approved):

adlius:
@futa-ikeda I remember we talked about adding some unit tests (without using the component) here to check the regex.

tests/unit/utils/email-validation-test.js
```diff
@@ -0,0 +1,32 @@
...
+import { module, test } from 'qunit';
+
+module('Unit | Utility | email validation');
+
+test('check against allowed email domains', function(assert) {
+    const allowedDomains = [
+        'endo@domain.org',
+        'tsukahara@email.domain.edu',
+        'kasa!matsu@[123.321.456.789]',
+        'yama#waki@0ph1.1pb2.hb',
```
brianjgeiger, line 11: Could you check with Eric to get some of the domains that have failed previously and include them in these tests? That way we're testing against the actual case we're trying to fix.

brianjgeiger (approved):

fabmiz (approved):
🥇

PR #752, opened 2023-07-03
[ENG-4572] Delete discover route

adlius:
## Ticket

https://openscience.atlassian.net/browse/IN-

## Notes for Reviewer

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


futa-ikeda (approved):
I think `subject` and `sources` queryparams will need to be updated to `filter[subject]` and `filter[publisher]`, but we can save that for integration, since we don't know what the IRI will look like for these just yet

futa-ikeda:
Note from scrum: there is a slight chance we may have to revert this PR if we need to support the `additionalProviders` going forward. The reason being, we may need to make a rule in NginX so that a provider with `additionalProviders` (livedata/ResearchAZ) goes to this version of the preprints/discover page and not the new one in ember-osf-web

